PR #31, opened 2024-02-15
Add single point calculations

ElliottKasoar:
Resolves #30

Initial code to configure calculators and perform single point calculations for energies, forces and stress.

Largely based on code already in [lavello](https://gitlab.com/drFaustroll/lavello/), but refactors, fixes various issues to get tests etc. to pass, and adds more general tests.

Now also adds the ability to return forces and stress.

To do:

- [x] Fix pylint
  - I had previously fixed with with try...except clauses, but it makes it quite verbose. Disabling pylint's complaint is an alternative, since the error message is pretty understandable.
- [x]  Add support for trajectory energies
  - The obvious solution is change `read` to read a specified slice/all structures, and attach the calculator to each image.
  - Are there better ways of doing this? As far as I can tell, calculators can only be attached to `Atoms` objects.
- [x] Update docstrings in various places
- [x] Add single point forces, stresses etc.
  - How (probably a dictionary)/when do we want to return these? 
- [x] Replace UiO66 with smaller structure that we are still able to calculate stress for e.g. NaCl from @alinelena

Note: CI is currently not working due to issues with pulling ASE from GitLab, but all tests pass locally.

PR #33, opened 2024-02-16
Add geo opt

### LINKED ISSUE #32: Add geometry optimisation
In addition to returning the optimised structure, we should optionally return/save the intermediate structures.

ElliottKasoar:
Resolves #32

This will require rebasing once #31 is merged.

To do:

- [ ] Check atoms returned correctly
- [ ] Add functionality to return intermediate structures


janus_core/geo_opt.py
oerc0122, line 1: Prefer `geom_opt` as a name, simply because it's more standard nomenclature for this.

janus_core/geo_opt.py
```diff
@@ -0,0 +1,55 @@
...
+try:
+    from ase.filters import FrechetCellFilter as DefaultFilter
+except ImportError:
+    from ase.constraints import ExpCellFilter as DefaultFilter
+
+from ase.optimize import LBFGS
+
+
+#  pylint: disable=dangerous-default-value
+#  Dictionaries are not modified within the function.
```
oerc0122, line 14: I would still prefer:
```python
, dyn_kwargs = None, ...
```

with:
```python
dyn_kwargs = dyn_kwargs or {}
```
or
```python
dyn_kwargs = dyn_kwargs if dyn_kwargs else {}
```
rather than disabling the warning

Also, I believe this is disabling file-wide which you certainly don't want. (c.f. https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html)

janus_core/geo_opt.py
```diff
@@ -0,0 +1,55 @@
...
+def optimize(
+    atoms: Atoms,
+    fmax: float = 0.1,
+    dyn_kwargs: dict = {},
```
oerc0122, line 18: May want to specify specific dict
```suggestion
    dyn_kwargs: dict[str, Any] = {},
```
If supporting pre-3.10, need to use `Dict` from `typing`

janus_core/geo_opt.py
```diff
@@ -0,0 +1,55 @@
...
+    filter_func: callable = DefaultFilter,
```
oerc0122, line 19: If `None` is an option, as provided below, should be `Optional[callable]`

Probably also doesn't want to be generic `callable`, but a subclass of an ASE filter base type if that's a thing.

tests/test_geo_opt.py
```diff
@@ -0,0 +1,126 @@
...
+import pytest
+
+from janus_core.geo_opt import optimize
+from janus_core.single_point import SinglePoint
+
+
+def test_optimize():
+    """Test optimizing geometry using MACE."""
+    data_path = Path(Path(__file__).parent, "data", "NaCl.cif")
+    model_path = Path(Path(__file__).parent, "models", "mace_mp_small.model")
```
oerc0122, line 18:
```suggestion
    curr_dir = Path(__file__).parent
    data_path = curr_dir / "data" / "NaCl.cif"
    model_path = curr_dir / "models" / "mace_mp_small.model"
```

tests/test_geo_opt.py
```diff
@@ -0,0 +1,126 @@
...
+
+    assert atoms.get_potential_energy() < init_energy
+    assert atoms.get_potential_energy() == pytest.approx(-27.046359959669214)
+
+
+@pytest.mark.slow
+def test_fmax():
+    """Test changing fmax."""
+    data_path = Path(Path(__file__).parent, "data", "NaCl.cif")
+    model_path = Path(Path(__file__).parent, "models", "mace_mp_small.model")
```
oerc0122, line 35: Looking at this, we probably want top-level globals pointing to `data` and `models` folders rather than this. But see above

```python
DATA_PATH = Path(__file__).parent / "data"
DATA_PATH = Path(__file__).parent / "models"
```

tests/test_geo_opt.py
```diff
@@ -0,0 +1,126 @@
...
+
+    atoms = optimize(single_point.sys)
+    init_energy = atoms.get_potential_energy()
+    atoms = optimize(single_point.sys, fmax=0.001)
+
+    assert atoms.get_potential_energy() < init_energy
+    assert atoms.get_potential_energy() == pytest.approx(-27.04636199814088)
+
+
+@pytest.mark.slow
```
oerc0122, line 48: Take a look at [`pytest.mark.parametrize`](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) to avoid needing to define this many identically structured test functions

oerc0122 (commented):
Mostly stylistic, but some over-generalisations on type-hints which would be nice to address.

ElliottKasoar:
> * [x]  Add functionality to return intermediate structures

This was actually already possible via `opt_kwargs` (see example in unit tests), which I think is sufficient, rather than adding another argument.

As with single point calculations, we need to document this soon and add examples (#34), even more so now the tests are a little harder to read off.

janus_core/geo_opt.py
ElliottKasoar: I was on the fence about this, as it feels unnecessarily verbose, but you're right, it's probably better to be safe. Should be fixed now.

ElliottKasoar:
> Mostly stylistic, but some over-generalisations on type-hints which would be nice to address.

Thanks for the suggestions! I think I've addressed most of them, with a couple of slightly open questions/comments.

One thing you may notice is that there's now a mix of `Union` and `x | y` typing. When we include `from __future__ import annotations`, `pyupgrade` prefers  `x | y` where possible.

However, this seemed to break linking to ASE's documentation via `intersphinx_mapping`, so I've reverted back to `Union` where necessary.

janus_core/single_point.py
```diff
@@ -13,13 +14,12 @@
...
     """Perpare and perform single point calculations."""
 
-    #  pylint: disable=dangerous-default-value
     def __init__(
         self,
         system: str,
         architecture: str = "mace_mp",
         device: str = "cpu",
-        read_kwargs: dict = {},
+        read_kwargs: dict[str, Any] | None = None,
```
oerc0122, line 22: I do think we want consistency in type hints. Either all with `__future__` and `|` or all with `Union`. But that can happily be another PR

janus_core/geom_opt.py
```diff
@@ -0,0 +1,72 @@
...
+except ImportError:
+    from ase.constraints import ExpCellFilter as DefaultFilter
+
+from ase.optimize import LBFGS
+
+
+def optimize(
+    atoms: Atoms,
+    fmax: float = 0.1,
+    dyn_kwargs: Union[dict[str, Any], None] = None,
```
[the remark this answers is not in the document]
ElliottKasoar: Ah I think when I added `Optional` with `__future__`, it replaced wanted to replace it with `| None`, but yes `Optional` seems preferable to `Union`, so should be fixed now.

janus_core/single_point.py
ElliottKasoar: I've dropped `__future__ ` for now, so I think it's all consistent, but happy to revisit later if we decide it was better before (and get get ASE to play along).

janus_core/single_point.py
```diff
@@ -19,7 +17,7 @@ def __init__(
         system: str,
         architecture: str = "mace_mp",
         device: str = "cpu",
```
oerc0122, line 19: The only other question I have is whether there is a list of valid entries in which case you can use e.g. `Literal["cpu", "gpu"]`, but am approving.

oerc0122 (approved):

janus_core/single_point.py
alinelena: shall be , cpu, cuda, and mps at the moment  if I missed it we shall open a bug about it.

ElliottKasoar:
One additional option I've realised that @alinelena included in lavello/md.py is to read saved trajectory (a pickled file) and rewrite e.g. as an extxyz:

```python
  x = read(traj, index=":")
  write(opt_traj, x, format="extxyz")
  write(pathlib.PurePosixPath(output).stem+'.cif',a,format="cif")
  try:
     p = pathlib.Path(traj)
     p.unlink()
  except:
     pass
```
Do we want to add this in as well, or is the current option of saving the .traj binary file enough?

oerc0122:
> One additional option I've realised that @alinelena included in lavello/md.py is to read the saved trajectory (a pickled file) and rewrite e.g. as an extxyz:
> 
> ```python
>   x = read(traj, index=":")
>   write(opt_traj, x, format="extxyz")
>   try:
>      p = pathlib.Path(traj)
>      p.unlink()
>   except:
>      pass
> ```
> 
> Do we want to add this in as well, or is the current option of saving the .traj binary file enough?

Why is the `Path` creation in the `try`? Also, never use a bare `except`, `unlink` also has [`missing_ok=False`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink) for this situation.

ElliottKasoar:
> Why is the `Path` creation in the `try`? Also, never use a bare `except`, `unlink` also has [`missing_ok=True`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink) for this situation.

I was mostly just quoting the code that was in md.py for the gist, but good points, I'll fix those if we decide it's worth adding.

conftest.py
```diff
@@ -14,21 +14,31 @@ def pytest_addoption(parser):
         default=False,
         help="Test additional MLIPs",
     )
+    parser.addoption(
```
alinelena, line 17: i think for this we shall use typer.

ElliottKasoar:
> One additional option I've realised that @alinelena included in lavello/md.py is to read the saved trajectory (a pickled file) and rewrite e.g. as an extxyz:

The way this is done isn't quite a neat as I'd like, as you specify both the .traj file and the .xyz file, but I thought it's probably safer to require both be specified since the former is deleted after creation of the latter.

I also changed saving the optimised structure to act in the same way as I've implemented this - requiring `filename` in the `kwargs` rather than its own variable, so it's slightly more condensed.

janus_core/single_point.py
```diff
@@ -34,16 +32,17 @@ def __init__(
             Default is "mace_mp".
         device : str
             Device to run model on. Default is "cpu".
-        read_kwargs : dict
-            kwargs to pass to ase.io.read. Default is {}.
+        read_kwargs : Optional[dict[str, Any]]
+            kwargs to pass to ase.io.read. Default is None.
```
oerc0122, line 36: Default is still `{}` due to first line using it.

janus_core/single_point.py
```diff
@@ -54,20 +53,23 @@ def read_system(self, **kwargs) -> None:
...
+        self, read_kwargs: Optional[dict[str, Any]] = None, **kwargs
+    ) -> None:
         """Configure calculator and attach to system.
 
         Parameters
         ----------
-        read_kwargs : dict
-            kwargs to pass to ase.io.read. Default is {}.
+        read_kwargs : Optional[dict[str, Any]]
+            kwargs to pass to ase.io.read. Default is None.
```
oerc0122, line 64: And elsewhere

janus_core/geom_opt.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Write out optimized structure
+    if struct_kwargs:
+        write(images=atoms, **struct_kwargs)
+
+    # Reformat trajectory file from binary
+    if traj_kwargs:
+        traj = read(opt_kwargs["trajectory"], index=":")
+        write(images=traj, **traj_kwargs)
+        Path(opt_kwargs["trajectory"]).unlink()
```
alinelena, line 92: will this fail if the path does not exist?

ElliottKasoar: Yes it will throw a `FileNotFoundError`, but I think there would have to have been an error in writing out the trajectory first during optimisation, so I think we probably want it to throw an error if it somehow gets here?

conftest.py
ElliottKasoar: Is it worth adding that now as a dependency when we only have these for pytest, or change it later when we start add CLI for the main code?

janus_core/geom_opt.py
```diff
@@ -0,0 +1,94 @@
...
+    ----------
+    atoms : Atoms
+        Atoms object to optimize geometry for.
+    fmax : float
+        Set force convergence criteria for optimizer in units eV/Å.
+    dyn_kwargs : Optional[dict[str, Any]]
+        kwargs to pass to dyn.run. Default is None.
+    filter_func : Optional[callable]
+        Apply constraints to atoms through ASE filter function.
+        Default is `FrechetCellFilter`.
```
oerc0122, line 40: Not strictly... if you don't have `Frechet` it is `ExpCell`, should we assume this module should fail if the `import` of `FrechetFilter` fails?

conftest.py
alinelena: happy with that but just put an issue about it.

janus_core/geom_opt.py
ElliottKasoar: We can't expect the module to fail if importing `FrechetCellFilter` fails, as for ASE 3.22.1 it didn't exist, hence the try...except on importing, but I intended it as the default in the sense that it must be overruled, whether by the user or the import error. Can docstrings cause the module to fail?

Is there a standard way of writing an import-dependent default? Should I just use `DefaultFilter`?

oerc0122: Docstrings don't cause failures unless you use `doctest` (which is a nice thing to do and we could make an issue about that as general recommendations for simple routines), but this is to be read by humans too could just go with
```suggestion
        Default is `FrechetCellFilter` if `ASE > 3.22.1` otherwise `ExpCellFilter`.
```

ElliottKasoar: Isn't that also not strictly true, as there will be a specific (but probably not very helpful to reference) commit where `FrechetCellFilter`, and the filters module as a whole, was added to ASE at some point after there were no new tags?

conftest.py
ElliottKasoar: Cool, will do. It's also possible we can't change this one easily, as we don't explicitly import argparse here, but we can look into it when we add it elsewhere.

janus_core/geom_opt.py
oerc0122: `if available` then? 

That's why I wouldn't recommend referencing a commit, rather a release. 

PyPI doesn't allow multiple releases with the same name, so that release should have it and if you want to install a compatible version this you can do `pip install ASE>=3.22.1` hence my notation.

oerc0122 (approved):

janus_core/geom_opt.py
alinelena: yap I will not hold my horses...

janus_core/geom_opt.py
alinelena: actually on second thought I will let the .traj on the hdd anyhow and the user can clean it up...

ElliottKasoar: As in just remove `Path(opt_kwargs["trajectory"]).unlink()`, or get rid of the entire reload/saving?

alinelena: just the uinlink line.

alinelena (approved):

oerc0122 (approved):

PR #42, opened 2024-02-23
Specify device consistently

### LINKED ISSUE #40: Fix device arguments
`CHGNetCalculator` expects `use_device`, whereas `MACECalculator` expects `device` as key word arguments, but the options for both should be the same as those defined by [PyTorch](https://pytorch.org/docs/stable/tensor_attributes.html#torch.device), so the interface could be unified.

Currently this is made particularly misleading by an explicit `device` argument in [single_point.py](https://github.com/stfc/janus-core/blob/main/janus_core/single_point.py#L21) that is not usable by `CHGNetCalculator`.

ElliottKasoar:
Resolves #40

Note: There may be a way to pass device to chgnet as well, but currently this makes things consistent between MACE, m3gnet and most likely other PyTorch-based models.

Also tidies single point calculation and calculator code addressing comments in #32.

### DIFF
--- a/janus_core/mlip_calculators.py
+++ b/janus_core/mlip_calculators.py
@@ -10,10 +10,13 @@ from typing import Literal
 from ase.calculators.calculator import Calculator
 
 architectures = ["mace", "mace_mp", "mace_off", "m3gnet", "chgnet"]
+devices = ["cpu", "cuda", "mps"]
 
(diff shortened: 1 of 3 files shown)


oerc0122 (approved):
Looks a lot neater to me

PR #48, opened 2024-02-29
Add extended typing through janus_types module

oerc0122:
Quick rough-and-ready types module
- Need to consider if trying to suggest arguments for ASE routines is sensible.


janus_core/single_point.py
```diff
@@ -72,44 +73,45 @@ def set_calculator(
             read_kwargs = read_kwargs if read_kwargs else {}
             self.read_system(**read_kwargs)
 
-        if isinstance(self.sys, list):
+        elif isinstance(self.sys, list):
```
oerc0122, line 76: This may have resulted in an interesting error if `sys` was none. Maybe shouldn't be fixed here.

alinelena: shall we add a test for it?

oerc0122: Ah, ok. Didn't realise that about `read_system` hence they may. mypy was complaining about potential NoneType

ElliottKasoar (changes requested):
@alinelena suggested adding the changes in #50 here, so it's easier to see the reason for the changes, which I think makes sense.

I would directly add the changes as commit suggestions, but I'm not sure you can if files haven't been touched in the PR?

janus_core/janus_types.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import IO, List, Literal, Optional, Sequence, TypedDict, TypeVar, Union
+
+from ase import Atoms
+import numpy as np
+from numpy.typing import NDArray
+
+# General
+
+T = TypeVar("T")
+MaybeList = Union[T, List[T]]
```
ElliottKasoar, line 15:
```suggestion
from typing import IO, Literal, Optional, Sequence, TypedDict, TypeVar, Union

from ase import Atoms
import numpy as np
from numpy.typing import NDArray

# General

T = TypeVar("T")
MaybeList = Union[T, list[T]]
```
I think `list` is preferable  `List` for [Python 3.9+](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections)?

alinelena: if I understand correctly typing(which may not be the case, list is to be used when you do not care about what the list is of... while use list when you want to specify the type of elements. so I suspect shall be List[T]... but @ElliottKasoar and @oerc0122 I can be corrected.

alinelena:
@oerc0122 except the need to settle the issue above I think we are fine with this one, will need rebased to be sure we do not have anything failing in ci.

janus_core/janus_types.py
oerc0122: That is incorrect as of 3.9. `list` and most other standard Python types adopted the generic interfaces allowing you to specify sub-types without importing `typing`

oerc0122:
> @oerc0122 except the need to settle the issue above I think we are fine with this one, will need rebased to be sure we do not have anything failing in ci.

Already done

janus_core/janus_types.py
alinelena: perfect so I will commit the changes.

alinelena (dismissed):

docs/source/conf.py
```diff
@@ -35,7 +35,12 @@
 ]
 
 numpydoc_validation_checks = {"all", "EX01", "SA01", "ES01"}
-numpydoc_validation_exclude = {r"\.__weakref__$", r"\.__repr__$"}
+numpydoc_validation_exclude = {
+    r"\.__weakref__$",
+    r"\.__repr__$",
```
ElliottKasoar, line 40: Are we able to remove these excludes safely now?

docs/source/conf.py
```diff
@@ -185,4 +190,5 @@
 # We should ignore any python built-in exception, for instance
 nitpick_ignore = [
     ("py:class", "Logger"),
+    ("py:class", "numpy.float64"),
```
ElliottKasoar, line 193: Why is this needed?

janus_core/janus_types.py
```diff
@@ -0,0 +1,69 @@
...
+    parallel: bool
+    append: bool
+
+
+class ASEOptArgs(TypedDict, total=False):
+    """Main arugments for ase optimisers."""
+
+    restart: Optional[bool]
+    logfile: Optional[PathLike]
+    trajectory: PathLike
```
ElliottKasoar, line 49:
```suggestion
    trajectory: Optional[str]
```

I don't think this actually accepts `pathlib.Path` etc., only `str`, unlike ase.io.read/write.

ElliottKasoar (changes requested):

docs/source/conf.py
ElliottKasoar: (We are not)

docs/source/conf.py
ElliottKasoar: Suppresses `WARNING: py:class reference target not found: numpy.float64`.

It's odd that this isn't resolved by linking the numpy docs via `intersphinx_mapping` (which resolves similar warnings for other classes e.g. `numpy.dtype`), but seems to be necessary

ElliottKasoar (approved):

janus_core/single_point.py
```diff
@@ -20,22 +29,22 @@ class SinglePoint:
...
         MLIP architecture to use for single point calculations.
         Default is "mace_mp".
-    device : Literal[devices]
+    device : Devices
         Device to run model on. Default is "cpu".
-    read_kwargs : Optional[dict[str, Any]]
+    read_kwargs : ASEReadArgs
         Keyword arguments to pass to ase.io.read. Default is {}.
-    calc_kwargs : Optional[dict[str, Any]]
+    calc_kwargs : ASEReadArgs
```
ElliottKasoar, line 36:
```suggestion
    calc_kwargs : Optional[dict[str, Any]]
```
Unless we want CalcOptArgs

ElliottKasoar (changes requested):

ElliottKasoar (approved):

Issue #47, opened 2024-02-29
Add option to provide Atoms objects for structure


ElliottKasoar:
Currently single point calculations and geometry optimization expect a file path/string, but as suggested by @alinelena, it would also be useful to be able to pass ASE `Atoms` objects directly.

We could overload `read_system` (or similar), which would also allow us to extend this to further similar Atoms-like objects in the future.

oerc0122:
The thing here is the optional(?) dependencies can get quite messy. Then again things like `functools.singledispatch` can make it quite clean.

ElliottKasoar:
ASE isn't optional, so I think for this it should be fine. If/when we want to extend it further, we'll definitely have to think more carefully.

PR #49, opened 2024-03-01
Add initial CLI

### LINKED ISSUE #39: Add typer for CLI

ElliottKasoar:
Resolves #39 

Adds CLI commands for single point calculations.

Includes a dummy command so that it runs via `janus singlepoint` rather than just `janus`.

To do:
- [x] Add tests

For reference, this doesn't intend to:
- Add logging
- Write output of single point calculations
- Add geometry optimization CLI command

These will be added (via the CLI and otherwise) in future PRs.

### DIFF
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -62,7 +62,9 @@ def parse_dict_class(value: str):
 
 @app.command()
 def singlepoint(
-    system: Annotated[str, typer.Option(help="Path to system to perform calculations")],
+    structure: Annotated[
+        str, typer.Option(help="Path to structure to perform calculations")
+    ],
     architecture: Annotated[
         str, typer.Option("--arch", help="MLIP architecture to use for calculations")
     ] = "mace_mp",
@@ -96,8 +98,8 @@ def singlepoint(
 
     Parameters
     ----------
-    system : str
-        System to simulate.
+    structure : str
+        Structure to simulate.
     architecture : Optional[str]
         MLIP architecture to use for single point calculations.
         Default is "mace_mp".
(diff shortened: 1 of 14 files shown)


ElliottKasoar:
Note: Some of the typing is a bit inconvenient, hence there are some slight differences compared to the module functions.

For example, it can't handle Unions or dictionaries (hence my custom type). Lists are handled by repeatedly specifying the option e.g. `--property energy --property forces`, which actually works out slightly better than I expected in this case, since it ends up a lot like `str | list[str]`.

janus_core/cli.py
```diff
@@ -0,0 +1,118 @@
...
+    read_kwargs = read_kwargs.value if read_kwargs else {}
+    calc_kwargs = calc_kwargs.value if calc_kwargs else {}
+
+    if not isinstance(read_kwargs, dict):
+        raise ValueError("read_kwargs must be a dictionary")
+    if not isinstance(calc_kwargs, dict):
+        raise ValueError("calc_kwargs must be a dictionary")
+
+    s_point = SinglePoint(
+        system=system,
```
alinelena, line 99: since in aiida they call system structure, shall we rename it?

ElliottKasoar: Just for the CLI, or shall I change it everywhere?

alinelena: everywhere will be nice to be consistent if is not a lot of work.

ElliottKasoar: > also where are --log and --output

I mention them in the PR description - since they're new features that we don't have for the main functions yet, I thought it made more sense to add the CLI skeleton first and them add the features, including CLI integration, as separate PRs

alinelena: cool, then the rename shall go in that one too

ElliottKasoar:
Sorry, I'd  just added the renaming here already. I could remove it, but it's worth noting, I expect to change things a bit for #47, as I think it might make more sense to change `structure` to an optional `struct_path`, and have the alternative being setting `struct` directly via the `Atoms` object (or future non-ASE objects).

alinelena (dismissed):

alinelena (dismissed):

PR #55, opened 2024-03-05
Write output from singlepoint

### LINKED ISSUE #53: Write output of single point calculations
Particularly when running single point calculations via the CLI (#51), we should write out the results of single point calculations to file.

ElliottKasoar:
Resolves #53

To do:

- [x] Add to CLI with test(s)
- [x] Rebase to include changes from #48

### DIFF
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -95,15 +95,12 @@ def singlepoint(
             metavar="DICT",
         ),
     ] = None,
-    write_results: Annotated[
-        bool, typer.Option(help="Write out structure with results of calculations")
-    ] = False,
     write_kwargs: Annotated[
         TyperDict,
         typer.Option(
             parser=parse_dict_class,
             help=(
-                "Keyword arguments to pass to ase.io.write to save "
+                "Keyword arguments to pass to ase.io.write when saving "
                 "results [default: {}]"
             ),
             metavar="DICT",
@@ -111,7 +108,7 @@ def singlepoint(
     ] = None,
 ):
     """
-    Perform single point calculations.
+    Perform single point calculations and save to file.
 
     Parameters
     ----------
@@ -128,10 +125,8 @@ def singlepoint(
         Keyword arguments to pass to ase.io.read. Default is {}.
     calc_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to the selected calculator. Default is {}.
-    write_results : bool
-        True to write out structure with results of calculations. Default is False.
     write_kwargs : Optional[dict[str, Any]]
-        Keyword arguments to pass to ase.io.write if saving results. Default is {}.
+        Keyword arguments to pass to ase.io.write when saving results. Default is {}.
     """
     read_kwargs = read_kwargs.value if read_kwargs else {}
     calc_kwargs = calc_kwargs.value if calc_kwargs else {}
@@ -151,12 +146,8 @@ def singlepoint(
         read_kwargs=read_kwargs,
         calc_kwargs=calc_kwargs,
     )
-    print(
-        s_point.run_single_point(
-            properties=properties,
-            write_results=write_results,
-            write_kwargs=write_kwargs,
-        )
+    s_point.run_single_point(
+        properties=properties, write_results=True, write_kwargs=write_kwargs
     )
 
 
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -95,6 +95,9 @@ def singlepoint(
             metavar="DICT",
         ),
     ] = None,
+    write_results: Annotated[
+        bool, typer.Option(help="Write out structure with results of calculations")
+    ] = False,
     write_kwargs: Annotated[
         TyperDict,
         typer.Option(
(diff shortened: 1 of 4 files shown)


janus_core/single_point.py
```diff
@@ -181,23 +183,33 @@ def run_single_point(
...
+
+        results = {}
         if "energy" in properties or len(properties) == 0:
             results["energy"] = self._get_potential_energy()
         if "forces" in properties or len(properties) == 0:
             results["forces"] = self._get_forces()
         if "stress" in properties or len(properties) == 0:
             results["stress"] = self._get_stress()
 
+        if write_kwargs:
```
alinelena, line 212: i will have a special argument write_results... also there is an issue that will notice when you write for a non periodic system, stress is nan or inf

ElliottKasoar: Having it specified as `write_kwargs` means it matches writing out for geometry optimisation (final structure and/or trajectory), but yes I can add a `write_results` flag if we prefer that here.

In terms of `NaN`/`inf` values, to some extent isn't that something that should be handled by the calculator, rather than us? 

I hadn't realised until I added this, but even if we only run `get_potential_energy`, it sets all of the other values (forces, stress etc.) in `calc.results`, which makes sense in some ways, but it makes it more complex to write out, as we currently use the fact that it automatically writes out these `results`, so do we need to manually check these?

janus_core/cli.py
```diff
@@ -128,7 +151,13 @@ def singlepoint(
         read_kwargs=read_kwargs,
         calc_kwargs=calc_kwargs,
     )
-    print(s_point.run_single_point(properties=properties))
+    print(
```
alinelena, line 154: do we need this print? and if yes why?

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
 
         return self.struct.get_stress()
 
-    def run_single_point(self, properties: MaybeSequence[str] = ()) -> CalcResults:
+    @staticmethod
+    def _remove_invalid_props(
+        struct: Atoms,
+        properties: Optional[list[str]] = None,
```
oerc0122, line 187: May want
```suggestion
        properties: Collection[str] = (),
```

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
+        properties : Optional[list[str]]
+            Physical properties requested to be calculated. Default is [].
+        """
+        properties = properties if properties else []
+        rm_keys = []
+
+        # Find any properties with non-finite values
+        for prop in struct.calc.results:
+            if not isfinite(struct.calc.results[prop]).all():
+                rm_keys.append(prop)
```
oerc0122, line 205:
```suggestion
        # Find any properties with non-finite values
        rm_keys = [prop for prop in struct.calc.results
                   if not isfinite(struct.calc.results[prop]).all()]
```

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
+        # Raise error if property was explicitly requested, otherwise remove
+        for prop in rm_keys:
+            if prop in properties:
+                raise ValueError(
+                    f"'{prop}' contains non-finite values for this structure."
+                )
+            struct.calc.results.pop(prop)
```
oerc0122, line 212: Do you think this gives better intent as:
```suggestion
            del struct.calc.results[prop]
```
?

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
+
+    def _clean_results(self, properties: Optional[list[str]] = None) -> None:
```
oerc0122, line 214:
```suggestion
    def _clean_results(self, properties: Collection[str] = ()) -> None:
```

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
+        """
+        Remove results with NaN or inf values from calc.results dictionary.
+
+        Parameters
+        ----------
+        properties : Optional[list[str]]
```
oerc0122, line 220:
```suggestion
        properties : Collection[str]
```

janus_core/single_point.py
```diff
@@ -179,7 +181,59 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
+            Physical properties requested to be calculated. Default is [].
+        """
+        properties = properties if properties else []
```
oerc0122, line 223:
```suggestion
```

tests/test_single_point.py
```diff
@@ -73,3 +75,78 @@ def test_single_point_traj():
     results = single_point.run_single_point("energy")
     assert results["energy"][0] == pytest.approx(-76.0605725422795)
     assert results["energy"][1] == pytest.approx(-74.80419118083256)
+
+
+def test_single_point_write():
+    """Test writing singlepoint results."""
+    data_path = DATA_PATH / "NaCl.cif"
+    results_path = Path(".").absolute() / "NaCl-results.xyz"
+    assert not Path(results_path).exists()
```
oerc0122, line 84: `results_path` is already a `Path`

oerc0122: Could also be a use case for tempfile with the correct suffix

oerc0122 (commented):
Couple of minor simplifications (I hope)

PR #61, opened 2024-03-08
Add logger skeleton

### LINKED ISSUE #52: Add logger
This should be included via the CLI (#51) and otherwise, once merged.

ElliottKasoar:
Resolves #52

Adds a skeleton for logging, that currently just captures the start and end times for the single point calculation and saves to a log file, so the CLI behaves as expected for aiida-mlip.

I'm not sure there's a way to test the log file more rigorously, as pytest seems to interfere, but it at least tests the contents of the log.

### DIFF
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -107,6 +107,9 @@ def singlepoint(
             metavar="DICT",
         ),
     ] = None,
+    log_file: Annotated[
+        str, typer.Option("--log", help="File to save logs")
+    ] = "singlepoint.log",
 ):
     """
     Perform single point calculations and save to file.
@@ -128,6 +131,8 @@ def singlepoint(
         Keyword arguments to pass to the selected calculator. Default is {}.
     write_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to ase.io.write when saving results. Default is {}.
+    log_file : Optional[str]
+        Name of log file to write logs to. Default is "singlepoint.log".
     """
     read_kwargs = read_kwargs.value if read_kwargs else {}
     calc_kwargs = calc_kwargs.value if calc_kwargs else {}
@@ -146,6 +151,7 @@ def singlepoint(
         device=device,
         read_kwargs=read_kwargs,
         calc_kwargs=calc_kwargs,
+        log_file=log_file,
     )
     s_point.run_single_point(
         properties=properties, write_results=True, write_kwargs=write_kwargs
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -108,7 +108,7 @@ def singlepoint(
         ),
     ] = None,
     log_file: Annotated[
-        str, typer.Option("--log", help="File to save logs")
+        Path, typer.Option("--log", help="Path to save logs to")
     ] = "singlepoint.log",
 ):
     """
@@ -131,8 +131,8 @@ def singlepoint(
         Keyword arguments to pass to the selected calculator. Default is {}.
     write_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to ase.io.write when saving results. Default is {}.
-    log_file : Optional[str]
-        Name of log file to write logs to. Default is "singlepoint.log".
+    log_file : Optional[Path]
+        Path to write logs to. Default is "singlepoint.log".
     """
     read_kwargs = read_kwargs.value if read_kwargs else {}
     calc_kwargs = calc_kwargs.value if calc_kwargs else {}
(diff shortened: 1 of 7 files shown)


janus_core/cli.py
```diff
@@ -127,6 +130,8 @@ def singlepoint(
         Keyword arguments to pass to the selected calculator. Default is {}.
     write_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to ase.io.write when saving results. Default is {}.
+    log_file : Optional[str]
```
oerc0122, line 133: Does that want to be a path rather than a string?

janus_core/janus_types.py
```diff
@@ -56,6 +58,18 @@ class ASEOptRunArgs(TypedDict, total=False):
...
+class Log(Enum):
+    """Supported options for logger levels."""
+
+    DEBUG = logging.DEBUG
+    INFO = logging.INFO
+    WARNING = logging.WARNING
+    ERROR = logging.ERROR
+
+
+LogType = Literal[Log.DEBUG, Log.INFO, Log.ERROR, Log.WARNING]
```
oerc0122, line 70: You should be able to use `Log` as the type-hint and it should work

janus_core/janus_types.py
```diff
@@ -56,6 +58,18 @@ class ASEOptRunArgs(TypedDict, total=False):
     steps: int
 
 
+class Log(Enum):
```
oerc0122, line 61: I would probably call this `LogLevel` or similar

janus_core/cli.py
oerc0122: At the very least we may want to set up a type-alias to indicate where strings should be pathy

janus_core/log.py
```diff
@@ -0,0 +1,51 @@
+"""Configure logger."""
```
oerc0122, line 1: Mention somewhere logger is yaml like?

janus_core/single_point.py
```diff
@@ -267,6 +279,9 @@ def run_single_point(
         if write_kwargs and "filename" not in write_kwargs:
             raise ValueError("'filename' must be included in write_kwargs")
 
+        if self.logger is not None:
```
oerc0122, line 282: Can just be:
```suggestion
        if self.logger:
```

oerc0122: Could also configure an alias to this above when you define the logger.
**N.B** should go in `if` above one line here to save me formatting
```python
loginfo = self.logger.info if self.logger else lamba *x, **y: None
```
or some other nop

janus_core/single_point.py
```diff
@@ -85,7 +90,14 @@ def __init__(
             Keyword arguments to pass to ase.io.read. Default is {}.
         calc_kwargs : Optional[dict[str, Any]]
             Keyword arguments to pass to the selected calculator. Default is {}.
+        log_file : Optional[str]
+            Name of log file if writing logs. Default is None.
         """
+        if log_file is not None:
+            self.logger = config_logger(name=__name__, filename=log_file)
```
oerc0122, line 97: Could make `config_logger` handle `None` log.

janus_core/single_point.py
ElliottKasoar: I quite like the alias, but I think it might be better to stick with the more explicit form.

janus_core/cli.py
ElliottKasoar: Ah thought this was in single_point.py.

Typer doesn't support `Union` types yet, annoyingly, so I've changed it to `Path` here, and `PathLike` in single_point.py

janus_core/janus_types.py
```diff
@@ -56,6 +58,18 @@ class ASEOptRunArgs(TypedDict, total=False):
...
+class LogLevel(Enum):
+    """Supported options for logger levels."""
+
+    DEBUG = logging.DEBUG
+    INFO = logging.INFO
+    WARNING = logging.WARNING
+    ERROR = logging.ERROR
+
+
+LogLevelType = Literal[LogLevel]
```
oerc0122, line 70: Not sure this is right, [this says](https://stackoverflow.com/questions/52624736/type-annotations-for-enum-attribute) you should just be able to use the `Enum` itself. As written `LogLevelType` asks you to pass the LogLevel class itself, not any of its members

ElliottKasoar: Ah sorry, I misinterpreted your previous suggestion.

Yes, based on https://typing.readthedocs.io/en/latest/spec/literal.html#literal-types, just using LogLevel as the type hint should be "approximately" equivalent to what I had before (`Literal[LogLevel.debug, ...]`, so hopefully it's ok now.

alinelena (dismissed):

alinelena (approved):

PR #63, opened 2024-03-08
Add option to pass Atoms structure directly

### LINKED ISSUE #47: Add option to provide Atoms objects for structure

ElliottKasoar:
Resolves #47

As mentioned  in #49, my preference is not to overload something like `read_struct`, as was originally proposed in #47, but to separate reading from file (which could be overloaded for non-ASE readers) with passing a structure directly (which could also include non-ASE objects in future).

This is largely to enable changing the confusingly named `structure` (as opposed to `struct`) attribute to `struct_path`.

### DIFF
--- a/janus_core/single_point.py
+++ b/janus_core/single_point.py
@@ -28,7 +28,7 @@ class SinglePoint:
     Parameters
     ----------
     struct : Optional[MaybeSequence[Atoms]]
-        Structure or list of structures to simulate. Required if `struct_path` is None.
+        ASE Atoms structure(s) to simulate. Required if `struct_path` is None.
         Default is None.
     struct_path : Optional[str]
         Path of structure to simulate. Required if `struct` is None.
@@ -51,7 +51,7 @@ class SinglePoint:
     architecture : Architectures
         MLIP architecture to use for single point calculations.
     struct : MaybeSequence[Atoms]
-        ASE Atoms or list of Atoms structures to simulate.
+        ASE Atoms structure(s) to simulate.
     device : Devices
         Device to run MLIP model on.
     struct_path : Optional[str]
@@ -85,8 +85,8 @@ class SinglePoint:
         Parameters
         ----------
         struct : Optional[MaybeSequence[Atoms]]
-            Structure or list of structures to simulate.
-            Required if `struct_path` is None. Default is None.
+            ASE Atoms structure(s) to simulate. Required if `struct_path`
+            is None. Default is None.
         struct_path : Optional[str]
             Path of structure to simulate. Required if `struct` is None.
             Default is None.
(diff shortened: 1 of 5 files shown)


janus_core/single_point.py
```diff
@@ -63,7 +70,9 @@ class SinglePoint:
 
     def __init__(
         self,
-        structure: str,
+        struct: Optional[MaybeList[Atoms]] = None,
```
oerc0122, line 73: Says `MaybeSequence` above

tests/test_single_point.py
```diff
@@ -144,9 +146,78 @@ def test_single_point_write_nan(tmp_path):
...
         architecture="mace",
         calc_kwargs={"model_paths": MODEL_PATH},
     )
     with pytest.raises(NotImplementedError):
         single_point.run_single_point("invalid")
+
+
+def test_atoms():
+    """Test passing ASE Atoms structure."""
+    struct = read(DATA_PATH / "NaCl.cif")
```
oerc0122, line 159: Out of curiosity does ASE name its molecule sensibly you can use as a default?

alinelena: they have a collection of molecules and bulk methods.

something like bulk("NaCl", "rocksalt",cubic=True, a=5.63) will give you salt... something like molecule("H2O") will give you water... some extended one. 
we shall test both, file loaded and ase generated atoms. Elliott/Federica are both aware we travel in that direction

ElliottKasoar: There's not a way to do the reverse though, is there?

E.g. `ase.io.bulk("NaCl", "rocksalt",cubic=True, a=5.63)` gives an Atoms object `Atoms(symbols='NaClNaClNaClNaCl', pbc=True, cell=[5.63, 5.63, 5.63])`, but there's no label other than the symbols?

alinelena: nope there is not, what label are you after?

ElliottKasoar: I was just thinking it would be useful for the default structure name when we pass `Atoms` rather than the path, but I don't think it's a problem really.

alinelena: you can extract the info from Atoms.symbols that shall be enough in theory can be quite complex.

ElliottKasoar: True, yeah that's probably still better than the completely generic `"struct"`.

(I've used `get_chemical_formula` as `get_chemical_symbols` gives a list)

janus_core/single_point.py
```diff
@@ -86,14 +102,29 @@ def __init__(
         calc_kwargs : Optional[dict[str, Any]]
             Keyword arguments to pass to the selected calculator. Default is {}.
         """
-        self.architecture = architecture
-        self.device = device
-        self.structure = structure
+        if struct and struct_path:
+            raise ValueError("Only one of `struct` and `struct_path` may be specified")
```
alinelena, line 106: maybe be more specific in here about what you mean by structure, ase Atoms object.

alinelena (approved):

PR #67, opened 2024-03-12
Fix removing NaN results

### LINKED ISSUE #66: stress contains nan inf for isolated molecules
this 
```python
from ase.build import molecule
from ase.io import write

from janus_core.single_point import SinglePoint

sp = SinglePoint(struct=molecule('H2O'),
                 architecture="mace_off",
                 device='cpu',
                 calc_kwargs={'model_paths':'medium'})

res = sp.run_single_point()
print(res)
```

produces 

```
{'energy': -2081.11639716893, 'forces': array([[ 0.        ,  0.        , -0.64748683],
       [ 0.        , -0.3437656 ,  0.32374341],
       [ 0.        ,  0.3437656 ,  0.32374341]]), 'stress': array([nan, inf, inf, nan, nan, nan])}
```

since stress is rubbish shall not be returned in the first place.

ElliottKasoar:
Resolves #66

Ideally, we'd probably remove the separate `results` dictionary and rely on clean-up of `calc.results`, but we would still need something of similar form in order to return lists of properties for trajectories.

An alternative could be to remove setting results until after clean-up, but then we don't make use of the values returned by `Atoms.get_potential_energy` etc. This might still be slightly tidier, but is perhaps slightly more dangerous too given that MACE sets multiple results simultaneously, so in many cases, `energy` would be taken from the results from `get_stress`.

### DIFF
--- a/janus_core/single_point.py
+++ b/janus_core/single_point.py
@@ -236,18 +236,23 @@ class SinglePoint:
     @staticmethod
     def _remove_invalid_props(
         struct: Atoms,
+        results: CalcResults = None,
         properties: Collection[str] = (),
     ) -> None:
         """
-        Remove any invalid properties from calculator results.
+        Remove any invalid properties from calculated results.
 
         Parameters
         ----------
         struct : Atoms
             ASE Atoms structure with attached calculator results.
+        results : CalcResults
+            Dictionary of calculated results. Default is {}.
         properties : Collection[str]
             Physical properties requested to be calculated. Default is ().
         """
+        results = results if results else {}
+
         # Find any properties with non-finite values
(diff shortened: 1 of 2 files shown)


janus_core/single_point.py
```diff
@@ -327,7 +340,7 @@ def run_single_point(
         if "stress" in properties or len(properties) == 0:
             results["stress"] = self._get_stress()
 
-        self._clean_results(properties=properties)
+        self._clean_results(results, properties=properties)
```
alinelena, line 343: add a note above this saying why we need to call it... cause may look odd to someone who checks the code.

alinelena (approved):

PR #70, opened 2024-03-13
Add geomopt to CLI

### LINKED ISSUE #65: Extend CLI
The CLI currently only provides an interface for single point calculations, and not geometry optimisation.

Ideally we might want to share some CLI arguments, but I believe this isn't necessarily straightforward via typer.

ElliottKasoar:
Resolves #65

See issue for remaining work.

This also implements the logger in geom_opt.py slightly differently to single_point.py, in order to pass kwargs (specifically, to allow appending rather than write only). It probably makes sense to update single_point.py to match this too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,11 @@ poetry add git+https://gitlab.com:ase/ase.git#b31569210d739bd12c8ad2b6ec0290108e
 
 ## Examples
 
-Perform a single point calcuation:
+### Single Point Calculations
+
+Perform a single point calcuation (using the [MACE-MP](https://github.com/ACEsuit/mace-mp) "small" force-field):
 ```shell
-janus singlepoint --struct tests/data/NaCl.cif
+janus singlepoint --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
 This will calculate the energy, stress and forces and save this in `NaCl-results.xyz`, in addition to generating a log file, `singlepoint.log`.
@@ -72,17 +74,22 @@ This will calculate the energy, stress and forces and save this in `NaCl-results
 Additional options may be specified. For example:
 
 ```shell
-janus singlepoint --struct tests/data/NaCl.cif --property energy --arch mace_mp --calc-kwargs "{'model' : './tests/models/mace_mp_small.model'}" --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
+janus singlepoint --struct tests/data/NaCl.cif --arch mace --calc-kwargs "{'model' : '/path/to/your/ml.model'}" --property energy --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
 ```
 
-This defines the MLIP architecture and path to the saved model, as well as changing where the log and results files are saved.
+This defines the MLIP architecture and path to your locally saved model, as well as changing where the log and results files are saved.
 
 Note: the MACE calculator currently returns energy, forces and stress together, so in this case the choice of property will not change the output.
 
+For all options, run `janus singlepoint --help`.
 
-Perform geometry optimization:
+
+### Geometry optimization
+
+Perform geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/mace-mp) "small" force-field):
+
 ```shell
-janus geomopt --struct tests/data/H2O.cif
+janus geomopt --struct tests/data/H2O.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
 This will calculate optimize the atomic positions and save the resulting structure in `H2O-opt.xyz`, in addition to generating a log file, `geomopt.log`.
@@ -90,10 +97,12 @@ This will calculate optimize the atomic positions and save the resulting structu
 Additional options may be specified. This shares most options with `singlepoint`, as well as a few additional options, such as:
 
 ```shell
-janus geomopt --struct tests/data/NaCl.cif --fully-opt --vectors-only --traj 'NaCl-traj.xyz'
+janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --fully-opt --vectors-only --traj 'NaCl-traj.xyz'
 ```
 
 This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
+
+For all options, run `janus geomopt --help`.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@ Tools for machine learnt interatomic potentials
 
 The code relies heavily on ASE, unless something else is mentioned.
 
-
 ## Development
 
 1. Install [poetry](https://python-poetry.org/docs/#installation)
@@ -60,6 +59,41 @@ To prevent poetry downgrading ASE when installing in future, add the commit to p
 ```shell
 poetry add git+https://gitlab.com:ase/ase.git#b31569210d739bd12c8ad2b6ec0290108e049eea
 ```
+
+## Examples
+
+Perform a single point calcuation:
+```shell
+janus singlepoint --struct tests/data/NaCl.cif
+```
+
+This will calculate the energy, stress and forces and save this in `NaCl-results.xyz`, in addition to generating a log file, `singlepoint.log`.
+
+Additional options may be specified. For example:
+
+```shell
+janus singlepoint --struct tests/data/NaCl.cif --property energy --arch mace_mp --calc-kwargs "{'model' : './tests/models/mace_mp_small.model'}" --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
+```
+
+This defines the MLIP architecture and path to the saved model, as well as changing where the log and results files are saved.
+
+Note: the MACE calculator currently returns energy, forces and stress together, so in this case the choice of property will not change the output.
+
+
+Perform geometry optimization:
+```shell
+janus geomopt --struct tests/data/H2O.cif
+```
+
+This will calculate optimize the atomic positions and save the resulting structure in `H2O-opt.xyz`, in addition to generating a log file, `geomopt.log`.
+
+Additional options may be specified. This shares most options with `singlepoint`, as well as a few additional options, such as:
+
+```shell
+janus geomopt --struct tests/data/NaCl.cif --fully-opt --vectors-only --traj 'NaCl-traj.xyz'
+```
+
+This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
 
 ## License
 
--- a/janus_core/cli.py
+++ b/janus_core/cli.py
@@ -6,6 +6,7 @@ from typing import Annotated
 
 import typer
 
+from janus_core.geom_opt import optimize
 from janus_core.single_point import SinglePoint
 
 app = typer.Typer()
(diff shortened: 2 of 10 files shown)


janus_core/cli.py
```diff
@@ -159,13 +168,99 @@ def singlepoint(
 
 
 @app.command()
-def test(name: str):
+def geomopt(  # pylint: disable=too-many-arguments,too-many-locals
```
alinelena, line 171: can we call it geom_opt?

ElliottKasoar:
Left to do:

- [ ] Examples in README

README.md
```diff
@@ -61,6 +60,41 @@ To prevent poetry downgrading ASE when installing in future, add the commit to p
 poetry add git+https://gitlab.com:ase/ase.git#b31569210d739bd12c8ad2b6ec0290108e049eea
 ```
 
+## Examples
+
+Perform a single point calcuation:
```
alinelena, line 65:
```suggestion
Perform a single point calculation(using mace_mp small forcefield):
```

README.md
```diff
@@ -61,6 +60,41 @@ To prevent poetry downgrading ASE when installing in future, add the commit to p
...
+```shell
+janus singlepoint --struct tests/data/NaCl.cif
+```
+
+This will calculate the energy, stress and forces and save this in `NaCl-results.xyz`, in addition to generating a log file, `singlepoint.log`.
+
+Additional options may be specified. For example:
+
+```shell
+janus singlepoint --struct tests/data/NaCl.cif --property energy --arch mace_mp --calc-kwargs "{'model' : './tests/models/mace_mp_small.model'}" --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
```
alinelena, line 75:
```suggestion
janus singlepoint --struct tests/data/NaCl.cif --property energy --arch mace_mp --calc-kwargs "{'model' : '/path/to/your/ml.model'}" --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
```

README.md
```diff
@@ -61,6 +60,41 @@ To prevent poetry downgrading ASE when installing in future, add the commit to p
...
+This will calculate optimize the atomic positions and save the resulting structure in `H2O-opt.xyz`, in addition to generating a log file, `geomopt.log`.
+
+Additional options may be specified. This shares most options with `singlepoint`, as well as a few additional options, such as:
+
+```shell
+janus geomopt --struct tests/data/NaCl.cif --fully-opt --vectors-only --traj 'NaCl-traj.xyz'
+```
+
+This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
+
```
alinelena, line 97:
```suggestion
for all options run **janus geomopt --help**
```

README.md
```diff
@@ -61,6 +60,41 @@ To prevent poetry downgrading ASE when installing in future, add the commit to p
...
+Additional options may be specified. For example:
+
+```shell
+janus singlepoint --struct tests/data/NaCl.cif --property energy --arch mace_mp --calc-kwargs "{'model' : './tests/models/mace_mp_small.model'}" --log './example.log' --write-kwargs "{'filename': './example.xyz'}"
+```
+
+This defines the MLIP architecture and path to the saved model, as well as changing where the log and results files are saved.
+
+Note: the MACE calculator currently returns energy, forces and stress together, so in this case the choice of property will not change the output.
+
```
alinelena, line 81:
```suggestion
for all options run **janus singlepoint --help**

```

janus_core/cli.py
```diff
@@ -151,21 +160,107 @@ def singlepoint(
...
+    ] = 0.1,
+    architecture: Architecture = "mace_mp",
+    device: Device = "cpu",
+    fully_opt: Annotated[
+        bool,
+        typer.Option("--fully-opt", help="Optimise cell and atomic positions"),
+    ] = False,
+    vectors_only: Annotated[
+        bool,
+        typer.Option("--vectors-only", help="Allow only hydrostatic deformations"),
```
alinelena, line 184:
```suggestion
        typer.Option("--vectors-only", help="allow only cell vectors to change"),
```

janus_core/cli.py
```diff
@@ -151,21 +160,107 @@ def singlepoint(
...
+def geomopt(  # pylint: disable=too-many-arguments,too-many-locals
+    struct_path: StructPath,
+    fmax: Annotated[
+        float, typer.Option("--max-force", help="Maximum force for convergence")
+    ] = 0.1,
+    architecture: Architecture = "mace_mp",
+    device: Device = "cpu",
+    fully_opt: Annotated[
+        bool,
+        typer.Option("--fully-opt", help="Optimise cell and atomic positions"),
```
alinelena, line 180:
```suggestion
        typer.Option("--fully-opt", help="fully optimise the cell, vectors and angles."),
```

janus_core/cli.py
```diff
@@ -151,21 +160,107 @@ def singlepoint(
...
+    ] = False,
+    vectors_only: Annotated[
+        bool,
+        typer.Option("--vectors-only", help="Allow only hydrostatic deformations"),
+    ] = False,
+    read_kwargs: ReadKwargs = None,
+    calc_kwargs: CalcKwargs = None,
+    write_kwargs: WriteKwargs = None,
+    traj_file: Annotated[
+        str, typer.Option("--traj", help="Path to save trajectory to")
```
alinelena, line 190:
```suggestion
        str, typer.Option("--traj", help="Path to save optimisation frames to")
```

janus_core/cli.py
```diff
@@ -151,21 +160,107 @@ def singlepoint(
...
+    ] = None,
+    log_file: LogFile = "geomopt.log",
+):
     """
-    Dummy alternative CLI command.
+    Perform geometry optimization and save to file.
```
oerc0122, line 195: Save what to file? Final state? All intermediate states?

janus_core/geom_opt.py
```diff
@@ -60,35 +69,56 @@ def optimize(
...
     opt_kwargs = opt_kwargs if opt_kwargs else {}
-    struct_kwargs = struct_kwargs if struct_kwargs else {}
+    write_kwargs = write_kwargs if write_kwargs else {}
     traj_kwargs = traj_kwargs if traj_kwargs else {}
+    log_kwargs = log_kwargs if log_kwargs else {}
 
-    if struct_kwargs and "filename" not in struct_kwargs:
-        raise ValueError("'filename' must be included in struct_kwargs")
+    if write_kwargs and "filename" not in write_kwargs:
+        raise ValueError("'filename' must be included in `write_kwargs`")
```
oerc0122, line 77:
```suggestion
    if not write_kwargs.get("filename", None):
        raise ValueError("'filename' must be included in `write_kwargs`")
```

oerc0122: Probably want a better check honestly. What is the necessity for a filename?
`not Path(filename).exists()`?

janus_core/geom_opt.py
```diff
@@ -60,35 +69,56 @@ def optimize(
...
     dyn.run(fmax=fmax, **dyn_kwargs)
 
     # Write out optimized structure
-    if struct_kwargs:
-        write(images=atoms, **struct_kwargs)
+    if write_results:
+        if "filename" not in write_kwargs:
+            write_kwargs["filename"] = Path(
+                f"./{atoms.get_chemical_formula()}-opt.xyz"
+            ).absolute()
```
oerc0122, line 113:
```suggestion
        write_kwargs.setdefault("filename",
                                Path(f"./{atoms.get_chemical_formula()}-opt.xyz").absolute()
                                )
```
Though given the check above is this possible to reach?

janus_core/log.py
```diff
@@ -19,6 +19,7 @@ def config_logger(
     filename: Optional[str] = None,
     level: LogLevel = logging.INFO,
     capture_warnings: bool = True,
+    filemode: str = "w",
```
oerc0122, line 22: Should be a `Literal`
```
Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+"]
```

tests/test_geomopt_cli.py
```diff
@@ -0,0 +1,144 @@
...
+            "--struct",
+            DATA_PATH / "NaCl.cif",
+            "--max-force",
+            "0.2",
+        ],
+    )
+    assert result.exit_code == 0
+    atoms = read(results_path)
+    assert isinstance(atoms, Atoms)
+    results_path.unlink()
```
oerc0122, line 42: This will never be reached if the read fails leaving the file extant. Probably want a `with Tempfile` or similar

oerc0122 (changes requested):

README.md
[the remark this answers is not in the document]
alinelena: yap shall be that since we pass the file... also shall be absolute... /path/to/your/mace.model since users will not have that structure

janus_core/cli.py
alinelena: both

ElliottKasoar: I've written it as the optimized structure, as that's the default behaviour, and hopefully it's clear from the options that both are true.

janus_core/log.py
```diff
@@ -19,6 +19,7 @@ def config_logger(
     filename: Optional[str] = None,
     level: LogLevel = logging.INFO,
     capture_warnings: bool = True,
+    filemode: str = Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+"],
```
oerc0122, line 22:
```suggestion
    filemode: Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+"] = "w",
```

janus_core/log.py
```diff
@@ -33,6 +34,8 @@ def config_logger(
         Threshold for logger. Default is logging.INFO.
     capture_warnings : bool
         Whether to capture warnings in log. Default is True.
+    filemode : str
```
oerc0122, line 37:
```suggestion
    filemode : Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+"] 
```

tests/test_geomopt_cli.py
ElliottKasoar: Maybe it's not an essential test, but part of the point of this was to test the default options, which includes a pre-defined path for the optimized structure.

Every other test does use temporary files, but I'm not sure there's a way to do that for this, unless I'm misunderstanding how Tempfile would work?

janus_core/geom_opt.py
ElliottKasoar: What are you wanting it to check that it doesn't currently?

The filename option is admittedly a bit convoluted.

Personally, I'm not a fan of default file names where possible, but I think we decided we'd rather have one.

How it works currently is that if you don't pass `write_kwargs` at all, it'll use the default filename specified below (using `atoms.get_chemical_formula`), but if you're already specifying `write_kwargs`, it requires that to include `filename`.

janus_core/log.py
ElliottKasoar: Spotted that, sorry!

janus_core/geom_opt.py
ElliottKasoar: Changed it now so it'll always allow the default. I'm still not sure there's a perfect solution, but at least its consistent.

tests/test_geomopt_cli.py
oerc0122: I mean, the only obvious way to do this would be:
```
    try:
        assert result.exit_code == 0
        atoms = read(results_path)
        assert isinstance(atoms, Atoms)
    finally:
        results_path.unlink()

alinelena: we were chatting exactly on that line but maybe with a try only around reading

alinelena (approved):

oerc0122 (approved):

PR #76, opened 2024-03-18
Add tests for geometry optimization

### LINKED ISSUE #75: add extra tests for geometry optimisaiton

ElliottKasoar:
Resolves #75

Includes tests for:

- Hydrostatic strain via Python/CLI via new NaCl structure with 89° starting configuration
- Setting the calculator manually rather than via `SinglePoint`

### DIFF
--- a/tests/test_geom_opt.py
+++ b/tests/test_geom_opt.py
@@ -10,6 +10,7 @@ from ase.io import read
 import pytest
 
 from janus_core.geom_opt import optimize
+from janus_core.mlip_calculators import choose_calculator
 from janus_core.single_point import SinglePoint
 
 DATA_PATH = Path(__file__).parent / "data"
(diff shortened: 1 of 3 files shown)


tests/test_geomopt_cli.py
```diff
@@ -124,6 +131,10 @@ def test_fully_opt_and_vectors(tmp_path, caplog):
         assert result.exit_code == 0
         assert "Using filter" in caplog.text
 
+        atoms = read(results_path)
```
alinelena, line 134: do we want read atoms in here?

ElliottKasoar: As in, rather than the utils read/delete file function?

The results are in a `tmp_path`, so just reading them is fine, and it's cleaned up

alinelena: cool

tests/test_geom_opt.py
```diff
@@ -120,3 +121,40 @@ def test_missing_traj_kwarg(tmp_path):
...
+
+    atoms_1 = optimize(
+        single_point_1.struct, filter_kwargs={"hydrostatic_strain": True}
+    )
+    atoms_2 = optimize(
+        single_point_2.struct, filter_kwargs={"hydrostatic_strain": False}
+    )
+
+    expected_1 = [5.69139709, 5.69139709, 5.69139709, 89.0, 90.0, 90.0]
+    expected_2 = [5.68834069, 5.68893345, 5.68932555, 89.75938298, 90.0, 90.0]
```
oerc0122, line 148: What is the tolerance on `pytest.approx`? May want to truncate these to match, otherwise it's somewhat misleading.

ElliottKasoar: It fails if you swap them the test fails - it's within a relative tolerance of `1e-6`.

The test is most significant for the first angle (fourth element), which is very comfortably outside the tolerance, but actually I think the first four elements all trigger the error.

oerc0122: The next question is... do we care about them being within `1e-6`? Or does it just have to have a reasonable fit? Imagine in the future we change default optimiser. This will fail, but it's not physically unreasonable for them to differ by a small amount (certainly not 10^-6 Ang or 10^-6 deg), may `1e-2`, `1e-3`?

alinelena: I will say is good if changing the optimiser the test fails.

ElliottKasoar: Not necessarily, no, but I don't think it's necessarily possible to choose a tolerance that can both distinguish between optimising the same structure in slightly different ways while also being flexible enough to change optimization method completely.

E.g. with `rel=10e-4` only the angle is triggered as an error.

ElliottKasoar: This new structure is an improvement over the normal NaCl.cif file, but in general I think we need to redo the tests more broadly to distinguish results for different options better before we can have a more reasonable tolerance, and even then I wouldn't assume it would transfer to new optimizers.

ElliottKasoar: One issue that could be problematic is that we do indeed already change the default optimizer if you have updated ASE (as we recommend), so while the CI etc. passes with the tagged ASE, it could fail for people developing with newer versions.

Issue #83, opened 2024-03-19
Add options for dyn_kwargs, filter_kwargs and opt_kwargs via CLI


ElliottKasoar:
E.g. setting `dyn_kwargs` to the number of steps to be set. 

Tests should include something to address #82

PR #84, opened 2024-03-20
Update geometry optimization

### LINKED ISSUE #82: add test for geom opt convergence
there needs to be some testing for when the geometry optimisation does not converge in a specified number of steps and raises some error

ElliottKasoar:
Resolves #82 and #83

- Replaces `dyn_kwargs `with `steps` option
  - Originally planned just for the CLI, but I can't see anything that allows any other options to be passed to `dyn.run` other than `steps` and `fmax`, so it seemed sensible to just replace it (we can always add the option again if it proves useful later)
- Adds convergence warning if fmax is not met
- Adds tests for restarting optimization via Python and CLI

To do(?)

- [ ] Store maximum force
  - This doesn't seem to be immediately accessible, as [only the boolean seems to be returned](https://gitlab.com/ase/ase/-/blob/master/ase/utils/abc.py#L32) (note: implemented slightly differently in the last tagged version, but still not stored)
  - We don't need it for the convergence test, so we'd have to calculate additionally anyway
- [ ] traj_kwargs
  - Probably not needed for now? Adding steps was the main motivation for #83

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,6 +39,10 @@ repos:
           docs/.*|
       )$
     entry: pylint
+    args:
+      [
+        "--good-names=e,i,j,k"
+      ]
 
 - repo: https://github.com/numpy/numpydoc
   rev: v1.6.0
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Tools for machine learnt interatomic potentials
   - M3GNET
   - CHGNET
 - [x] Single point calculations
-- [ ] Geometry optimisation
+- [x] Geometry optimisation
 - [ ] Molecular Dynamics
   - NVE
   - NVT (Langevin(Eijnden/Ciccotti flavour) and Nosé-Hoover (Melchionna flavour))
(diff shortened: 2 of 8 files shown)


alinelena:
> Resolves #82 and #83
> 
> * Replaces `dyn_kwargs `with `steps` option
>   
>   * Originally planned just for the CLI, but I can't see anything that allows any other options to be passed to `dyn.run` other than `steps` and `fmax`, so it seemed sensible to just replace it (we can always add the option again if it proves useful later)
> * Adds convergence warning if fmax is not met
> * Adds tests for restarting optimization via Python and CLI
> 
> To do(?)
> 
> * [ ]  Store maximum force
>   
>   * This doesn't seem to be immediately accessible, as [only the boolean seems to be returned](https://gitlab.com/ase/ase/-/blob/master/ase/utils/abc.py#L32) (note: implemented slightly differently in the last tagged version, but still not stored)
>   * We don't need it for the convergence test, so we'd have to calculate additionally anyway
> * [ ]  traj_kwargs
>   
>   * Probably not needed for now? Adding steps was the main motivation for [Add options for dyn_kwargs, filter_kwargs and opt_kwargs via CLI #83](https://github.com/stfc/janus-core/issues/83)

alinelena:
this was closed by my siliness

alinelena:
reopen

ElliottKasoar:
Ah I don't think it will let us since I squashed the typo commits, I'll open a new one

alinelena:
ok, sorry for mistake

PR #87, opened 2024-03-21
Add MD

ElliottKasoar:
Resolves: #15, #16, #17, #18

Currently mostly a skeleton based on tidying up/refactoring the [proof of concept](https://gitlab.com/drFaustroll/lavello/-/blob/main/md.py)

I wouldn't recommend spending too much time reviewing the logic yet, as most things probably don't work quite as expected - this is mostly here so we have an idea of broadly how it will look, rather than what specifically it does.

To do:

- [x] CLI
- [x] Additional tests
- [x] Initial documentation via README
- [x] Check/fix equilibration before starting MD
- [x] Check/fix equilibration during MD
- [x] Check/fix restarting MD including use of `offset`
- [x] Separate writing `md_log` and other trajectory so frequencies are more meaningful
- [x] Consider adding options for what to write out to `md_log` and `traj_file`

Note: constraints and heating will be added separately to this PR. (Heating will use MD but likely be a separate module)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,17 +92,45 @@ Perform geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/ma
 janus geomopt --struct tests/data/H2O.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
-This will calculate optimize the atomic positions and save the resulting structure in `H2O-opt.xyz`, in addition to generating a log file, `geomopt.log`.
+This will optimize the atomic positions and save the resulting structure in `H2O-opt.xyz`, in addition to generating a log file, `geomopt.log`.
 
 Additional options may be specified. This shares most options with `singlepoint`, as well as a few additional options, such as:
 
 ```shell
-janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --fully-opt --vectors-only --traj 'NaCl-traj.xyz'
+janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --vectors-only --traj 'NaCl-traj.xyz'
 ```
 
 This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
 
 For all options, run `janus geomopt --help`.
+
+
+### Molecular dynamics
+
+Run an NPT molecular dynamics simulation (using the [MACE-MP](https://github.com/ACEsuit/mace-mp) "small" force-field) at 300K and 1 bar for 1000 steps (1 ps):
+
+```shell
+janus md npt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --temp 300 --steps 1000 --pressure 1.0
+```
+
+This will generate several output files:
+
+- Thermodynamical statistics every 100 steps, written to `NaCl-npt-T300.0-p1.0-stats.dat`
+- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0--traj.xyz`
+- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0--res-1000.xyz`
+- A log of the processes carried out, written to `md.log`
+
+
+Additional options may be specified. For example:
+
+```shell
+janus md nvt --struct tests/data/NaCl.cif --steps 1000 --timestep 0.5 --temp 300 --minimize --minimize-every 100 --rescale-velocities --remove-rot --rescale-every 100 --equil-steps 200
+```
+
+This performs an NVT molecular dynamics simulation at 300K for 1000 steps (0.5 ps), including performing geometry optimization, rescaling velocities, and removing rotation, both before beginning dynamics and at steps 100 and 200 of the simulation.
+
+For all options, run `janus md --help`.
+
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -116,8 +116,8 @@ janus md npt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model'
 This will generate several output files:
 
 - Thermodynamical statistics every 100 steps, written to `NaCl-npt-T300.0-p1.0-stats.dat`
-- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0--traj.xyz`
-- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0--res-1000.xyz`
+- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0-traj.xyz`
+- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0-res-1000.xyz`
 - A log of the processes carried out, written to `md.log`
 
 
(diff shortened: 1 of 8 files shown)


tests/test_md.py
```diff
@@ -0,0 +1,362 @@
...
+        architecture="mace",
+        calc_kwargs={"model": MODEL_PATH},
+    )
+    npt = NPT(
+        struct=single_point.struct,
+        pressure=1.0,
+        temp=300.0,
+        steps=4,
+        traj_file=traj_path,
+        traj_every=1,
```
alinelena, line 58:
```suggestion
        traj_every=10,
```

alinelena, line 58:
```suggestion
        traj_every=10,
```

tests/test_md.py
```diff
@@ -0,0 +1,362 @@
...
+        lines = md_file.readlines()
+        assert " | T [K]" in lines[0]
+        assert len(lines) == 4
+
+
+def test_nve(tmp_path):
+    """Test NVE molecular dynamics."""
+    restart_stem = tmp_path / "res-Cl4Na4-nve"
+    restart_path = tmp_path / "res-Cl4Na4-nve-300.0-4.xyz"
+    traj_path = tmp_path / "Cl4Na4-nve-300.0-traj.xyz"
```
alinelena, line 123:
```suggestion
    traj_path = tmp_path / "Cl4Na4-nve-T300.0-traj.xyz"
```

tests/test_md.py
```diff
@@ -0,0 +1,362 @@
...
+    md_path = tmp_path / "Cl4Na4-nve-300.0-md.log"
```
alinelena, line 124:
```suggestion
    md_path = tmp_path / "Cl4Na4-nve-T300.0-md.log"
```

ElliottKasoar: Do we want different defaults for the ensembles (other than just `self.ensemble`)?

At the moment we have

```python
if not self.md_file:
    self.md_file = f"{self.struct_name}-{self.ensemble}-{self.temp}-md.log"

if not self.traj_file:
    self.traj_file = f"{self.struct_name}-{self.ensemble}-{self.temp}-traj.xyz"

if not self.restart_stem:
    self.restart_stem = f"res-{self.struct_name}-{self.ensemble}"
```
etc. for all of them. Of course we can add "T" before `self.temp`, but should we instead define a different set of key parameters for each ensemble?

alinelena: if npt you want alse the pressure

ElliottKasoar: Anything else for the others?

Also, any reason not to include T in the restart files? If not, we can use the same label for all of them

alinelena: all files shall have it
also shall be -restart rather than -res

alinelena: the stem shall be the same and then prefix it with -stats.dat for -md.log, -traj and -restart
what was .out in poc shall be the .log but with no extra suffix

ElliottKasoar: > all files shall have it also shall be -restart rather than -res

Sorry, as in include pressure for non-NPT file names too?

alinelena: nope, nve,nvt will have T only npt, nph will have t and p

tests/test_md_cli.py
```diff
@@ -45,8 +48,39 @@ def test_md(ensemble, tmp_path):
...
+                "md",
+                "nvt",
+                "--struct",
+                DATA_PATH / "NaCl.cif",
+                "--temp",
+                300,
+                "--file-prefix",
+                file_prefix,
+                "--steps",
+                2,
```
alinelena, line 82: I will run more like 20 steps and try to check the numbers at the end.

tests/test_md.py
```diff
@@ -0,0 +1,497 @@
...
+            lines = stats_file.readlines()
+            assert " | T [K]" in lines[0]
+            assert len(lines) == 4
+    finally:
+        restart_path.unlink(missing_ok=True)
+        traj_path.unlink(missing_ok=True)
+        stats_path.unlink(missing_ok=True)
+
+
+def test_nve(tmp_path):
```
alinelena, line 133: I will add an extra test where we do not initialise the struicture from a single point but explicitly read the atoms and set the calculator.

README.md
```diff
@@ -92,18 +92,46 @@ Perform geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/ma
...
+- Thermodynamical statistics every 100 steps, written to `NaCl-npt-T300.0-p1.0-stats.dat`
+- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0--traj.xyz`
+- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0--res-1000.xyz`
+- A log of the processes carried out, written to `md.log`
+
+
+Additional options may be specified. For example:
+
+```shell
+janus md nvt --struct tests/data/NaCl.cif --steps 1000 --timestep 0.5 --temp 300 --minimize --minimize-every 100 --rescale-velocities --remove-rot --rescale-every 100 --equil-steps 200
```
alinelena, line 127:
```suggestion
janus md nvt --struct tests/data/NaCl.cif --steps 1000 --timestep 1.0 --temp 300 --minimize --minimize-every 100 --rescale-velocities --remove-rot --rescale-every 100 --equil-steps 200
```

alinelena, line 127: also add an example of npt and one with trajectory writing

tests/test_md_cli.py
```diff
@@ -0,0 +1,108 @@
...
+                "--steps",
+                20,
+                "--stats-every",
+                1,
+            ],
+        )
+        assert result.exit_code == 0
+        assert " Starting molecular dynamics simulation" in caplog.text
+
+        with open(stats_path, encoding="utf8") as stats_file:
```
alinelena, line 91: in long term we shall have an util that reads the log and gives some numpy object that can be analysed after, maybe add an issue about it

README.md
ElliottKasoar: I just wanted to show changing the timestep since 1.0 is the default, so is implicitly the case in the first example

ElliottKasoar: The first example is NPT, and both write out at least one image to the trajectories, although I can show setting the traj_start and traj_every?

README.md
```diff
@@ -92,18 +92,46 @@ Perform geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/ma
...
+Run an NPT molecular dynamics simulation (using the [MACE-MP](https://github.com/ACEsuit/mace-mp) "small" force-field) at 300K and 1 bar for 1000 steps (1 ps):
+
+```shell
+janus md npt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --temp 300 --steps 1000 --pressure 1.0
+```
+
+This will generate several output files:
+
+- Thermodynamical statistics every 100 steps, written to `NaCl-npt-T300.0-p1.0-stats.dat`
+- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0--traj.xyz`
```
alinelena, line 119:
```suggestion
- The structure trajectory every 100 steps, written to `NaCl-npt-T300.0-p1.0-traj.xyz`
```

README.md
```diff
@@ -92,18 +92,46 @@ Perform geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/ma
...
+- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0--res-1000.xyz`
```
alinelena, line 120:
```suggestion
- The structure to be able to restart the dynamics every 1000 steps, written to `NaCl-npt-T300.0-p1.0-res-1000.xyz`
```

README.md
alinelena: ok

janus_core/md.py
```diff
@@ -0,0 +1,874 @@
...
+        if not file_prefix and not isinstance(self, NVT_NH):
+            self.file_prefix = (
+                f"{self.struct_name}-{self.ensemble}-T{self.temp}-p{self.pressure}"
+            )
+            if not kwargs.get("stats_file"):
+                self.stats_file = ""
+            if not kwargs.get("traj_file"):
+                self.traj_file = ""
+            if not kwargs.get("restart_stem"):
+                self.restart_stem = ""
```
oerc0122, line 578: `setdefault` here

oerc0122 (approved):

PR #96, opened 2024-04-03
add md stats reader

alinelena:
add a simple md stats reader, that returns a simple structure.
corrects md header.


alinelena:
needs and extra test for summary function

janus_core/md.py
```diff
@@ -345,7 +345,7 @@ def get_log_header() -> str:
         """
         log_header = (
             "Step | real time[s] | Time [fs] | Epot/N [eV] | Ekin/N [eV] | "
-            "c_T [eV/K] | Etot/N [eV] | Density [g/cm^3] | Volume [A^3] | "
+            "T [K] | Etot/N [eV] | Density [g/cm^3] | Volume [A^3] | "
```
ElliottKasoar, line 348: I think the corresponding `c_T` in `get_log_stats` needs updating as well? (Although surely it's not always valid to use 3/2 kT?)

janus_core/md.py
```diff
@@ -800,7 +800,7 @@ def get_log_header() -> str:
             Header for molecular dynamics log.
         """
         log_header = MolecularDynamics.get_log_header()
-        return log_header + " | T [K]"
+        return log_header + " | T* [K]"
```
ElliottKasoar, line 803: Is this a standard way to represent the target temperature?

janus_core/stats.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Configure shared molecular dynamics simulation options.
+
+    Parameters
+    ----------
+    source : PathLike
+        File that contains the stats of a molecular dynamics simulation.
+    """
+
+    def __init__(self, source: PathLike = None) -> None:
```
ElliottKasoar, line 22: As it is, this should probably be `Optional[PathLike]`, but I'm not sure I see the benefit in it being optional given that the other functions both require a file to have any utility?

ElliottKasoar (changes requested):

janus_core/md.py
alinelena: there is no standard, we have freedom

janus_core/md.py
alinelena: do you mean the name of the variable?, formula is fine for 3d systems...

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+        Parameters
+        ----------
+        filename : PathLike
+            File that contains the stats of a molecular dynamics simulation.
+        """
+        self.data = genfromtxt(filename, skip_header=1)
+        with open(filename, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self.units = [
+                re.search(r"\[.+?\]", x).group(0) if re.search(r"\[.+?\]", x) else ""
```
oerc0122, line 53:
```suggestion
                match[0] if (match := re.search(r"\[.+?\]", x)) else ""
```

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+                for x in head
+            ]
+            self.labels = [
+                re.sub(r"[\[].*?[\]]", "", x).rstrip().lstrip() for x in head
```
oerc0122, line 57:
```suggestion
                re.sub(r"[\[].*?\]", "", x).strip() for x in head
```

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+import re
+
+from numpy import genfromtxt
+
+from janus_core.janus_types import PathLike
+
+
+class Stats:
```
oerc0122, line 12: We should have methods to get names of contained stats. I realise this is in `labels`, but getting a zip of `labels` and `units` may also be handy.
Do we want `__getitem__` definitions to pull out rows? 
Do we want getters for contained stats?

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+        self.columns = 0
+        self.data = None
+        self.labels = None
+        self.units = None
+        self.source = source
+        self.read(source)
+
+    def read(self, filename: PathLike) -> None:
+        """
+        Read MD stats and store them.
```
oerc0122, line 42:
```suggestion
        Read MD stats and store them in `data`.
```

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+        ----------
+        source : PathLike
+            File that contains the stats of a molecular dynamics simulation.
+        """
+
+        self.rows = 0
+        self.columns = 0
+        self.data = None
+        self.labels = None
+        self.units = None
```
oerc0122, line 36: Would these things be better as `@property`s? Do we expect them to be manually set?

janus_core/stats.py
```diff
@@ -0,0 +1,69 @@
...
+        Initialise MD stats reader.
+
+        Parameters
+        ----------
+        source : PathLike
+            File that contains the stats of a molecular dynamics simulation.
+        """
+
+        self.rows = 0
+        self.columns = 0
```
oerc0122, line 33: These should definitely be `@property`s

oerc0122 (changes requested):

janus_core/stats.py
alinelena: that will be nice but not something I have time to do @ElliottKasoar maybe at tomorrow's hackaton? the only issue I have is some of the fields are T* i doubt a getter will like that

janus_core/stats.py
alinelena: yes

janus_core/md.py
ElliottKasoar: > do you mean the name of the variable?, formula is fine for 3d systems...

Yes, it wasn't a particularly good variable name anyway (it required disabling the pylint warning, but I thought it was ok since it matched the header).

(Initially the `c_` and 3/2 factor made me think of heat capacities/internal energy etc. with the different plateaus, but yep it's fine)

janus_core/md.py
ElliottKasoar: Would it be better just be explicit then - `target T` or something, which also would also simplify getting the headers?

janus_core/stats.py
```diff
@@ -0,0 +1,116 @@
...
+        Set number of columns.
+
+        Parameters
+        ----------
+        val_cols : int
+            Number of columns in `data`.
+        """
+        self._columns = val_cols
+
+    def read(self, filename: PathLike) -> None:
```
ElliottKasoar, line 90: Since source is required, we don't need `filename` as a parameter. It would actually be quite confusing to have `Stats(file_1).read(file_2)`

janus_core/md.py
alinelena: in fairness probably we do not need the target at all in that file, since in theory it is in the log file anyhow... maybe is better to get rid of it alltogether, same for pressure.

ElliottKasoar: I suppose right now it doesn't serve much purpose, but when we reintroduce the temperature ramp, it can be useful to have them next to each other rather than having to figure out at which steps it changes. Arguably even then it may not be worth the extra IO though.

alinelena: good point! T_target shall be, same for pressure, p_target

alinelena: since this is a bigger pr that what I did maybe you prepare a separate one dealing with this and then I merge it in this?

janus_core/stats.py
alinelena: done

janus_core/stats.py
alinelena: quite right...

janus_core/stats.py
alinelena: opened an issue about it.

janus_core/stats.py
```diff
@@ -0,0 +1,208 @@
...
+        self.data = genfromtxt(self.source, skip_header=1)
+        with open(self.source, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self.units = [
+                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
+            ]
+            self.labels = [re.sub(r"[\[].*?\]", "", x).strip() for x in head]
+        self.rows, self.columns = self.data.shape
+
+    def summary(self) -> None:
```
oerc0122, line 200: Might actually want to make this the `__repr__` of the data

janus_core/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    def labels(self, val_labels: list[str]) -> None:
+        """
+        Set labels for columns.
+
+        Parameters
+        ----------
+        val_labels : list[str]
+            List of labels for columns in `data`.
+        """
+        self._labels = val_labels
```
oerc0122, line 137: We shouldn't be touching labels or data
```suggestion
```

janus_core/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    def units(self, val_units: list[str]) -> None:
+        """
+        Set units for columns.
+
+        Parameters
+        ----------
+        val_units : list[str]
+            List of units for columns in `data`.
+        """
+        self._units = val_units
```
oerc0122, line 161:
```suggestion
```

janus_core/stats.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Read MD stats and store them in `data`.
+        """
+        self.data = genfromtxt(self.source, skip_header=1)
+        with open(self.source, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self.units = [
+                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
+            ]
+            self.labels = [re.sub(r"[\[].*?\]", "", x).strip() for x in head]
```
oerc0122, line 197:
```suggestion
        self._data = genfromtxt(self.source, skip_header=1)
        with open(self.source, "r+", encoding="utf-8") as file:
            head = file.readline().split("|")
            self._units = [
                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
            ]
            self._labels = [re.sub(r"[\[].*?\]", "", x).strip() for x in head]
```

janus_core/stats.py
```diff
@@ -0,0 +1,208 @@
...
+        ----------
+        source : PathLike
+            File that contains the stats of a molecular dynamics simulation.
+        """
+
+        self._rows = 0
+        self._columns = 0
+        self._data = None
+        self._labels = None
+        self._units = None
```
oerc0122, line 39:
```suggestion
        self._data = np.zeros(0, 0)
        self._labels = ()
        self._units = ()
```

oerc0122 (changes requested):

janus_core/stats.py
alinelena: done

tests/test_stats.py
```diff
@@ -0,0 +1,33 @@
...
+
+    assert stat_data.rows == 100
+    assert stat_data.columns == 18
+    assert stat_data.data[99, 17] == approx(300.0)
+    assert stat_data.units[0] == ""
+    assert stat_data.units[17] == "[K]"
+    assert stat_data.labels[0] == "Step"
+    assert stat_data.labels[17] == "T*"
+
+    print(stat_data)
```
oerc0122, line 26: Debugging statement?

janus_core/stats.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Read MD stats and store them in `data`.
+        """
+        self._data = genfromtxt(self.source, skip_header=1)
+        with open(self.source, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self._units = [
+                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
+            ]
+            self._labels = [re.sub(r"[\[].*?\]", "", x).strip() for x in head]
```
oerc0122, line 123: Since we want these fixed, might be safer as:
```suggestion
            self._units = tuple(
                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
            )
            self._labels = tuple(re.sub(r"[\[].*?\]", "", x).strip() for x in head)
```
But really not all that important.

janus_core/stats.py
```diff
@@ -0,0 +1,139 @@
...
+    def units(self) -> list[str]:
+        """
+        Return a list of units for the columns in `data`.
+
+        Returns
+        -------
+        list[str]
+            List of units for the columns in `data`.
+        """
+        return self._units
```
oerc0122, line 99: If you do go with the change below:
```suggestion
    @property
    def labels(self) -> tuple[str, ...]:
        """
        Return a list of labels for the columns in `data`.

        Returns
        -------
        tuple[str, ...]
            List of labels for the columns in `data`.
        """
        return self._labels

    @property
    def units(self) -> tuple[str, ...]:
        """
        Return a list of units for the columns in `data`.

        Returns
        -------
        tuple[str, ...]
            List of units for the columns in `data`.
        """
        return self._units
```

oerc0122 (dismissed):
Apart from a print statement and a minor stylistic choice, this is ready to go.

tests/test_stats.py
alinelena: I am asserting it below.

oerc0122: Ok, didn't notice the `capsys`

oerc0122 (dismissed):

janus_core/md.py
```diff
@@ -799,6 +799,7 @@ def get_log_header() -> str:
             Header for molecular dynamics log.
         """
         log_header = MolecularDynamics.get_log_header()
+
```
oerc0122, line 802: Did you mean to add this?

tests/test_stats.py
```diff
@@ -0,0 +1,33 @@
...
+
+    stat_data = Stats(data_path)
+
+    assert stat_data.rows == 100
+    assert stat_data.columns == 18
+    assert stat_data.data[99, 17] == approx(300.0)
+    assert stat_data.units[0] == ""
+    assert stat_data.units[17] == "[K]"
+    assert stat_data.labels[0] == "Step"
+    assert stat_data.labels[17] == "T*"
```
ElliottKasoar, line 24:
```suggestion
    assert stat_data.labels[17] == "Target T"
```

janus_core/stats.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Read MD stats and store them in `data`.
+        """
+        self._data = genfromtxt(self.source, skip_header=1)
+        with open(self.source, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self._units = tuple(
+                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
+            )
+            self._labels = tuple(re.sub(r"[\[].*?\]", "", x).strip() for x in head)
```
oerc0122, line 123:
```suggestion
            self._labels = tuple(re.sub(r"\[.*?\]", "", x).strip() for x in head)
```

janus_core/stats.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def read(self) -> None:
+        """
+        Read MD stats and store them in `data`.
+        """
+        self._data = genfromtxt(self.source, skip_header=1)
+        with open(self.source, "r+", encoding="utf-8") as file:
+            head = file.readline().split("|")
+            self._units = tuple(
+                match[0] if (match := re.search(r"\[.+?\]", x)) else "" for x in head
```
oerc0122, line 121:
```suggestion
                match[1] if (match := re.search(r"\[(.+?)\]", x)) else "" for x in head
```

tests/data/md-stats.dat
```diff
@@ -0,0 +1,101 @@
+Step | real time[s] | Time [fs] | Epot/N [eV] | Ekin/N [eV] | T [K] | Etot/N [eV] | Density [g/cm^3] | Volume [A^3] | Pressure [bar] | Pxx [bar] | Pyy [bar] | Pzz[bar] | Pyz[bar] | Pxz[bar] | Pxy[bar] | Pressure[bar] | T* [K]
```
ElliottKasoar, line 1:
```suggestion
Step | Real Time [s] | Time [fs] | Epot/N [eV] | Ekin/N [eV] | T [K] | Etot/N [eV] | Density [g/cm^3] | Volume [A^3] | P [bar] | Pxx [bar] | Pyy [bar] | Pzz [bar] | Pyz [bar] | Pxz [bar] | Pxy [bar] | Target P [bar] | Target T [K]
```

ElliottKasoar (changes requested):

janus_core/md.py
alinelena: no very odd

alinelena: fixed

ElliottKasoar (approved):

oerc0122 (approved):

PR #98, opened 2024-04-04
Update stats headers

ElliottKasoar:
- Clarifies pressure and temperature headers are target values (will be useful when they may vary across a simulation)
- Tidies up code for header and variable name consistency

### DIFF
--- a/tests/test_md_cli.py
+++ b/tests/test_md_cli.py
@@ -103,7 +103,7 @@ def test_md_log(tmp_path, caplog):
             assert init_volume == pytest.approx(final_volume)
 
             # Test constant temperature
-            assert lines[0].split(" | ")[16] == "T [K]\n"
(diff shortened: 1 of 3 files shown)


tests/test_md_cli.py
```diff
@@ -103,7 +103,7 @@ def test_md_log(tmp_path, caplog):
             assert init_volume == pytest.approx(final_volume)
 
             # Test constant temperature
-            assert lines[0].split(" | ")[16] == "T [K]\n"
+            assert lines[0].split(" | ")[16] == "Target T [K]\n"
             init_temp = float(lines[1].split()[16])
             final_temp = float(lines[-1].split()[16])
             assert init_temp == 300.0
```
alinelena, line 109: just a note, all ok otherwise, once stats is merged probably will be cool to use that in testing...

alinelena (approved):

PR #99, opened 2024-04-04
Fix MACE model

ElliottKasoar:
This will address errors with MACE-OFF due to a change of URL, but currently only fixes a misleading error if no model is specified for "mace" (`TypeError: object of type 'NoneType' has no len()` -> `ValueError: No mace file names supplied`)

### DIFF
--- a/janus_core/mlip_calculators.py
+++ b/janus_core/mlip_calculators.py
@@ -51,7 +51,7 @@ def choose_calculator(
         from mace.calculators import MACECalculator
 
         # `model_paths` is keyword for path to model, so take from kwargs if specified
-        # Otherwise, take `model` if specified, then default to `None`, which will
+        # Otherwise, take `model` if specified, then default to `""`, which will
         # raise a ValueError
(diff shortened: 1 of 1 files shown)


ElliottKasoar:
Changes to the [URLs](https://github.com/ACEsuit/mace/blob/main/mace/calculators/foundations_models.py#L159) have been reverted, so looks like no further changes are needed.

janus_core/mlip_calculators.py
```diff
@@ -53,7 +53,7 @@ def choose_calculator(
         # `model_paths` is keyword for path to model, so take from kwargs if specified
         # Otherwise, take `model` if specified, then default to `None`, which will
```
ElliottKasoar, line 54:
```suggestion
        # Otherwise, take `model` if specified, then default to `""`, which will
```

ElliottKasoar:
There's a minor fix for the normal "mace"

alinelena (approved):

PR #100, opened 2024-04-05
Add config reader for CLI

### LINKED ISSUE #92: Read CLI options from config file
Most likely a yaml config file

ElliottKasoar:
Resolves #92

- Uses `typer_config` to allow a config file to be passed, but overwritten with explicit command line options.
- Fixes TyperDict to allow Python dictionaries as well as dictionaries wrapped in strings as was needed to parse dictionaries via the CLI

To do:

- [x] Add tests for geomopt/md config
- [x] Add an example to the README


README.md
```diff
@@ -140,6 +140,38 @@ This performs an NVE molecular dynamics simulation at 300K for 200 steps (0.2 ps
 For all options, run `janus md --help`.
 
 
+### Using configuration files
+
+Default values for all commandline options may be specifed through a Yaml 1.1 formatted configuration file by adding the `--config` option.
```
alinelena, line 145:
```suggestion
Default values for all command line options may be specifed through a Yaml 1.1 formatted configuration file by adding the `--config` option. If an option is present on both command line and config file, the command line one takes precedence.
```

README.md
```diff
@@ -140,6 +140,38 @@ This performs an NVE molecular dynamics simulation at 300K for 200 steps (0.2 ps
...
+```
+
+This will run a singlepoint energy calculation on `KCl.cif` using the [MACE-MP](https://github.com/ACEsuit/mace-mp) "medium" force-field, saving the results to `KCl-results.cif`.
+
+
+> [!NOTE]
+> `properties` must be passed as a Yaml list, as above, not as a string.
+
+> [!WARNING]
+> Options in the Yaml file must match their Python variable name, which may differ from the option flag.
```
alinelena, line 170: this is not cool... we need to think.

ElliottKasoar: Yes it's not ideal. The obvious solution is just make them all match

janus_core/cli.py
```diff
@@ -90,7 +93,10 @@ def _parse_typer_dicts(typer_dicts: list[TyperDict]) -> list[dict]:
     for i, typer_dict in enumerate(typer_dicts):
         typer_dicts[i] = typer_dict.value if typer_dict else {}
         if not isinstance(typer_dicts[i], dict):
-            raise ValueError(f"{typer_dicts[i]} must be passed as a dictionary")
+            raise ValueError(
+                f"""{typer_dicts[i]} must be passed as a dictionary wrapped in quotes.\
+ For example, "{{'key' : value}}" """
+            )
```
oerc0122, line 99: I usually prefer separating error from process if possible
```suggestion
if any(not isinstance(td.value, dict) for td in typer_dicts if td):
    raise ValueError(
                f"""Dicts must be passed as a dictionary wrapped in quotes.\
 For example, "{{'key' : value}}" """
 
typer_dicts = [typer_dict.value if typer_dict else {} for typer_dict in typer_dicts]
```

README.md
alinelena: yap i think that is the solution otherwise too much confusion.

ElliottKasoar: There will still be a `-` vs `_` difference, but probably close enough otherwise.

It would be much less of an issue if it didn't allow unspecified options too, so I need to figure out if that's fixable too

oerc0122 (approved):
Looks good apart from one minor stylistic choice, I have no problems if it works.

janus_core/cli.py
ElliottKasoar: Only problem with this solution is that it doesn't tell you which (non-)dict is causing the error.

oerc0122: That is true, but could also do a find in case of failure? Could also do a filter and iterate through. I don't know. How useful is printing that "non-dict"?

oerc0122: Also like I say, what's there works, and I'm happy with it

README.md
```diff
@@ -140,6 +140,41 @@ This performs an NVE molecular dynamics simulation at 300K for 200 steps (0.2 ps
...
 
+### Using configuration files
+
+Default values for all command line options may be specifed through a Yaml 1.1 formatted configuration file by adding the `--config` option. If an option is present in both the command line and configuration file, the command line value takes precedence.
+
+For example, with the following configuration file and command:
+
+```yaml
+struct: "NaCl.cif"
+properties:
```
alinelena, line 151: ? based on another comment is this correct?

ElliottKasoar: It's no longer correct (it doesn't work very well defining the path here for tests while keeping it generic and easily removable) but I kept it as something to be overwritten by the explicit `--struct`.

Maybe I should make it more explicitly wrong.

ElliottKasoar: Oh sorry I was thinking this is the test config, which bit is the problem?

ElliottKasoar: `janus singlepoint --config config.yml` works with this config if `NaCl.cif` is in your current directory, so perhaps the comment is wrong/misleading?

janus_core/cli.py
ElliottKasoar: Think I'll leave this as is just to be as explicit as possible with the cause of the error, given that I can imagine passing dictionaries incorrectly is likely to be one of the more common issues.

alinelena (approved):

PR #111, opened 2024-04-10
smaller logo than before

alinelena:


ElliottKasoar (changes requested):
I think this is missing the new image.

ElliottKasoar (approved):

Issue #108, opened 2024-04-10
Fix restart


ElliottKasoar:
Restart correctly identifies the step from the data file, but I'm not sure it actually use the restart file to set the structure.

We can probably update the step-reading to use the data loader too.

Restart also needs some reworking for heating from #107

harveydevereux:
Hi @ElliottKasoar, sorry to piggy back on this issue, but I was just thinking about restart and heating, when designing a workflow using the heating. 

It seems that restart loads up the data [file as you say but not explicitly the ```.xyz``` that gets written by ```_write_restart```?](https://github.com/stfc/janus-core/blob/8b86a8749230f9ae7ae6a5362ebf368f0d0f71e7/janus_core/md.py#L474C4-L501C78)

My assumption now was if I pass in ```--struct NAME_OF_LAST_RESTART_FILE --stats-file NAME_OF_LAST_DATA_FILE```, then janus would truely restart the sim, as I have explicitly set the structure? Which is a way round the issue.

@alinelena and I were just discussing also the possibility of having a "final" restart file that has a known filename to aid automation of workflows.

Currently If I change the final temperature any follow on workflow will have to be changed to account for the new file name that depends on the step (I think). And it may not be the final step either if the user does not set the interval correctly. [e.g.](https://github.com/stfc/janus-core/blob/8b86a8749230f9ae7ae6a5362ebf368f0d0f71e7/janus_core/md.py#L478C13-L478C61)```restart_file = f"{self.restart_stem}-{step}.xyz"```.

I might even suggest we could have an option to explicitly override the restart file name, so we always know it would be say ```restart_file = f"{self.SOME_USER_SPECIFIED_NAME}.xyz"```, even though that means overwriting it.

I'd try to do it now, but currently catching up on how the attachments work ```self.dyn.attach(self._write_restart, interval=self.restart_every)```.

Not sure what the best course of action is.

ElliottKasoar:
Actually revisiting this, it's while it may not be entirely intuitive, I think it currently works as I originally intended for normal MD.

In the original [proof of concept](https://gitlab.com/drFaustroll/lavello/-/blob/main/md.py), there's a separate auto restart option, which I didn't feel essential for the initial implementation. It's probably worth adding that option back in, but I'm not sure it's necessarily a problem that by default it relies on the user specifying the correct files.


[...]

PR #116, opened 2024-04-18
Add final restart files, and fixed heating file stems

### LINKED ISSUE #115: Add final restart file at terminal heating points/ md runs
As discussed in #108 the addition is to dump additional ```xyz``` files at the end of heating steps, and md runs

harveydevereux:
The desired behaviour introduced is to produce a restart file (```.xyz```) at the end of a heating cycle or md run. As discussed in #108 the user specified restart frequency may not align with the final time step, and this results in a predictable file name for use in automated meta-workflows.

Schematically the file stems will be (NPT is currently no working in this PR),

```
# NVT/NVE, not heating
[ENSEMBLE]-[TEMPERATURE]-[STEP]-res
[ENSEMBLE]-[TEMPERATURE]-final
# heating, where TEMPERATURE is dynamically updated as heating progresses
[ENSEMBLE]-[TEMPERATURE]-[STEP]-res
[ENSEMBLE]-[TEMPERATURE]-final
[ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-stats
[ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-traj
# NPT also has pressure, after termperature, e.g.
[ENSEMBLE]-[TEMPERATURE]-[PRESSURE]-final
[ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-[PRESSURE]-stats
```

@ElliottKasoar One point that definitely needs addressing is what is desired when we do a heating run and md run, in the same janus run? Does there need to be some indication the md run is post heating for example?

This PR Closes #115

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 """Run molecular dynamics simulations."""
 
 import datetime
@@ -360,36 +361,73 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
                 self.logger.info("Minimizing at step %s", self.dyn.nsteps)
             optimize(self.struct, **self.minimize_kwargs)
 
+    @property
     def _parameter_prefix(self) -> str:
         """
-        Ouput additional ensemble parameters for file stems.
+        Ensemble parameters for output files.
 
         Returns
         -------
         str
-            Additional ensemble parameters written to a str.
+           Formatted temperature range if heating and target temp if running md.
         """
-        return ""
+
+        temperature_prefix = ""
+        if self.temp_start and self.temp_end:
+            temperature_prefix = f"-T{self.temp_start}-T{self.temp_end}"
+
+        if self.steps > 0:
+            temperature_prefix = f"{temperature_prefix}-T{self.temp}"
+
+        return temperature_prefix
+
+    @property
+    def _final_file(self) -> str:
+        """
+        Final state file name.
+
+        Returns
+        -------
+        str
+           File name for final state.
+        """
+
+        if not self.restart_stem:
+            return f"{self.struct_name}-{self.ensemble}-T{self.temp}-final.xyz"
+        # respect the users choice
+        return f"{self.restart_stem}-T{self.temp}-final.xyz"
+
+    @property
+    def _restart_file(self) -> str:
+        """
+        Restart file name.
+
+        Returns
+        -------
+        str
+           File name for restart files.
+        """
+        step = self.offset + self.dyn.nsteps
+        if not self.restart_stem:
+            return f"{self.struct_name}-{self.ensemble}-T{self.temp}-res-{step}.xyz"
+        return f"{self.restart_stem}-T{self.temp}-res-{step}.xyz"
 
     def configure_filenames(self) -> None:
         """Setup filenames for output files."""
 
-        temp = ""
         if not self.file_prefix:
             self.file_prefix = (
-                f"{self.struct_name}-{self.ensemble}{self._parameter_prefix()}"
+                f"{self.struct_name}-{self.ensemble}{self._parameter_prefix}"
             )
-            if self.temp_start and self.temp_end:
-                temp = f"-T{self.temp_start}-T{self.temp_end}"
-
-            if self.steps > 0:
-                temp = f"{temp}-T{self.temp}"
+        else:
+            if not self.restart_stem:
+                self.restart_stem = f"{self.file_prefix}"
 
         if not self.stats_file:
-            self.stats_file = f"{self.file_prefix}{temp}-stats.dat"
+            self.stats_file = f"{self.file_prefix}-stats.dat"
 
         if not self.traj_file:
-            self.traj_file = f"{self.file_prefix}{temp}-traj.xyz"
+            self.traj_file = f"{self.file_prefix}-traj.xyz"
 
     @staticmethod
     def get_log_header() -> str:
@@ -490,26 +528,25 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
 
     def _write_final_state(self) -> None:
         """Write the final system state."""
-        if not self.restart_stem:
-            restart_file = f"{self.file_prefix}-T{self.temp}-final.xyz"
-        write(restart_file, self.struct, write_info=True, columns=None)
+        write(
+            self._final_file,
+            self.struct,
+            write_info=True,
+            columns=["symbols", "positions", "momenta", "masses"],
+        )
 
     def _write_restart(self) -> None:
         """Write restart file and (optionally) rotate files saved."""
         step = self.offset + self.dyn.nsteps
         if step > 0:
-            if not self.restart_stem:
-                restart_file = f"{self.file_prefix}-T{self.temp}-res-{step}.xyz"
-            else:
-                restart_file = f"{self.restart_stem}-res-{step}.xyz"
             write(
-                restart_file,
+                self._restart_file,
                 self.struct,
                 write_info=True,
                 columns=["symbols", "positions", "momenta", "masses"],
             )
             if self.rotate_restart:
-                self.restart_files.append(restart_file)
+                self.restart_files.append(self._restart_file)
                 self._rotate_restart_files()
 
     def run(self) -> None:
@@ -671,16 +708,61 @@ class NPT(MolecularDynamics):
             externalstress=self.pressure * units.bar,
         )
 
+    @property
     def _parameter_prefix(self) -> str:
         """
-        Ouput additional ensemble parameters for file stems.
+        Ensemble parameters for output files.
 
         Returns
         -------
         str
-            Additional ensemble parameters written to a str.
+           Formatted temperature range if heating and target temp if running md.
         """
-        return f"-p{self.pressure}"
+
+        temperature_prefix = ""
+        if self.temp_start and self.temp_end:
+            temperature_prefix = f"-T{self.temp_start}-T{self.temp_end}"
+
+        if self.steps > 0:
+            temperature_prefix = f"{temperature_prefix}-T{self.temp}"
+
+        pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
+        return f"{pressure}{temperature_prefix}"
+
+    @property
+    def _final_file(self) -> str:
+        """
+        Final state file name.
+
+        Returns
+        -------
+        str
+           File name for final state, includes pressure.
+        """
+
+        pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
+        if not self.restart_stem:
+            return (
+                f"{self.struct_name}-{self.ensemble}{pressure}-T{self.temp}-final.xyz"
+            )
+        return f"{self.restart_stem}{pressure}-T{self.temp}-final.xyz"
+
+    @property
+    def _restart_file(self) -> str:
+        """
+        Restart file name.
+
+        Returns
+        -------
+        str
+           File name for restart file, includes pressure.
+        """
+        step = self.offset + self.dyn.nsteps
+        pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
+        ending = f"-T{self.temp}-res-{step}.xyz"
+        if not self.restart_stem:
+            return f"{self.struct_name}-{self.ensemble}{pressure}{ending}"
+        return f"{self.restart_stem}{pressure}{ending}"
 
     def get_log_stats(self) -> str:
         """
@@ -870,17 +952,6 @@ class NVT_NH(NPT):  # pylint: disable=invalid-name
             *args,
             **kwargs,
         )
-
-    def _parameter_prefix(self) -> str:
-        """
-        Ouput additional ensemble parameters for file stems.
-
-        Returns
-        -------
-        str
-            Additional ensemble parameters written to a str.
-        """
-        return ""
 
     def get_log_stats(self) -> str:
         """
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -377,7 +377,7 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
             temperature_prefix = f"-T{self.temp_start}-T{self.temp_end}"
 
         if self.steps > 0:
-            temperature_prefix = f"{temperature_prefix}-T{self.temp}"
+            temperature_prefix += f"-T{self.temp}"
 
         return temperature_prefix
 
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -523,7 +523,6 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
         self.dyn.attach(self._write_stats_file, interval=self.stats_every)
         self.dyn.attach(self._write_traj, interval=self.traj_every)
         self.dyn.attach(self._write_restart, interval=self.restart_every)
-        self.dyn.attach(lambda : self._write_restart("final.xyz"), interval=self.steps)
 
         if self.rescale_velocities:
             self.dyn.attach(self._reset_velocities, interval=self.rescale_every)
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -393,7 +393,7 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
         """
 
         if not self.restart_stem:
-            return f"{self.struct_name}-{self.ensemble}-T{self.temp}-final.xyz"
+            return f"{self.file_prefix}-T{self.temp}-final.xyz"
         # respect the users choice
         return f"{self.restart_stem}-T{self.temp}-final.xyz"
 
@@ -409,25 +409,25 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
         """
         step = self.offset + self.dyn.nsteps
         if not self.restart_stem:
-            return f"{self.struct_name}-{self.ensemble}-T{self.temp}-res-{step}.xyz"
+            return f"{self.file_prefix}-T{self.temp}-res-{step}.xyz"
         return f"{self.restart_stem}-T{self.temp}-res-{step}.xyz"
 
     def configure_filenames(self) -> None:
         """Setup filenames for output files."""
 
         if not self.file_prefix:
-            self.file_prefix = (
-                f"{self.struct_name}-{self.ensemble}{self._parameter_prefix}"
-            )
+            self.file_prefix = f"{self.struct_name}-{self.ensemble}"
+            data_prefix = f"{self.file_prefix}{self._parameter_prefix}"
         else:
+            data_prefix = f"{self.file_prefix}"
             if not self.restart_stem:
                 self.restart_stem = f"{self.file_prefix}"
 
         if not self.stats_file:
-            self.stats_file = f"{self.file_prefix}-stats.dat"
+            self.stats_file = f"{data_prefix}-stats.dat"
 
         if not self.traj_file:
-            self.traj_file = f"{self.file_prefix}-traj.xyz"
+            self.traj_file = f"{data_prefix}-traj.xyz"
 
     @staticmethod
     def get_log_header() -> str:
@@ -735,9 +735,7 @@ class NPT(MolecularDynamics):
 
         pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
         if not self.restart_stem:
-            return (
-                f"{self.struct_name}-{self.ensemble}-T{self.temp}{pressure}-final.xyz"
-            )
+            return f"{self.file_prefix}-T{self.temp}{pressure}-final.xyz"
         return f"{self.restart_stem}-T{self.temp}{pressure}-final.xyz"
 
     @property
@@ -752,10 +750,9 @@ class NPT(MolecularDynamics):
         """
         step = self.offset + self.dyn.nsteps
         pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
-        ending = f"-T{self.temp}{pressure}-res-{step}.xyz"
         if not self.restart_stem:
-            return f"{self.struct_name}-{self.ensemble}{ending}"
-        return f"{self.restart_stem}{ending}"
+            return f"{self.file_prefix}-T{self.temp}{pressure}-res-{step}.xyz"
+        return f"{self.restart_stem}-T{self.temp}{pressure}-res-{step}.xyz"
 
     def get_log_stats(self) -> str:
         """
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -360,23 +360,37 @@ class MolecularDynamics:  # pylint: disable=too-many-instance-attributes
                 self.logger.info("Minimizing at step %s", self.dyn.nsteps)
             optimize(self.struct, **self.minimize_kwargs)
 
-    def configure_filenames(self, pressure: Optional[float] = None) -> None:
+    def _parameter_prefix(self) -> str:
+        """
+        Ouput additional ensemble parameters for file stems.
+
+        Returns
+        -------
+        str
+            Additional ensemble parameters written to a str.
+        """
+        return ""
+
+    def configure_filenames(self) -> None:
         """Setup filenames for output files."""
+
         temp = ""
         if not self.file_prefix:
-            self.file_prefix = f"{self.struct_name}-{self.ensemble}"
-            if pressure is not None:
-                 self.file_prefix = f"{self.file_prefix}-p{pressure}" 
+            self.file_prefix = (
+                f"{self.struct_name}-{self.ensemble}{self._parameter_prefix()}"
+            )
             if self.temp_start and self.temp_end:
-               temp = f"-T{self.temp_start}-T{self.temp_end}"
-            else:
-               temp = f"-T{self.temp}"
+                temp = f"-T{self.temp_start}-T{self.temp_end}"
+
+            if self.steps > 0:
+                temp = f"{temp}-T{self.temp}"
 
         if not self.stats_file:
             self.stats_file = f"{self.file_prefix}{temp}-stats.dat"
 
         if not self.traj_file:
             self.traj_file = f"{self.file_prefix}{temp}-traj.xyz"
+
     @staticmethod
     def get_log_header() -> str:
         """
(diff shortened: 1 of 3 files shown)


harveydevereux:
One thought maybe instead of ```-final``` we put the final timestep ```[FINAL_STEP]-res```?

alinelena:
can be but be careful cause STEP and FINAL_STEP may be the same and you do not want to write twice...

alinelena:
you can have a check to write it only if it does not exist already

harveydevereux:
Tests now pass, with some modifications since pressure and temperature have been permuted, and also since restart now adds temperature.

Will add tests for the heating and final restart files next

ElliottKasoar:
> The desired behaviour introduced is to produce a restart file (`.xyz`) at the end of a heating cycle or md run. As discussed in #108 the user specified restart frequency may not align with the final time step, and this results in a predictable file name for use in automated meta-workflows.
> 
> Schematically the file stems will be (NPT is currently no working in this PR),
> 
> ```
> # NVT/NVE, not heating
> [ENSEMBLE]-[TEMPERATURE]-[STEP]-res
> [ENSEMBLE]-[TEMPERATURE]-final
> # heating, where TEMPERATURE is dynamically updated as heating progresses
> [ENSEMBLE]-[TEMPERATURE]-[STEP]-res
> [ENSEMBLE]-[TEMPERATURE]-final
> [ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-stats
> [ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-traj
> # NPT also has pressure, after termperature, e.g.
> [ENSEMBLE]-[TEMPERATURE]-[PRESSURE]-final
> [ENSEMBLE]-[TEMPERATURE_START]-[TEMPERATURE_END]-[PRESSURE]-stats
> ```
> 
> @ElliottKasoar One point that definitely needs addressing is what is desired when we do a heating run and md run, in the same janus run? Does there need to be some indication the md run is post heating for example?
> 
> This PR Closes #115

At the moment these changes seems to output the "final" structure at every temperature in the ramp, not just at the end of heating? Was this intended?

alinelena:
> At the moment these changes seems to output the "final" structure at every temperature in the ramp, not just at the end of heating? Was this intended?
yes, I have asked for it. we may consider to make it optional but for the moment I will say keep it like that.

ElliottKasoar:
> > At the moment these changes seems to output the "final" structure at every temperature in the ramp, not just at the end of heating? Was this intended?

>  yes, I have asked for it. we may consider to make it optional but for the moment I will say keep it like that.

What's the benefit in all of these extra files, in addition to the periodic restart files (and trajectory)?

I see the benefit in storing a final structure at the end of whatever simulation has run, as that might be missed from outputs due to when things happen to line up with frequencies, but for anything intermediate, what's the advantage to storing these in addition to the restart files?

alinelena:
> > yes, I have asked for it. we may consider to make it optional but for the moment I will say keep it like that.
> 
> What's the benefit in all of these extra files, in addition to the periodic restart files (and trajectory)?
> 
> I see the benefit in storing a final structure at the end of whatever simulation has run, as that might be missed from outputs due to when things happen to line up with frequencies, but for anything intermediate, what's the advantage to storing these in addition to the restart files?

if you want to do a campaign and simulate systems at various T you want configurations consistent with that T. final, shall be saved only if the restart does not coincide with the step at which is write a restart file.

ElliottKasoar:
> if you want to do a campaign and simulate systems at various T you want configurations consistent with that T. 

Ah ok, so the purpose isn't really a restart file in the same sense as we had previously. I think that makes sense though, although I'm not sure it should necessarily be the default behaviour.

> final, shall be saved only if the restart does not coincide with the step at which is write a restart file.

I don't think we currently prevent this?

Also worth noting that the final and restart files currently store the data with slightly different formats (e.g. final seems to include `spacegroup_kinds`).

janus_core/md.py
```diff
@@ -471,18 +474,25 @@ def _write_traj(self) -> None:
...
         if step > 0:
-            restart_file = f"{self.restart_stem}-{step}.xyz"
+            if not self.restart_stem:
+                if not final:
+                    restart_file = f"{self.file_prefix}-T{self.temp}-res-{step}.xyz" 
+                else:
+                    restart_file = f"{self.file_prefix}-T{self.temp}-final.xyz"
+            else:
+                restart_file = f"{self.restart_stem}-res-{step}.xyz"
+            columns = ["symbols", "positions", "momenta", "masses"] if not final else None
```
ElliottKasoar, line 488: Why do we want `final` and `not final` to be save different things?

janus_core/md.py
```diff
@@ -360,20 +360,23 @@ def _optimize_structure(self) -> None:
                 self.logger.info("Minimizing at step %s", self.dyn.nsteps)
             optimize(self.struct, **self.minimize_kwargs)
 
-    def configure_filenames(self) -> None:
+    def configure_filenames(self, pressure: Optional[float] = None) -> None:
```
ElliottKasoar, line 363: Is there a way to do this without explicitly passing pressure? Given that pressure can be an attribute, it might be nice if it could be used directly rather than passing it around, but maybe this would end up messier than it needs to.

janus_core/md.py
alinelena: mainly because one is restart the other one not a restart, the issue of extra info printed is an ase one I will not be too bothered about...

harveydevereux:
> Can you update the README with the changes to the output files?

will do

janus_core/md.py
```diff
@@ -360,20 +360,23 @@ def _optimize_structure(self) -> None:
...
         """Setup filenames for output files."""
+        temp = ""
         if not self.file_prefix:
-            self.file_prefix = f"{self.struct_name}-{self.ensemble}-T{self.temp}"
+            self.file_prefix = f"{self.struct_name}-{self.ensemble}"
+            if pressure is not None:
+                 self.file_prefix = f"{self.file_prefix}-p{pressure}" 
+            if self.temp_start and self.temp_end:
+               temp = f"-T{self.temp_start}-T{self.temp_end}"
```
ElliottKasoar, line 371: I think I might have had this conversation already with @alinelena, but if the MD temperature is different to the final temperature, the MD temperature won't appear in the file name. This is probably fine, but just checking this is intentional.

I was initially going to suggest we might want to include the MD temperature if it differs from the final temperature, but actually I wonder if it makes sense to include anyway, to distinguish between simulations with and without MD at the end of heating?

janus_core/md.py
harveydevereux: i think we could if configure_filenames is overridden in NPT, which would duplicate most of the code. 

Or if in configure_filenames we check the type and if it is NPT we add the pressure.

We could also have a function, say ```self.parameters()``` which would format the parameters (T, P) for the file names in which case the NPT would override to add the pressure instead of passing it in and the default in MolecularDynamics add just the temperature.  That would save duplicating the rest of the code.

Will give it a think/go later

janus_core/md.py
ElliottKasoar: If the `final` files are saved in order to store configurations consistent with that different temperatures, does that not need more or less the same information as for restarting?

I suppose we don't need the current momenta, but it's currently stored anyway (in fact, everything in the restart is included, as well as the extra `spacegroup_kinds`), so I'm not sure I understand what we gain from distinguishing the two?

janus_core/md.py
alinelena: we fixed that one commits to show...

janus_core/md.py
ElliottKasoar: I have a feeling this created new problems for a similar issue when setting up the headers, but adding pressure with default as None in the main MD class might be an option?

janus_core/md.py
ElliottKasoar: Sorry, I'm not sure what you mean?

Also, sorry @harveydevereux I just saw that this was essentially your question at the start of the PR. My feeling is yes, ideally we'll have some indication that MD is run. Adding the extra temperature is probably the simplest option, but maybe there's a nicer format given that usually it should match the heating final temperature...

janus_core/md.py
harveydevereux: yes that would be another way around

janus_core/md.py
harveydevereux: I think @alinelena means I have that on my to-do to add the extra T if it is different to the temperature at the end of the heating. We talked about it at lunch.

janus_core/md.py
ElliottKasoar: It's still not completely clear to me exactly how specifically we would expect the two to differ. I can see in principle we may want to write out some things at the end rather than many times, but since the information essentially needs to be all be there in order to restart a simulation, I'm not sure how much there is to gain.

(Arguably just removing writing out `spacegroup_kinds` is not insignificant, but then it's unclear it there's any benefit in ever writing it out?)

I was actually just thinking about your second point in terms of including these changes within `_write_restart`. If the `final` outputs serve a different purpose, are formatted differently, and have no need to be worried about at a regular interval or influenced by any of the `restart` options, I think it would make a lot more sense to define it elsewhere in the code.

If there's any significant amount of code duplication, they can always call a separate shared function.

alinelena: by having a final we address two issues.
1. issue in the heating when you want to have the last configuration for each T ramp to be reused as input file for another run. there is the issue that restart can coincide with the final. that is a duplication we may want to avoid or we may not want to avoid. if one wants to avoid it will be a conflict with the restart in terms of meanings.
2. we address the point when restart frequencies, does not include the last step.
in terms of information restart and final may contain different information... final is to start a new simulation and not to restart it. restart will contain more info than a final.. so one can use restart for the purpose of final, given the steps align correctly.

I am happy to have them implemented separately but since this somehow addresses a mistmatch in restart I am not opposed in addressing it there, @harveydevereux what are your thoughts?

harveydevereux: I do think having some method ```_write_final``` or another name to handle the case of final is better in my opinion, even if the data is the same, because we need to invoke it in special circumstances.

The crux as Elliott is pointing out is what (if any) additional data needs to be in a ```-final``` over a  ```-res```, and why? I'm not familiar enough with Janus yet to make the decision.

I don't actually know what ```columns``` does, its not documented in ASE write, I can't find it anywhere in ASE IO actually. 

```
            write(
                restart_file,
                self.struct,
                write_info=True,
                columns=columns
            )
```

janus_core/md.py
```diff
@@ -657,6 +683,17 @@ def __init__(
             externalstress=self.pressure * units.bar,
         )
 
+    def _parameter_prefix(self) -> str:
```
harveydevereux, line 686: This is using a method to return parameters, the base version in MolecularDynamics just returns "" which is a bit clunky.

The advantage is if in future we extend and want even more parameters for certain ensembles it saves having to add all those variables to base as None's and checking them each

janus_core/md.py
```diff
@@ -471,11 +488,20 @@ def _write_traj(self) -> None:
                 append=append,
             )
 
+    def _write_final_state(self) -> None:
+        """Write the final system state."""
+        if not self.restart_stem:
+            restart_file = f"{self.file_prefix}-T{self.temp}-final.xyz"
+        write(restart_file, self.struct, write_info=True, columns=None)
```
harveydevereux, line 495: Currently left the final state as ```columns=None``` for the moment

ElliottKasoar: Sorry just to check, does `columns=None` write everything (as you had it before)?

harveydevereux: I am a bit confused about this according to [ASE](https://gitlab.com/ase/ase/-/blob/master/ase/io/extxyz.py?ref_type=heads#L829) ```None``` writes this, so I think it writes less? 

```
fr_cols = (['symbols', 'positions']
                       + [key for key in atoms.arrays.keys() if
                          key not in ['symbols', 'positions', 'numbers',
                                      'species', 'pos']])
```

janus_core/md.py
```diff
@@ -360,19 +360,36 @@ def _optimize_structure(self) -> None:
                 self.logger.info("Minimizing at step %s", self.dyn.nsteps)
             optimize(self.struct, **self.minimize_kwargs)
 
+    def _parameter_prefix(self) -> str:
```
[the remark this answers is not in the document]
harveydevereux: I was thinking that would be a good idea actually, it does make sense

janus_core/md.py
```diff
@@ -360,19 +360,36 @@ def _optimize_structure(self) -> None:
...
         """Setup filenames for output files."""
+
+        temp = ""
         if not self.file_prefix:
-            self.file_prefix = f"{self.struct_name}-{self.ensemble}-T{self.temp}"
+            self.file_prefix = (
+                f"{self.struct_name}-{self.ensemble}{self._parameter_prefix()}"
+            )
+            if self.temp_start and self.temp_end:
+                temp = f"-T{self.temp_start}-T{self.temp_end}"
```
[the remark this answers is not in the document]
oerc0122: Best I can think of are `Thermostat` and `Barostat` mixins which can be inherited by base ensembles defining these things.

janus_core/md.py
```diff
@@ -360,19 +360,36 @@ def _optimize_structure(self) -> None:
...
+        -------
+        str
+            Additional ensemble parameters written to a str.
+        """
+        return ""
+
     def configure_filenames(self) -> None:
         """Setup filenames for output files."""
+
+        temp = ""
```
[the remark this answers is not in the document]
harveydevereux: very true, temperature is probably the way to go

janus_core/md.py
```diff
@@ -846,6 +871,17 @@ def __init__(
             **kwargs,
         )
 
+    def _parameter_prefix(self) -> str:
```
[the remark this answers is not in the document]
harveydevereux: The issue is that we want the final state file to use the file prefix but not include the temperature ranges, only the spot temperature, and not the MD temperature, unless I've miss understood @alinelena on the other hand the stats and traj also want to use the file_prefix and the temperature ranges.

Currently if we do that those restart files will include also the temperature range, unless we also have a special file prefix for the final states, which we could. But then we would also need a special function to build that, as we would want the pressure from an NPT but not the T...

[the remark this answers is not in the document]
harveydevereux: @alinelena I missunderstodd what @oerc0122 was asking here but this hits at something I am not sure of

NVT_NH inherits from NPT, and therefore inherits the method which puts a pressure in the parameters

But the tests dictate that NVT_NH, should *not* have this pressure listed in the file names CF the tests

[NPT asks for a pressure](https://github.com/stfc/janus-core/blob/main/tests/test_md.py#L168-L172)

[NVT_NH does not, but inhertis from NPT](https://github.com/stfc/janus-core/blob/main/tests/test_md.py#L92-L96), but it will inherit these [_parameter_prefix methods from NPT](https://github.com/stfc/janus-core/blob/main/janus_core/md.py#L804) 

and therefore needs to be "re-overriden"

hence Jacob's question.


Basically should the test change to include the pressure, or should I do what is done here and respect the tests.

janus_core/md.py
harveydevereux: Currently working on this and also addressing some things Jacob mentioned.

I am going to assume from now that final should take the same columns, in pseudo-code

```python
write(
    final_file, 
    self.struct,  
    write_info=True, 
    columns=["symbols", "positions", "momenta", "masses"]
)
```

Unless there is a specific reason posted on this thread to not do this, there are some

we can always do it in future

ElliottKasoar: No objections here.

It may well be something to give users more control of at some point too.

janus_core/md.py
alinelena: None will write a standard xyz. we can consider adding columns as an option and user may select what wants to put in for final. but i will separate that as a different issue. we shall conclude this do not let perfect in the way of usable.

janus_core/md.py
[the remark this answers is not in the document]
ElliottKasoar: NVT_NH uses NPT, but sets the barostat time (so `pfactor`) to None (underlying ASE dynamics [here](https://gitlab.com/ase/ase/-/blob/master/ase/md/npt.py?ref_type=heads)), so I don't think we want the pressure labelled, as it's not a meaningful parameter.

Perhaps we should explicitly set `pressure=None` when we call the NPT init via NVT_NH, but I don't think it should make a difference, and it's an extra thing to deal with when setting `externalstress` within NPT.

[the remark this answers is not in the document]
harveydevereux: ok great, this is what I thought

I will assume the tests are correct and make sure the pressure is not there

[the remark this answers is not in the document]
alinelena: be carefull we need to document clear, pressure as target pressure and current pressure that is computed for all ensembles and displayed in -stats

janus_core/md.py
harveydevereux: done this in latest push

janus_core/md.py
harveydevereux: still ongoing with this, the MD T is added but currently always (not just if different)

tests/test_md.py
```diff
@@ -630,3 +646,61 @@ def test_heating_md(tmp_path):
...
+        assert isinstance(final_20_atoms, Atoms)
+
+        traj = read(traj_heating_path, index=":")
+        assert all(isinstance(image, Atoms) for image in traj)
+        assert len(traj) == 2
+
+        with open(stats_heating_path, encoding="utf8") as stats_file:
+            lines = stats_file.readlines()
+            assert "Target T [K]" in lines[0]
+            assert len(lines) == 54
```
ElliottKasoar, line 696: It might be good to check that the correct target temperature specifically here.

tests/test_md.py
```diff
@@ -630,3 +646,61 @@ def test_heating_md(tmp_path):
...
+
+    single_point = SinglePoint(
+        struct_path=DATA_PATH / "NaCl.cif",
+        architecture="mace",
+        calc_kwargs={"model": MODEL_PATH},
+    )
+    nvt = NVT(
+        struct=single_point.struct,
+        temp=25.0,
+        steps=100,
```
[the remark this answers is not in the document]
harveydevereux: Yes we should capture both

tests/test_md.py
harveydevereux: I think that is a good idea

janus_core/md.py
```diff
@@ -657,6 +708,62 @@ def __init__(
...
+
+        temperature_prefix = ""
+        if self.temp_start and self.temp_end:
+            temperature_prefix = f"-T{self.temp_start}-T{self.temp_end}"
+
+        if self.steps > 0:
+            temperature_prefix = f"{temperature_prefix}-T{self.temp}"
+
+        pressure = f"-p{self.pressure}" if not isinstance(self, NVT_NH) else ""
+        return f"{pressure}{temperature_prefix}"
```
ElliottKasoar, line 730: Not particularly important, but is there a benefit in swapping pressure and temperature from the original ordering?

Could we not also call the base function to get the temperature prefix, and append (or prepend) the pressure here?

janus_core/md.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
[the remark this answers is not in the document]
harveydevereux: The file is above the 1000 line limit (its ~1048)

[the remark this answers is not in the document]
ElliottKasoar: A lot of it's just parameters/docstrings, so may be fine. Not a problem for this PR anyway, but we'll probably want to think about if it makes sense to break it up.

janus_core/md.py
harveydevereux: Honestly I have gone through a lot of revisions of this code, it may no longer make sense to permute. 

I *think* it is a hangover from when the pressure was passed in as a variable? Maybe.

We could have the base handle the temperature yes

harveydevereux: @ElliottKasoar  The pressure is now restored as after the temperature, and we now use the base to get the temperature

tests/test_md.py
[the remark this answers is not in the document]
harveydevereux: I have added an additional test which only does the heating

tests/test_md.py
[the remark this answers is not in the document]
harveydevereux: There are some other tests with this pattern (I think?) but I've gone with using the ```Stats``` object directly and then querying its properties

harveydevereux:
> Can you update the README with the changes to the output files?

I need to re-update the docs now I have swapped the pressure back again

ElliottKasoar (changes requested):
This will need rebasing now that #120 is merged, sorry for the inconvenience.

harveydevereux:
> This will need rebasing now that #120 is merged, sorry for the inconvenience.

Just done a rebase

tests/test_md.py
```diff
@@ -630,3 +646,114 @@ def test_heating_md(tmp_path):
     assert stat_data.units[0] == ""
     assert stat_data.units[16] == "K"
     assert stat_data.labels[16] == "Target T"
+
+
+def test_heating_files():
+    "Test default heating file names" ""
```
ElliottKasoar, line 652:
```suggestion
    """Test default heating file names."""
```

tests/test_md.py
```diff
@@ -630,3 +646,114 @@ def test_heating_md(tmp_path):
...
+
+    single_point = SinglePoint(
+        struct_path=DATA_PATH / "NaCl.cif",
+        architecture="mace",
+        calc_kwargs={"model": MODEL_PATH},
+    )
+    nvt = NVT(
+        struct=single_point.struct,
+        temp=25.0,
+        steps=100,
```
ElliottKasoar, line 726: I think we might want to reduce this and/or look at other options, as this test takes significantly longer to run than any others, and I'm not sure we gain much by having a large number of additional steps. E.g. the slowest currently:

```
11.47s call     tests/test_md.py::test_heating_md_files
2.16s call     tests/test_md.py::test_npt
2.15s call     tests/test_md.py::test_nve
2.06s call     tests/test_md.py::test_minimize_every
1.68s call     tests/test_md.py::test_init[NVT-nvt]
1.66s call     tests/test_md.py::test_nph
```

janus_core/calculations/md.py
```diff
@@ -360,20 +361,74 @@ def _optimize_structure(self) -> None:
...
+        str
+           Formatted temperature range if heating and target temp if running md.
+        """
+
+        temperature_prefix = ""
+        if self.temp_start and self.temp_end:
+            temperature_prefix = f"-T{self.temp_start}-T{self.temp_end}"
+
+        if self.steps > 0:
+            temperature_prefix = f"{temperature_prefix}-T{self.temp}"
```
ElliottKasoar, line 380:
```suggestion
            temperature_prefix += f"-T{self.temp}"
```

janus_core/calculations/md.py
```diff
@@ -360,20 +361,74 @@ def _optimize_structure(self) -> None:
...
+        Restart file name.
+
+        Returns
+        -------
+        str
+           File name for restart files.
+        """
+        step = self.offset + self.dyn.nsteps
+        if not self.restart_stem:
+            return f"{self.struct_name}-{self.ensemble}-T{self.temp}-res-{step}.xyz"
```
ElliottKasoar, line 412: Can we not use `self.file_prefix` here (and in `_final_file`)? We seem to use `self.struct_name}-{self.ensemble}` etc. in a few difference places, when they ought to be handled together if possible.

ElliottKasoar (changes requested):

janus_core/calculations/md.py
harveydevereux: I think we can, simply the issue comes from the parameters potentially containing the temperature range from heating which only some files require.

I think then the file_prefix should be whatever is in common to all which is ```{self.struct_name}-{self.ensemble}``` only

harveydevereux: This is done

But I had to introduce a local for the stats and traj as their behaviour depends on whether ```file_prefix``` is set by the user or not. Which is lost information once ```file_prefix``` gets set if it was not. 

The difference in behaviour is that in the override case stats and traj's file name should not contain the temperature/ heating ranges/ pressure etc. (according to tests)

tests/test_md.py
harveydevereux: Hmm I did just run this on scarf and obtained, I don't think I did anything to change the performance characteristics, but I'm now seeing ```test_npt``` as the slowest

```shell
pytest -v tests/test_md.py --durations=0
```

```
============================================================================================ slowest durations =============================================================================================
15.80s call     tests/test_md.py::test_init[NVT-nvt]
10.78s call     tests/test_md.py::test_npt
7.05s call     tests/test_md.py::test_atoms_struct
4.57s call     tests/test_md.py::test_heating_md_files
0.70s call     tests/test_md.py::test_minimize_every
0.64s call     tests/test_md.py::test_restart
0.63s call     tests/test_md.py::test_heating_md
0.49s call     tests/test_md.py::test_traj_start
0.43s call     tests/test_md.py::test_minimize
0.43s call     tests/test_md.py::test_heating_files
0.41s call     tests/test_md.py::test_nvt_nh
0.39s call     tests/test_md.py::test_nph
0.38s call     tests/test_md.py::test_nve
0.38s call     tests/test_md.py::test_rescale_every
0.36s call     tests/test_md.py::test_rotate_restart
0.21s call     tests/test_md.py::test_heating
0.16s call     tests/test_md.py::test_init[NVE-nve]
0.13s call     tests/test_md.py::test_remove_rot
0.11s call     tests/test_md.py::test_init[NPT-npt]
0.11s call     tests/test_md.py::test_init[NPH-nph]
0.10s call     tests/test_md.py::test_init[NVT_NH-nvt-nh]
0.08s call     tests/test_md.py::test_reset_velocities
0.07s call     tests/test_md.py::test_noramp_heating
0.04s setup    tests/test_md.py::test_nve
```

alinelena: scarf can be moody on io so i suspect you see the variation based on that.

ElliottKasoar: Testing locally, `test_heating_md_files`, is fairly consistently the slowest for me by a noticeable amount.

However, I think the biggest thing isn't the relative speed, but more whether we gain anything from doing say 100 steps rather than a smaller number during MD, when the test is just about checking it does MD after heating, not testing the MD itself as such (and similarly for how long we spend heating).

For reference, I think the largest number of MD steps we do elsewhere is 6.

alinelena: let us look at real time per step, the test itself will be slower since I expect we do few T. for unit testing small steps number eg 5 per T is ok... we shall have examples with more steps that we run manually/tag that are more science relevant

harveydevereux: There should be a reduced load now, still covering the 10, 20, and 25 (MD) temperatures

```
7.03s call     tests/test_md.py::test_atoms_struct
4.03s call     tests/test_md.py::test_npt
2.04s call     tests/test_md.py::test_init[NVT-nvt]
0.74s call     tests/test_md.py::test_minimize_every
0.67s call     tests/test_md.py::test_heating_md
0.56s call     tests/test_md.py::test_heating_md_files
0.56s call     tests/test_md.py::test_restart
0.51s call     tests/test_md.py::test_heating_files
0.50s call     tests/test_md.py::test_nve
0.48s call     tests/test_md.py::test_traj_start
0.41s call     tests/test_md.py::test_rescale_every
0.39s call     tests/test_md.py::test_minimize
0.39s call     tests/test_md.py::test_nph
0.38s call     tests/test_md.py::test_rotate_restart
0.37s call     tests/test_md.py::test_nvt_nh
0.28s call     tests/test_md.py::test_heating
0.14s call     tests/test_md.py::test_init[NVE-nve]
0.11s call     tests/test_md.py::test_init[NPT-npt]
0.11s call     tests/test_md.py::test_remove_rot
0.10s call     tests/test_md.py::test_init[NPH-nph]
0.10s call     tests/test_md.py::test_init[NVT_NH-nvt-nh]
0.08s call     tests/test_md.py::test_noramp_heating
0.08s call     tests/test_md.py::test_reset_velocities
0.03s setup    tests/test_md.py::test_nve
```

ElliottKasoar: That seems more reasonable, thanks!

ElliottKasoar (approved):
I think at some point this module may need to be refactored, so we may revisit some of the interactions e.g. between NPT and NVT_NH, but I think for now this looks good, and particularly having (mostly) settled on the file naming convention we want is perhaps the most important thing.

Thanks, @harveydevereux!

harveydevereux:
> I think at some point this module may need to be refactored, so we may revisit some of the interactions e.g. between NPT and NVT_NH, but I think for now this looks good, and particularly having (mostly) settled on the file naming convention we want is perhaps the most important thing.
> 
> Thanks, @harveydevereux!

No problem, thanks for keeping my code in check. Refactoring may be a good idea, also for the file length

PR #119, opened 2024-04-23
Ensure config options are valid

### LINKED ISSUE #103: Investigate error for non-existent config args
> There will still be a `-` vs `_` difference, but probably close enough otherwise.
> 
> It would be much less of an issue if it didn't allow unspecified options too, so I need to figure out if that's fixable too

_Originally posted by @ElliottKasoar in https://github.com/stfc/janus-core/pull/100#discussion_r1553919917_

ElliottKasoar:
Resolves #103

Compares the options defined in configuration file to CLI function options to ensure only valid options are defined.

Also restructures code, as the CLI module was getting a bit unwieldy. This is entirely done within the first commit, so happy to separate out into two PRs if preferable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,10 +187,6 @@ This will run a singlepoint energy calculation on `KCl.cif` using the [MACE-MP](
 > [!NOTE]
 > `properties` must be passed as a Yaml list, as above, not as a string.
 
-> [!WARNING]
-> If an option in the configuration file does not match any variable names, an error will **not** be raised.
-> Please check the summary file to ensure the configuration has been read correctly.
-
 
 ## License
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -192,4 +192,5 @@ html_search_language = "en"
 nitpick_ignore = [
     ("py:class", "Logger"),
     ("py:class", "numpy.float64"),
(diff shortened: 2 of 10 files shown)


ElliottKasoar:
> My primary comment is that I think the restructure (which looks good) should be separated from any changes/additions.

Yeah, it became much bigger than I'd originally planned. I've separated the commit for #120, so this can be rebased once we're happy with that.

alinelena:
> My primary comment is that I think the restructure (which looks good) should be separated from any changes/additions.

can we let it go this time? in interest of time, @ElliottKasoar will note this

oerc0122:
> > My primary comment is that I think the restructure (which looks good) should be separated from any changes/additions.
> 
> can we let it go this time? in interest of time, @ElliottKasoar will note this

It's been done

alinelena:
> > > My primary comment is that I think the restructure (which looks good) should be separated from any changes/additions.
> > 
> > 
> > can we let it go this time? in interest of time, @ElliottKasoar will note this
> 
> It's been done

perfect... then will get the merge way

alinelena (approved):

oerc0122 (approved):

PR #123, opened 2024-04-25
Define __init__ version via pyproject.toml

ElliottKasoar:
This should mean we only need to change the version in one place.

Note: it updates when `poetry install` (or similar) is run, but I think that should be fine.

### DIFF
--- a/janus_core/__init__.py
+++ b/janus_core/__init__.py
@@ -1,3 +1,5 @@
 """Tools for machine learnt interatomic potentials."""
 
-__version__ = "0.2.0b6"
+from importlib.metadata import version
+
+__version__ = version("janus-core")


oerc0122:
Yeah, the alternative choice is defining it in the base `__init__`, and using `pyproject`'s [`dynamic = [version]`](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/#). Then it's attached to the code rather than the installer.

alinelena:
how will this work if user want to use pip to install rather than poetry?

ElliottKasoar:
> how will this work if user want to use pip to install rather than poetry?

Works the same way. When you run `pip install`, it set the correct version based on pyproject.toml. It only requires the correct distribution package to be installed (see the [docs](https://docs.python.org/3/library/importlib.metadata.html#overview)).

ElliottKasoar:
> Yeah, the alternative choice is defining it in the base `__init__`, and using `pyproject`'s [`dynamic = [version]`](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/#). Then it's attached to the code rather than the installer.

I have a feeling I might have tried this originally, and poetry didn't like the dynamic versioning. I think this way around probably makes slightly more sense to me anyway.

alinelena (approved):

PR #127, opened 2024-05-03
Add training for MACE

### LINKED ISSUE #36: add train

ElliottKasoar:
Resolves #36

Adds training (for MACE, but extensible), currently by passing the configuration file straight through to the MLIP training.

No tests currently as this depends on MACE's develop branch. We could change the pyproject.toml dependency, but the required changes should be merged to main and tagged in the next couple of weeks.

### DIFF
--- a/janus_core/helpers/train.py
+++ b/janus_core/helpers/train.py
@@ -1,0 +1,66 @@
+"""Train MLIP."""
+
+from pathlib import Path
+from typing import Optional
+
+from mace.cli.run_train import run as run_train
+from mace.tools import build_default_arg_parser as mace_parser
+import yaml
+
+from janus_core.helpers.janus_types import PathLike
+
+
+def check_files_exist(config: dict, req_file_keys: list[PathLike]) -> None:
+    """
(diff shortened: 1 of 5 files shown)


janus_core/helpers/train.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    if req_file_keys is None:
+        req_file_keys = ["train_file", "test_file", "valid_file", "statistics_file"]
+
+    # Validate inputs
+    with open(mlip_config, encoding="utf8") as file:
+        options = yaml.safe_load(file)
+    check_files_exist(options, req_file_keys)
+
+    if "foundation_model" in options:
```
alinelena, line 64: we may want fine tuning as a separate option since we need to check more stuff..
also will need a restart

alinelena (approved):

janus_core/helpers/train.py
alinelena: also will need tests once is in main branch, I already have the files prepared

PR #134, opened 2024-05-04
fix heating fixes #133

### LINKED ISSUE #133: heating keep T of thermostat fix
during heating we never reinitialise the thermostat T also we do not set the velocity consistent with the current target temperature. 

unrelated we may want to think if better to initialse the dynamic class for every step during the heating rather than having it like now. 
this has the advantage to be closer to someone will have if they script their own based on individual md runs. 
we will need to avoid also subtleties with random numbers for reproducible results if we go this route.

alinelena:


janus_core/calculations/md.py
```diff
@@ -607,7 +617,12 @@ def _run_dynamics(self) -> None:
             if self.logger:
                 self.logger.info("Beginning temperature ramp at %sK", temps[0])
             for temp in temps:
+                if temp == 0.0:
+                    continue
                 self.temp = temp
+                self._set_velocity_distribution()
+                if self.ensemble != "nve":
```
ElliottKasoar, line 624: I don't think it should make a difference, but might it be safer to check against the class, as we do elsewhere (e.g. NVT_NH)?

janus_core/calculations/md.py
```diff
@@ -292,10 +292,16 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-locals,too-many-sta
 
         # Warn if mix of None and not None
         self.ramp_temp = (
-            self.temp_start and self.temp_end and self.temp_step and self.temp_time
+            self.temp_start is not None
```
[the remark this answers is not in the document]
alinelena: yes, otherwise we end up with silly numbers.

janus_core/calculations/md.py
alinelena: yap good point I will update

oerc0122 (changes requested):

janus_core/calculations/md.py
```diff
@@ -607,7 +617,12 @@ def _run_dynamics(self) -> None:
             if self.logger:
                 self.logger.info("Beginning temperature ramp at %sK", temps[0])
             for temp in temps:
+                if temp == 0.0:
```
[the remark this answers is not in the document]
alinelena: is intentional.. i want to catch when the user puts 0.0 in the input... although you may be right since there is a np.arange that provides 0.0 now and not the user.

janus_core/calculations/md.py
```diff
@@ -607,7 +618,12 @@ def _run_dynamics(self) -> None:
             if self.logger:
                 self.logger.info("Beginning temperature ramp at %sK", temps[0])
             for temp in temps:
+                if isclose(temp, 0.0):
+                    continue
```
ElliottKasoar, line 622: Similar to @oerc0122's comment, if we were to decrease the temperature to 0, wouldn't this mean the velocity distribution was wrong here?

alinelena: ok I did not understand the comment let me think about it... why shall it be wrong?

ElliottKasoar: I was thinking for the same reason as why `set_velocity_distribution` is necessary in general.

An extreme example, but if `temps = [100, 50, 0]`, then this would call `_set_velocity_distribution` at T=100 and T=50, but not set it for T=0, so the velocity distribution would be wrong (non-zero) for the final `dyn.run`?

Possibly the fact that we still set `temperature_K=0` means this isn't a problem though?

janus_core/calculations/md.py
```diff
@@ -292,10 +292,16 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-locals,too-many-sta
 
         # Warn if mix of None and not None
         self.ramp_temp = (
-            self.temp_start and self.temp_end and self.temp_step and self.temp_time
+            self.temp_start is not None
+            and self.temp_end
```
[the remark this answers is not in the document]
ElliottKasoar: I think we just need `and self.temp_end is not None` to allow `temp_end = 0` as an option

janus_core/calculations/md.py
alinelena: ok just checked it... as long as step is negative shall be fine, updated to account for temp_end being possible 0

janus_core/calculations/md.py
```diff
@@ -373,7 +384,7 @@ def _parameter_prefix(self) -> str:
         """
 
         temperature_prefix = ""
-        if self.temp_start and self.temp_end:
+        if self.temp_start is not None and self.temp_end:
```
ElliottKasoar, line 387:
```suggestion
        if self.temp_start is not None and self.temp_end is not None:
```

ElliottKasoar (approved):

PR #143, opened 2024-05-13
Add phonons

### LINKED ISSUE #19: poc: phonons
create poc to calculate phonons using phonopy

ElliottKasoar:
Resolves #19, #149, #150, #151

Based on PoC: https://gitlab.com/drFaustroll/lavello/-/blob/main/phonons.py?ref_type=heads

To do:

- [x] Tidy docstrings etc.
- [x] Add tests for Python/CLI
- [x] Add options for specifying output files
- [x] Add plotting back in (separate PR) #149 
- [x] separate band structure from force constants (seprate pr) #150 
- [x] symmetrize force constants and cell option(separate pr) #151 
- [x] Document in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,18 +173,18 @@ When MD is run with heating the trajectory ```NaCl-nvt-T20.0-T300.0-T300.0-traj.
 
 ### Phonons
 
-Calculate phonons with a 2x3x4 supercell, after geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/mace-mp):
+Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/mace-mp):
 
 ```shell
-janus phonons --struct tests/data/NaCl.cif --supercell 2 --supercell 3 --supercell 4 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
+janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
 This will save the Phonopy parameters, including displacements and force constants, to `NaCl-params.yml`, and the calculated band structure to `NaCl-auto-band.yml`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
 
-Further calculations, including thermal properties, DOS, and PDOS, can also be calculated (using a 2x2x2 supercell):
+Further calculations, including thermal properties, DOS, and PDOS, can also be calculated (using a 2x3x4 supercell):
 
 ```shell
-janus phonons --struct tests/data/NaCl.cif --supercell 2 --dos --pdos --thermal --temp-start 0 --temp-end 300 --temp-step 50
+janus phonons --struct tests/data/NaCl.cif --supercell 2x3x4 --dos --pdos --thermal --temp-start 0 --temp-end 300 --temp-step 50
 ```
 
 This will create additional output files: `NaCl-cv.dat` for the thermal properties (heat capacity, entropy, and free energy) between 0K and 300K, `NaCl-dos.dat` for the DOS, and `NaCl-pdos.dat` for the PDOS.
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ Tools for machine learnt interatomic potentials
   - NVT (Langevin(Eijnden/Ciccotti flavour) and Nosé-Hoover (Melchionna flavour))
   - NPT (Nosé-Hoover (Melchiona flavour))
 - [ ] Nudge Elastic Band
-- [ ] Phonons
-  - vibroscopy
+- [x] Phonons
+  - Phonopy
 - [x] Training ML potentials
   - MACE
 - [x] Fine tunning MLIPs
@@ -169,6 +169,27 @@ janus md --ensemble nvt --struct tests/data/NaCl.cif --temp-start 20 --temp-end 
 This performs the same initial heating, before running a further 1000 steps (1 ps) at 300K.
 
 When MD is run with heating the trajectory ```NaCl-nvt-T20.0-T300.0-T300.0-traj.xyz``` and statistics ```NaCl-nvt-T20.0-T300.0-T300.0-stats.dat``` files will indicate the heating range and MD temperature (which may be different). With heating and MD trajectories/statistics within the same files.
+
+
+### Phonons
+
+Calculate phonons with a 2x3x4 supercell, after geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/mace-mp):
+
+```shell
+janus phonons --struct tests/data/NaCl.cif --supercell 2 --supercell 3 --supercell 4 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
+```
+
+This will save the Phonopy parameters, including displacements and force constants, to `NaCl-params.yml`, and the calculated band structure to `NaCl-auto-band.yml`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
+
+Further calculations, including thermal properties, DOS, and PDOS, can also be calculated (using a 2x2x2 supercell):
+
+```shell
+janus phonons --struct tests/data/NaCl.cif --supercell 2 --dos --pdos --thermal --temp-start 0 --temp-end 300 --temp-step 50
+```
+
+This will create additional output files: `NaCl-cv.dat` for the thermal properties (heat capacity, entropy, and free energy) between 0K and 300K, `NaCl-dos.dat` for the DOS, and `NaCl-pdos.dat` for the PDOS.
+
+For all options, run `janus phonons --help`.
 
 
 ### Using configuration files
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -130,6 +130,28 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
         self.calc = self.struct.calc
         self.results = {}
 
+    def _set_filename(
+        self, default_suffix: str, filename: Optional[PathLike] = None
+    ) -> str:
+        """
+        Set filename using the file prefix and suffix if not specified otherwise.
+
+        Parameters
+        ----------
+        default_suffix : str
+            Default suffix to use if `filename` is not specified.
+        filename : Optional[PathLike]
+            Filename to use, if specified. Default is None.
+
+        Returns
+        -------
+        str
+            Filename specified, or default filename.
+        """
+        if filename:
+            return filename
+        return f"{self.file_prefix}-{default_suffix}"
+
     def calc_phonons(self, write_results: bool = True) -> None:
         """
         Calculate phonons and optionally write results.
@@ -204,10 +226,9 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
             Name of hdf5 file to save force constants. Unused if `force_consts_to_hdf5`
             is False. Default is inferred from `file_prefix`.
         """
-        params_file = params_file if params_file else f"{self.file_prefix}-params.yml"
-        bands_file = bands_file if bands_file else f"{self.file_prefix}-auto_band.yml"
-        if not force_consts_file:
-            force_consts_file = f"{self.file_prefix}-force_consts.hdf5"
+        params_file = self._set_filename("params.yml", params_file)
+        bands_file = self._set_filename("auto_band.yml", bands_file)
+        force_consts_file = self._set_filename("force_consts.hdf5", force_consts_file)
 
         phonon = self.results["phonon"]
 
@@ -259,7 +280,7 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
             Name of data file to save thermal properties. Default is inferred from
             `file_prefix`.
         """
-        filename = filename if filename else f"{self.file_prefix}-cv.dat"
+        filename = self._set_filename("cv.dat", filename)
 
         with open(filename, "w", encoding="utf8") as out:
             temps = self.results["thermal_properties"]["temperatures"]
@@ -312,7 +333,7 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
             Name of data file to save the calculated DOS. Default is inferred from
             `file_prefix`.
         """
-        filename = filename if filename else f"{self.file_prefix}-dos.dat"
+        filename = self._set_filename("dos.dat", filename)
         self.results["phonon"].total_dos.write(filename)
 
     def calc_pdos(
@@ -356,7 +377,7 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
             Name of data file to save the calculated PDOS. Default is inferred from
             `file_prefix`.
         """
-        filename = filename if filename else f"{self.file_prefix}-pdos.dat"
+        filename = self._set_filename("pdos.dat", filename)
         self.results["phonon"].projected_dos.write(filename)
 
     # No magnetic moments considered
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -40,8 +40,11 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
     hdf5 : bool
         Whether to write force constants in hdf format or not.
         Default is True.
-    plot : bool
+    plot_to_file : bool
         Whether to plot various graphs as band stuctures, dos/pdos in svg.
+        Default is False.
+    symmetrize : bool
+        Whether to symmetrize force constants after cauclation.
         Default is False.
     minimize_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to geometry optimizer. Default is {}.
@@ -71,8 +74,9 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
         t_min: float = 0.0,
         t_max: float = 1000.0,
         minimize: bool = False,
-        hdf5: bool = False,
-        plot: bool = False,
+        hdf5: bool = True,
+        plot_to_file: bool = False,
+        symmetrize: bool = False,
         minimize_kwargs: Optional[dict[str, Any]] = None,
         file_prefix: Optional[PathLike] = None,
         log_kwargs: Optional[dict[str, Any]] = None,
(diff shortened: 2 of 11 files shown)


janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,302 @@
...
+        disp_supercells = phonon.supercells_with_displacements
+
+        phonon.forces = [
+            self.calc_forces(supercell)
+            for supercell in disp_supercells
+            if supercell is not None
+        ]
+
+        phonon.produce_force_constants()
+        phonon.run_mesh()
```
alinelena, line 153: computing thermal properties shall be optional.

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,353 @@
...
+        self.results["phonon"] = phonon
+
+        if self.logger:
+            self.logger.info("Phonons calculation complete")
+
+        if write_results:
+            self.results["phonon"].save(
+                f"{self.file_prefix}-params.yml", settings={"force_constants": True}
```
alinelena, line 179: will be usesful to write them in hdf5
somethins like phonon.save(filename=f"{sysname}-{model}.yaml", settings={"force_constants": False})
write_force_constants_to_hdf5(phonon.force_constants, filename=f"force_constants.hdf5")

ElliottKasoar: Ok, why do we not want them in the same file?

README.md
```diff
@@ -171,6 +171,27 @@ This performs the same initial heating, before running a further 1000 steps (1 p
 When MD is run with heating the trajectory ```NaCl-nvt-T20.0-T300.0-T300.0-traj.xyz``` and statistics ```NaCl-nvt-T20.0-T300.0-T300.0-stats.dat``` files will indicate the heating range and MD temperature (which may be different). With heating and MD trajectories/statistics within the same files.
 
 
+### Phonons
+
+Calculate phonons with a 2x3x4 supercell, after geometry optimization (using the [MACE-MP](https://github.com/ACEsuit/mace-mp):
+
+```shell
+janus phonons --struct tests/data/NaCl.cif --supercell 2 --supercell 3 --supercell 4 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
```
oerc0122, line 179: It's a bit confusing to specify the x-y-z supercell as 3 separate calls to supercell. If there isn't a better way to specify a list, perhps a string like:
```suggestion
janus phonons --struct tests/data/NaCl.cif --supercell 2x3x4 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
```
And then split it in code on `x`

README.md
```diff
@@ -171,6 +171,27 @@ This performs the same initial heating, before running a further 1000 steps (1 p
...
+```
+
+This will save the Phonopy parameters, including displacements and force constants, to `NaCl-params.yml`, and the calculated band structure to `NaCl-auto-band.yml`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
+
+Further calculations, including thermal properties, DOS, and PDOS, can also be calculated (using a 2x2x2 supercell):
+
+```shell
+janus phonons --struct tests/data/NaCl.cif --supercell 2 --dos --pdos --thermal --temp-start 0 --temp-end 300 --temp-step 50
```
oerc0122, line 187: This also makes it more explicit here, because you would have to write `2x2x2` which is not a lot of extra typing

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,421 @@
...
+            Name of hdf5 file to save force constants. Default is inferred from
+            `file_prefix`.
+        bands_file : Optional[PathLike]
+            Name of yaml file to save band structure. Default is inferred from
+            `file_prefix`.
+        """
+        params_file = params_file if params_file else f"{self.file_prefix}-params.yml"
+        bands_file = bands_file if bands_file else f"{self.file_prefix}-auto_band.yml"
+        if not force_const_file:
+            force_const_file = f"{self.file_prefix}-force_consts.hdf5"
```
oerc0122, line 206: Consistency?

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        Write results of thermal properties calculations.
+
+        Parameters
+        ----------
+        filename : Optional[PathLike]
+            Name of data file to save thermal properties. Default is inferred from
+            `file_prefix`.
+        """
+        filename = filename if filename else f"{self.file_prefix}-cv.dat"
```
oerc0122, line 253: This is such a common pattern, we may want a function somewhere which is something like:
```python
def _default_filename(self, filename: Optional[PathLike], default: str):
    if filename:
        return filename
    return f"{self.file_prefix}-{default}"
```

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,421 @@
...
+            temps = self.results["thermal_properties"]["temperatures"]
+            c_vs = self.results["thermal_properties"]["heat_capacity"]
+            entropies = self.results["thermal_properties"]["entropy"]
+            free_energies = self.results["thermal_properties"]["free_energy"]
+
+            print("#Temperature [K] | Cv | H | S ", file=out)
+            for temp, c_v, free_energy, entropy in zip(
+                temps, c_vs, free_energies, entropies
+            ):
+                print(f"{temp} {c_v} {free_energy} {entropy}", file=out)
```
oerc0122, line 265: Default separator of print is a space, since you aren't taking advantage of tablular formatting here, could be better expressed as:
```suggestion
                print(temp, c_v, free_energy, entropy, file=out)
```
or if you just had the `tuple` rather than unpacking it
```suggestion
                print(*line, file=out)
```

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Parameters
+        ----------
+        mesh : MaybeList[float]
+            Mesh for sampling. Default is [10, 10, 10].
+        write_results : bool
+            Whether to write out results to file. Default is True.
+        """
+        if not mesh:
+            mesh = [10, 10, 10]
```
oerc0122, line 281:
```suggestion
        self, mesh: MaybeList[float] = (10, 10, 10), write_results=True
    ) -> None:
        """
        Calculate density of states and optionally write results.

        Parameters
        ----------
        mesh : MaybeList[float]
            Mesh for sampling. Default is [10, 10, 10].
        write_results : bool
            Whether to write out results to file. Default is True.
        """
```

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,421 @@
...
+        ----------
+        filename : Optional[PathLike]
+            Name of data file to save the calculated DOS. Default is inferred from
+            `file_prefix`.
+        """
+        filename = filename if filename else f"{self.file_prefix}-dos.dat"
+        self.results["phonon"].total_dos.write(filename)
+
+    def calc_pdos(
+        self, mesh: Optional[MaybeList[float]] = None, write_results: bool = True
```
oerc0122, line 313: See: above

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,263 @@
...
+    Parameters
+    ----------
+    ctx : Context
+        Typer (Click) Context. Automatically set.
+    struct : Path
+        Path of structure to simulate.
+    struct_name : Optional[PathLike]
+        Name of structure to simulate. Default is inferred from filepath or chemical
+        formula.
+    supercell : List[int]
```
oerc0122, line 115: It isn't strictly a `list`, it can also be a single `int`.

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,263 @@
...
+        raise ValueError("'fmax' must be passed through the --fmax option")
+    minimize_kwargs["fmax"] = fmax
+
+    # Set default supercell
+    if not supercell:
+        supercell = [2, 2, 2]
+
+    # Convert single value to list for supercell matrix
+    if len(supercell) == 1:
+        supercell *= 3
```
oerc0122, line 193: Ok, I pass in `[[2, 2, 2]]` this is duplicated to `[[2, 2, 2], [2, 2, 2], [2, 2, 2]]`, which has length 3, it would fall over later, but I think the check might be overly simplistic. Not certain we need to care, but just to be aware...

tests/test_phonons_cli.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def test_help():
+    """Test calling `janus phonons --help`."""
+    result = runner.invoke(app, ["phonons", "--help"])
+    assert result.exit_code == 0
+    assert "Usage: janus phonons [OPTIONS]" in result.stdout
+
+
+def test_phonons(tmp_path):
```
oerc0122, line 21: Where is `tmp_path` coming from? Is this part of `typer`? If so, pretty neat.

oerc0122 (changes requested):

janus_core/calculations/phonons.py
ElliottKasoar: True, this may solve the consistency issue too.

janus_core/cli/phonons.py
ElliottKasoar: typer doesn't allow Union types yet, so even a single integer would be passed as [x]

Doesn't matter too much if we do change it to a string though.

janus_core/cli/phonons.py
ElliottKasoar: It checks that the value after --supercell is an integer at the moment, so you can't pass  `--supercell [2, 2, 2]` or even just `--supercell [2]` (with or without quotes around the [ ] )

tests/test_phonons_cli.py
ElliottKasoar: No, it's built in to pytest: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

It is very useful

janus_core/calculations/phonons.py
ElliottKasoar: In theory I think it can be most array-like things, or a float, so `MaybeSequence`?

oerc0122: Yeah, because that's what you're meant to pass in. While the default doesn't *strictly* conform it's close enough. The alternative is `MaybeSequence` but I think this is more misleading. Maybe we ought to create a `3-vector` type.

oerc0122: `ArrayLike` could work, but unfortunately:
> Among others this includes the likes of:
> 
> Scalars.

ElliottKasoar: Hmm ok I'll leave it as `MaybeList` for now

ElliottKasoar:
Things left, but could all be changed/added in future PRs(?):

- [ ] Plotting
- [ ] Option not to calculate force constants etc. without band structure
- [ ] Level of approximation (quasi-harmonic, but also ...)
- [ ] Symmetrization

janus_core/cli/phonons.py
```diff
@@ -184,18 +184,17 @@ def phonons(
...
-        raise ValueError(
-            """Please pass the lattice vectors via either --supercell x --supercell y \
---supercell z, or --supercell x."""
-        )
+        raise ValueError("Please pass three lattice vectors in the form 1x2x3")
+
+    if not all(vector.isdigit() for vector in supercell):
+        raise ValueError("Please pass lattice vectors as integers in the form 1x2x3")
+
+    supercell = [int(vector) for vector in supercell]
```
oerc0122, line 197: Minor stylistic thing
```suggestion
    try:
        supercell = [int(x) for x in supercell.split("x")]
    except ValueError as exc:
        raise ValueError("Please pass lattice vectors as integers in the form 1x2x3") from exc

    # Validate supercell list
    if len(supercell) != 3:
        raise ValueError("Please pass three lattice vectors in the form 1x2x3")
```

tests/test_phonons_cli.py
```diff
@@ -115,3 +115,63 @@ def test_pdos(tmp_path):
...
+
+    # Check parameters
+    with open(param_file, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+
+    assert "supercell_matrix" in params
+    assert len(params["supercell_matrix"]) == 3
+    assert params["supercell_matrix"][0] == [2, 0, 0]
+    assert params["supercell_matrix"][1] == [0, 3, 0]
+    assert params["supercell_matrix"][2] == [0, 0, 4]
```
oerc0122, line 151: Another minor stylistic thing
```suggestion
    assert params["supercell_matrix"] == [[2, 0, 0],
                                          [0, 3, 0],
                                          [0, 0, 4]]
```
?

oerc0122 (approved):
Just a couple of minor stylistic suggestions. All looks good.

janus_core/cli/phonons.py
ElliottKasoar: Ah I originally considered int(x), but didn't realise something like `int("2.4")` throws an error, since `int(2.4)` doesn't. That's good to know

oerc0122 (approved):

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,261 @@
...
+        float,
+        Option(help="Displacement for force constants calculation, in A."),
+    ] = 0.01,
+    thermal: Annotated[
+        bool,
+        Option(help="Whether to calculate thermal properties."),
+    ] = False,
+    temp_start: Annotated[
+        float,
+        Option(help="Start temperature for CV calculations, in K."),
```
alinelena, line 58:
```suggestion
        Option(help="Start temperature for thermal properties calculations, in K."),
```

janus_core/calculations/phonons.py
```diff
@@ -0,0 +1,443 @@
...
+    supercell : MaybeList[int]
+        Size of supercell for calculation. Default is 2.
+    displacement : float
+        Displacement for force constants calculation, in A. Default is 0.01.
+    t_step : float
+        Temperature step for CV calculations, in K. Default is 50.0.
+    t_min : float
+        Start temperature for CV calculations, in K. Default is 0.0.
+    t_max : float
+        End temperature for CV calculations, in K. Default is 1000.0.
```
alinelena, line 36:
```suggestion
    t_step : float
        Temperature step for thermal properties calculations, in K. Default is 50.0.
    t_min : float
        Start temperature for thermal properties calculations, in K. Default is 0.0.
    t_max : float
        End temperature for thermal properties calculations, in K. Default is 1000.0.
```

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,261 @@
...
+        float,
+        Option(help="Start temperature for thermal properties calculations, in K."),
+    ] = 0.0,
+    temp_end: Annotated[
+        float,
+        Option(help="End temperature for CV calculations, in K."),
+    ] = 1000.0,
+    temp_step: Annotated[
+        float,
+        Option(help="Temperature step for CV calculations, in K."),
```
alinelena, line 66:
```suggestion
        Option(help="End temperature for thermal properties calculations, in K."),
    ] = 1000.0,
    temp_step: Annotated[
        float,
        Option(help="Temperature step for thermal properties calculations, in K."),
```

alinelena (dismissed):

janus_core/calculations/phonons.py
```diff
@@ -191,50 +208,90 @@ def calc_phonons(self, write_results: bool = True) -> None:
...
+        plot_file : Optional[PathLike]
+            Name of svg file to save band structure. Default is inferred from
+            `file_prefix`.
+        """
+
+        bands_file = self._set_filename("auto_bands.yml", bands_file)
+        self.results["phonon"].auto_band_structure(
+            write_yaml=write_bands, filename=bands_file
+        )
+        if self.plot:
```
oerc0122, line 259: What about plot to screen? If you say "plot", that implies to screen to me.

janus_core/cli/phonons.py
```diff
@@ -55,16 +55,28 @@ def phonons(
...
+        Option(help="Temperature step for thermal properties calculations, in K."),
     ] = 50,
+    band: Annotated[
+        bool,
+        Option(help="Whether to save fc in hdf5."),
+    ] = False,
+    hdf5: Annotated[
+        bool,
+        Option(help="Whether to calculate the band structure."),
+    ] = True,
```
oerc0122, line 75: Are these descriptions backwards?

janus_core/calculations/phonons.py
alinelena: will not plot to screen, the ascii plot is cool but... but to file is nicer I will rename to plot_to_file

janus_core/cli/phonons.py
```diff
@@ -55,16 +55,28 @@ def phonons(
...
+    plot: Annotated[
+        bool,
+        Option(help="Whether to plot bandstructure and/pr dos/pdos when calculated."),
```
oerc0122, line 78:
```suggestion
        Option(help="Whether to plot bandstructure and/or dos/pdos when calculated."),
```

janus_core/cli/phonons.py
alinelena: yes I need to updated them... copy paste...

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,295 @@
...
+    thermal : bool
+        Whether to calculate thermal properties. Default is False.
+    temp_start : float
+        Start temperature for CV calculations, in K. Unused if `thermal` is False.
+        Default is 0.0.
+    temp_end : float
+        End temperature for CV calculations, in K. Unused if `thermal` is False.
+        Default is 1000.0.
+    temp_step : float
+        Temperature step for CV calculations, in K. Unused if `thermal` is False.
```
oerc0122, line 147: Do these need to match the definitions above?

janus_core/cli/phonons.py
```diff
@@ -0,0 +1,295 @@
...
+    summary : Path
+        Path to save summary of inputs and start/end time. Default is
+        phonons.yml.
+    config : Path
+        Path to yaml configuration file to define the above options. Default is None.
+    """
+    # Check options from configuration file are all valid
+    check_config(ctx)
+
+    [read_kwargs, calc_kwargs, minimize_kwargs] = parse_typer_dicts(
```
oerc0122, line 191: Square brackets on the left-hand side seems odd, what does this mean?

alinelena: this is from elliott so nee dot check

janus_core/cli/phonons.py
alinelena: what do you mean?

janus_core/cli/phonons.py
alinelena: removed them

janus_core/cli/phonons.py
oerc0122: Above it's swapped from "CV calculations" to "thermal properties calculations"

alinelena: fair point I will harmonise...

oerc0122 (approved):

PR #154, opened 2024-05-26
add option to write eigenvectors and extra info in bands, useful for …

### LINKED ISSUE #155: print eigenvectors with the bands
this is to allow visualisation

alinelena:
…visualisation, but we end up with bigger files fixes #155

- [x] tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ This will save the Phonopy parameters, including displacements and force constan
 Additionaly, to calculate band structure and generate `NaCl-auto-bands.yml` use:
 
 ```shell
-janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
+janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}" --band
 ```
 
 If you need eigenvectors and group velocities written, add the `--write-full` option. This will generate a much larger file, but can be used to visualise phonon modes.
--- a/README.md
+++ b/README.md
@@ -204,8 +204,9 @@ Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the
 janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
-This will save the Phonopy parameters, including displacements and force constants, to `NaCl-phonopy.yml` and `NaCl-force\_constants.hdf5`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
-Additionaly, to calculate band structure and generate `NaCl-auto-bands.yml` use:
+This will save the Phonopy parameters, including displacements and force constants, to `NaCl-phonopy.yml` and `NaCl-force_constants.hdf5`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
+
+Additionally, the `--band` option can be added to calculate the band structure and save the results to `NaCl-auto_bands.yml`:
 
 ```shell
 janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}" --band
--- a/README.md
+++ b/README.md
@@ -204,7 +204,14 @@ Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the
 janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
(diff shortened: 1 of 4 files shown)


janus_core/calculations/phonons.py
```diff
@@ -65,6 +68,7 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
     """
 
     def __init__(  # pylint: disable=too-many-arguments
+        # pylint: disable=too-many-locals
```
ElliottKasoar, line 71:
```suggestion
    def __init__(  # pylint: disable=too-many-arguments,disable=too-many-locals
```

tests/test_phonons_cli.py
```diff
@@ -184,6 +212,9 @@ def test_plot(tmp_path):
     assert autoband_results.exists()
     for svg in svgs:
         assert svg.exists()
+    with open(autoband_results, encoding="utf8") as file:
+        bands = yaml.safe_load(file)
+        assert "eigenvector" in bands["phonon"][0]["band"][0].keys()
```
ElliottKasoar, line 217:
```suggestion
        assert "eigenvector" in bands["phonon"][0]["band"][0].keys()
        assert "group_velocity" in bands["phonon"][0]["band"][0].keys()
```

ElliottKasoar (commented):
We may not need to explicitly add a mention of the new option in the README, but we should do a pass to make sure everything is consistent with the new (and old) behaviour - see also #158

janus_core/calculations/phonons.py
```diff
@@ -44,8 +44,11 @@ class Phonons:  # pylint: disable=too-many-instance-attributes
         Whether to plot various graphs as band stuctures, dos/pdos in svg.
         Default is False.
     symmetrize : bool
-        Whether to symmetrize force constants after cauclation.
+        Whether to symmetrize force constants after calculation.
         Default is False.
+    write_full : bool
+        Whether to maximize information written in various output files.
```
oerc0122, line 50: Not sure `maximize` is the right word here. 

```suggestion
        Alias to write all output files (hdf5, plot_to_file).
```
?

ElliottKasoar: Isn't it just setting what is written out via `auto_band_structure`, not changing which files are written? (That probably could be clearer too)

alinelena: yes at the moment it is just maximizing the info in bands, may be needed for dos/pdos too but I did not check yet.

alinelena: ok refrashed, recheck please.

README.md
```diff
@@ -204,15 +204,22 @@ Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the
 janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
 ```
 
-This will save the Phonopy parameters, including displacements and force constants, to `NaCl-params.yml`, and the calculated band structure to `NaCl-auto-band.yml`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
+This will save the Phonopy parameters, including displacements and force constants, to `NaCl-phonopy.yml` and `NaCl-force\_constants.hdf5`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.
+Additionaly, to calculate band structure and generate `NaCl-auto-bands.yml` use:
```
ElliottKasoar, line 208:
```suggestion
This will save the Phonopy parameters, including displacements and force constants, to `NaCl-phonopy.yml` and `NaCl-force_constants.hdf5`, in addition to generating a log file, `phonons.log`, and summary of inputs, `phonons_summary.yml`.

Additionally, the `--band` option can be added to calculate the band structure and save the results to `NaCl-auto_bands.yml`:
```

README.md
```diff
@@ -204,15 +204,22 @@ Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the
...
+
+```shell
+janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}"
```
ElliottKasoar, line 211:
```suggestion
janus phonons --struct tests/data/NaCl.cif --supercell 2x2x2 --minimize --arch mace_mp --calc-kwargs "{'model' : 'small'}" --band
```

README.md
```diff
@@ -204,15 +204,22 @@ Calculate phonons with a 2x2x2 supercell, after geometry optimization (using the
...
+```
+If you need eigenvectors and group velocities written add `--write-full` option. Will generate a much bigger bands file but can be
+used to visualise phonon modes.
```
ElliottKasoar, line 214:
```suggestion

If you need eigenvectors and group velocities written, add the `--write-full` option. This will generate a much larger file, but can be used to visualise phonon modes.
```

janus_core/calculations/phonons.py
oerc0122, line 50: If that's what it's doing, then I think `write_full_bands` or `write_full_info` would be a better name and specify which data are affected nad possibly how.

alinelena: initially I was looking at write_full_info but wanted to keep it shorter. plan is to extend it to other things.

ElliottKasoar (dismissed):

PR #162, opened 2024-05-29
separate optimizer from the optimize to be able to use geom opt in zn…

alinelena:
…draw, also keey pylint happy in md

### DIFF
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -31,7 +31,7 @@ def set_optimizer(
     logger: Optional[Logger] = None,
 ) -> tuple[Optimizer, Optional[Atoms]]:
     """
-    Set optimizer for geometry optimisation.
+    Set optimizer for geometry optimization.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


ElliottKasoar (commented):
Looks like this needs rebasing to include the pressure logging change, but otherwise looks good

janus_core/calculations/geom_opt.py
```diff
@@ -20,8 +22,61 @@
...
+def set_optimizer(
+    struct: Atoms,
+    filter_func: Optional[Callable] = DefaultFilter,
+    filter_kwargs: Optional[dict[str, Any]] = None,
+    optimizer: Callable = LBFGS,
+    opt_kwargs: Optional[ASEOptArgs] = None,
+    logger: Optional[Logger] = None,
+) -> tuple[Optimizer, Optional[Atoms]]:
+    """
+    Set optimizer for geometry optimisation.
```
ElliottKasoar, line 34:
```suggestion
    Set optimizer for geometry optimization.
```
To be consistent

oerc0122 (dismissed):
Fine with this following @ElliottKasoar 's suggestion for consistency

ElliottKasoar (approved):

PR #163, opened 2024-05-30
Add descriptors

### LINKED ISSUE #86: add posibility to calculate descriptors
the descriptors for a structure or for a set of structures shall be returned in dictionary for python or a file for cli... 

some bad poc in below

```python

# -*- coding: utf-8 -*-
# Author; alin m elena, alin@elena.re
# Contribs;
# Date: 01-03-2024
# ©alin m elena, GPL v3 https://www.gnu.org/licenses/gpl-3.0.en.html
from ase.io import read
import numpy as np
from mace.calculators import mace_mp


import argparse

cli=argparse.ArgumentParser()
cli.add_argument(
  "--device",
  type=str,
  default='cpu',
  choices = ['cuda','cpu','mps'],
  help = ' provide device to run. default %(default)s'
)

cli.add_argument(
  "--precision",
  type=str,
  default='float32',
  choices = ['float64','float32'],
  help = ' provide precision to run. default %(default)s'
)

cli.add_argument(
  "--invariant",
  action="store_true",
  help = 'enable invariant only? default %(default)s'
  )

cli.add_argument(
  "--reference",
  type=str,
  default=None,
  help = 'rederence structure, ase readable, default %(default)s'
)

cli.add_argument(
  "--trajectory",
  type=str,
  default=None,
  help = 'trajectory, default %(default)s'
)

cli.add_argument(
  "--model",
  type=str,
  default='medium',
  help = 'provide the mace model file. default: %(default)s'
)

cli.add_argument(
  "--out",
  type=str,
  default='descriptors.dat',
  help = 'output file. default: %(default)s'
)


def pme(t,el,mem,d,rmem=None,rd=None):
  m = " ".join(list(map(lambda x: f"{x:12.5E}" , [mem[e] for e in el ])))
  s = f"{t:10d} {m} {d:12.5E}"
  if rmem and rd:
     s += " ".join(list(map(lambda x: f"{x:12.5E}" , [mem[e]-rmem[e] for e in el ])))
     s += f"{d-rd:12.5E}"
  return s

def simil(conf,inv):
  species = sorted(set(conf.get_chemical_symbols()))
  patterns = {}
  for e in species:
    patterns[e]=[atom.index for atom in conf if atom.symbol == e]
  descriptors = calculator.get_descriptors(conf,invariants_only=inv)
  me = {}
  mem = {}
  for e in species:
    me[e] = descriptors[patterns[e],:].mean(axis=0)
    mem[e] = np.mean(me[e])
  return species, me, mem, np.mean(descriptors)

args = cli.parse_args()

model = args.model
device = args.device
precision = args.precision

reference = args.reference
traj = args.trajectory

inv = args.invariant
out = args.out

out = open(out,"w+")

# for descriptors single precision is ok
calculator = mace_mp(model=model, default_dtype=precision, device=device)
ref = read(reference)

el, rme, rmem, rd  = simil(ref, inv)
header = f"#Time     | {' | '.join([f'{x:10s}' for x in el])} | Descriptor |"
print("# ", header,file=out)
print("# ref ", pme(0,el, rmem,rd),file=out)
header += f"{' | '.join([f'*{x:10s}' for x in el])}| Descriptor*|"
print(header, file=out)

conf = read(traj, index=":")
for t,f in enumerate(conf):
  el, me, mem, d  = simil(f, inv)
  print(f"{pme(t,el,mem,d,rmem,rd)}",file=out)

out.close()
```

ElliottKasoar:
Resolves #86

To do:
- [x] Update README
- [x] Update API docs

### DIFF
--- a/janus_core/cli/descriptors.py
+++ b/janus_core/cli/descriptors.py
@@ -1,0 +1,174 @@
+"""Set up MLIP descriptors commandline interface."""
+
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Annotated
+
+from typer import Context, Option, Typer
+from typer_config import use_config
+
+from janus_core.calculations.single_point import SinglePoint
+from janus_core.cli.types import (
+    Architecture,
+    CalcKwargs,
+    Device,
+    LogPath,
+    ReadKwargs,
+    StructPath,
+    Summary,
+    WriteKwargs,
+)
+from janus_core.cli.utils import (
+    check_config,
+    end_summary,
+    parse_typer_dicts,
+    start_summary,
+    yaml_converter_callback,
+)
+from janus_core.helpers.descriptors import calc_descriptors
+from janus_core.helpers.utils import dict_paths_to_strs
+
+app = Typer()
+
+
+@app.command(help="Calculate MLIP descriptors.")
+@use_config(yaml_converter_callback)
+def descriptors(
+    # pylint: disable=too-many-arguments,too-many-locals
+    # numpydoc ignore=PR02
+    ctx: Context,
+    struct: StructPath,
+    invariants_only: Annotated[
+        bool,
+        Option(help="Only calculate invariant descriptors."),
+    ] = True,
+    calc_elements: Annotated[
+        bool,
+        Option(help="Calculate mean descriptors for each element."),
+    ] = False,
+    arch: Architecture = "mace_mp",
+    device: Device = "cpu",
+    out: Annotated[
+        Path,
+        Option(
+            help=(
+                "Path to save structure with calculated descriptors. Default is "
+                "inferred from name of structure file."
+            ),
+        ),
+    ] = None,
+    read_kwargs: ReadKwargs = None,
+    calc_kwargs: CalcKwargs = None,
+    write_kwargs: WriteKwargs = None,
+    log: LogPath = "descriptors.log",
+    summary: Summary = "descriptors_summary.yml",
+):
+    """
+    Calculate MLIP descriptors for the given structure(s).
+
+    Parameters
+    ----------
+    ctx : Context
+        Typer (Click) Context. Automatically set.
+    struct : Path
+        Path of structure to simulate.
+    invariants_only : bool
+        Whether only the invariant descriptors should be returned. Default is True.
+    calc_elements : bool
+        Whether to calculate mean descriptors for each element. Default is False.
+    arch : Optional[str]
+        MLIP architecture to use for single point calculations.
+        Default is "mace_mp".
+    device : Optional[str]
+        Device to run model on. Default is "cpu".
+    out : Optional[Path]
+        Path to save structure with calculated results. Default is inferred from name
+        of the structure file.
+    read_kwargs : Optional[dict[str, Any]]
+        Keyword arguments to pass to ase.io.read. Default is {}.
+    calc_kwargs : Optional[dict[str, Any]]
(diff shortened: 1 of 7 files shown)


ElliottKasoar:
Thinking about our discussion yesterday @alinelena, did you actually want the mean descriptor for each atom (since you mentioned it's a lot of data), rather than per element, or is it fine as it is?

If we want an option for each atom, I'd probably swap out the current flag and always save the mean for each element.

janus_core/cli/descriptors.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Parameters
+    ----------
+    ctx : Context
+        Typer (Click) Context. Automatically set.
+    struct : Path
+        Path of structure to simulate.
+    invariants_only : bool
+        Whether only the invariant descriptors should be returned. Default is True.
+    calc_elements : bool
```
alinelena, line 77:
```suggestion
    calc_per_element : bool
```

tests/test_descriptors.py
```diff
@@ -0,0 +1,56 @@
...
+
+    atoms = calc_descriptors(
+        single_point.struct,
+        calc_elements=True,
+        log_kwargs={"filename": log_file},
+    )
+    assert isinstance(atoms, Atoms)
+    assert "descriptor" in atoms.info
+    assert "Na_descriptor" in atoms.info
+    assert "Cl_descriptor" in atoms.info
```
alinelena, line 56: I will also check the value not only if it is calculated

janus_core/helpers/descriptors.py
```diff
@@ -0,0 +1,143 @@
...
+    [write_kwargs, log_kwargs] = none_to_dict([write_kwargs, log_kwargs])
+    log_kwargs.setdefault("name", __name__)
+    logger = config_logger(**log_kwargs)
+
+    # Set default name for output file
+    if not struct_name:
+        if isinstance(struct, Sequence):
+            struct_name = struct[0].get_chemical_formula()
+        else:
+            struct_name = struct.get_chemical_formula()
```
oerc0122, line 61: Do we just want to convert non-`Sequence` structs into a single-element `tuple`?

janus_core/helpers/descriptors.py
```diff
@@ -0,0 +1,143 @@
...
+        "filename",
+        Path(f"./{struct_name}-descriptors.xyz").absolute(),
+    )
+
+    if isinstance(struct, Sequence):
+        if any(not image.calc for image in struct):
+            raise ValueError("Please attach a calculator to all images in `struct`.")
+    else:
+        if not struct.calc:
+            raise ValueError("Please attach a calculator to `struct`.")
```
oerc0122, line 73: Is this the same thing, but flatter?
```suggestion
    if isinstance(struct, Sequence) and any(not image.calc for image in struct):
            raise ValueError("Please attach a calculator to all images in `struct`.")
    elif not struct.calc:
            raise ValueError("Please attach a calculator to `struct`.")
```

oerc0122 (changes requested):
Looks good, but Alin's point needs addressing. Otherwise stylistic.

janus_core/helpers/descriptors.py
ElliottKasoar: We could do, but then we'd have to convert back at the end, or put in `Atoms` and get back a `tuple`, which might be a bit annoying.

oerc0122 (approved):

PR #165, opened 2024-05-30
Read all structures in trajectory by default

### LINKED ISSUE #114: singlepoint shall be able to operate on multi image inputs
at the moment if i pass a multiframe xyx for example only first is calculated.

ElliottKasoar:
Resolves #114

Tests are modified so that where previously we had to specify `index=":"` via read_kwargs, the same result is obtained by default, while specifying the index is instead tested by choosing a single structure from a trajectory.


alinelena (approved):

oerc0122 (changes requested):
Apart from docstring, looks good.

alinelena (approved):

alinelena (approved):

oerc0122 (approved):

PR #173, opened 2024-05-31
Add minimize kwargs for geomopt

### LINKED ISSUE #171: Add geomopt filter kwargs (constant volume)
Currently the geomopt CLI command takes `opt_kwargs`, but not `filter_kwargs`, preventing certain options being specified.

We can add these manually, as we have done for hydrostatic strain and scalar pressure, but we probably should provide a more general option for anything else.

I think there are two things still to decide:

1. Whether we want to replace `opt_kwargs` with `minimize_kwargs`, which would add another (unnecessary) layer of nesting (e.g. `--minimize-kwargs "{'filter_kwargs': {'hydrostatic_strain' : True}}"`), but mean the options match those for MD, phonons and EoS, or simply add `filter_kwargs`

2. Whether we want an additional explicit option for constant volume, similar to the pressure/hydrostatic strain options

Thoughts (@alinelena)?

ElliottKasoar:
Resolves #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,15 @@ Additional options may be specified. This shares most options with `singlepoint`
 janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --vectors-only --traj 'NaCl-traj.xyz'
 ```
 
-This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
+This allows the cell vectors to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
+
+Further options for the optimizer and filter can be specified using the `--minimize-kwargs` option. For example:
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -110,7 +110,15 @@ Additional options may be specified. This shares most options with `singlepoint`
 janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --calc-kwargs "{'model' : 'small'}" --vectors-only --traj 'NaCl-traj.xyz'
 ```
 
-This allows the cell to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
+This allows the cell vectors to be optimised, allowing only hydrostatic deformation, and saves the optimization trajector in addition to the final structure and log.
```
alinelena, line 113:
```suggestion
This allows the cell vectors to be optimised, allowing only hydrostatic deformation, and saves the optimization trajectory in addition to the final structure and log.
```

PR #175, opened 2024-05-31
Add integrator kwargs to MD

### LINKED ISSUE #172: Add kwargs for MD integrators
We currently cannot set options including `fixcm` for NVT, `mask` for NPT etc.

ElliottKasoar:
Resolves #172

To do:

- [x] Improve CLI test. Possibly just pick a value for one of the outputs to check it doesn't change, as we can't check the mask itself is set as we do in the Python.

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -669,7 +669,7 @@ class NPT(MolecularDynamics):
     file_prefix : Optional[PathLike]
         Prefix for output filenames. Default is inferred from structure, ensemble,
         temperature, and pressure.
-    integrator_kwargs : Optional[dict[str, Any]]
+    ensemble_kwargs : Optional[dict[str, Any]]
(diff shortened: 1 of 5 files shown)


janus_core/calculations/md.py
```diff
@@ -669,6 +669,8 @@ class NPT(MolecularDynamics):
     file_prefix : Optional[PathLike]
         Prefix for output filenames. Default is inferred from structure, ensemble,
         temperature, and pressure.
+    integrator_kwargs : Optional[dict[str, Any]]
```
alinelena, line 672:
```suggestion
    ensemble_kwargs : Optional[dict[str, Any]]
```

alinelena: integrator is slightly lower level than the ensemble and the extra options.

alinelena (approved):

Issue #178, opened 2024-06-03
Update ASE


ElliottKasoar:
A new release of ASE has now been published, which breaks a few tests:

- [x] tests/test_eos_cli.py::test_eos
- [x] tests/test_geom_opt.py::test_hydrostatic_strain
- [x] tests/test_geom_opt.py::test_restart
- [x] tests/test_geomopt_cli.py::test_traj
- [x] tests/test_geomopt_cli.py::test_fully_opt
- [x] tests/test_geomopt_cli.py::test_fully_opt_and_vectors
- [x] tests/test_single_point.py::test_single_point_write
- [x] tests/test_single_point.py::test_single_point_write_kwargs
- [x] tests/test_singlepoint_cli.py::test_singlepoint

These need resolving before ASE can be unpinned (see: #177).

alinelena:
i will suggest to update the ase version and the tests in the same pr.

PR #179, opened 2024-06-03
Allow strings for optimization function names

ElliottKasoar:


janus_core/calculations/geom_opt.py
```diff
@@ -2,17 +2,20 @@
...
 from logging import Logger
 from pathlib import Path
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Union
 import warnings
 
 from ase import Atoms
 from ase.io import read, write
+import ase.optimize
 from ase.optimize import LBFGS
```
alinelena, line 11:
```suggestion
```

pyproject.toml
```diff
@@ -27,7 +27,7 @@ janus = "janus_core.cli.janus:app"
 
 [tool.poetry.dependencies]
 python = "^3.9"
-ase = "^3.22.1"
+ase = "3.22.1"
```
alinelena, line 30: if you fix this there is no FrechetCellFilter... and will simplify things around the default filter.. but of course we will reset this after.

janus_core/calculations/geom_opt.py
```diff
@@ -22,11 +25,39 @@
...
+        ASE filter function, or name of function from ase.filters or ase.constraints.
+        Default is None.
+
+    Returns
+    -------
+    tuple[Callable, Optional[Callable]]
+        Tuple of optimizer function and filter function, if set.
+    """
+    if isinstance(optimizer, str):
+        optimizer = getattr(ase.optimize, optimizer)
```
oerc0122, line 48: You can return a default from `getattr`, so that may be useful for providing more helpful errors, or `try`-`except` to say what the issue is?

janus_core/calculations/geom_opt.py
```diff
@@ -37,13 +68,15 @@ def set_optimizer(
     ----------
     struct : Atoms
         Atoms object to optimize geometry for.
-    filter_func : Optional[callable]
-        Apply constraints to atoms through ASE filter function.
-        Default is `FrechetCellFilter` if available otherwise `ExpCellFilter`.
+    filter_func : Optional[Union[Callable, str]]
+        Fitler function, or name of function from ase.filters or ase.constraints, to
```
oerc0122, line 72:
```suggestion
        Filter function, or name of function from ase.filters or ase.constraints, to
```

janus_core/calculations/geom_opt.py
```diff
@@ -109,13 +146,15 @@ def optimize(  # pylint: disable=too-many-arguments,too-many-locals,too-many-bra
     angle_tolerance : float
         Angle precision for spglib symmetry determination, in degrees. Default is -1.0,
         which means an internally optimized routine is used to judge symmetry.
-    filter_func : Optional[callable]
-        Apply constraints to atoms through ASE filter function.
-        Default is `FrechetCellFilter` if available otherwise `ExpCellFilter`.
+    filter_func : Optional[Union[Callable, str]]
+        Fitler function, or name of function from ase.filters or ase.constraints, to
```
oerc0122, line 150:
```suggestion
        Filter function, or name of function from ase.filters or ase.constraints, to
```

janus_core/cli/geomopt.py
```diff
@@ -33,6 +33,61 @@
...
+    minimize_kwargs : dict[str, Any]
+        Other keyword arguments to pass to geometry optimizer.
+    traj : Optional[str]
+        Path if saving optimization frames.
+    vectors_only : bool
+        Whether to optimize cell vectors, as well as atomic positions, by setting
+        `hydrostatic_strain` in the filter function.
+    pressure : float
+        Scalar pressure when optimizing cell geometry, in bar. Passed to the filter
+        function if either `vectors_only` or `fully_opt` is True.
```
oerc0122, line 56: Does `fully_opt` exist here? is it part of `minimize_kwargs`?

janus_core/cli/geomopt.py
```diff
@@ -33,6 +33,61 @@
...
+    """
+    Set minimize_kwargs dictionary values.
+
+    Parameters
+    ----------
+    minimize_kwargs : dict[str, Any]
+        Other keyword arguments to pass to geometry optimizer.
+    traj : Optional[str]
+        Path if saving optimization frames.
+    vectors_only : bool
```
oerc0122, line 51: `vectors_only` isn't the clearest name to me as to what that means. The help explains it, but I think `cell_only` would possibly be better since the atomic positions are also vectors.

janus_core/cli/geomopt.py
```diff
@@ -42,6 +97,10 @@ def geomopt(
     # numpydoc ignore=PR02
     ctx: Context,
     struct: StructPath,
+    optimizer: Annotated[
+        str,
+        Option(help="Name of ASE optimizer function to use.  [default: LBFGS]"),
```
oerc0122, line 102: Different form of default to other things?

janus_core/cli/geomopt.py
```diff
@@ -157,45 +235,23 @@ def geomopt(
...
-        if "hydrostatic_strain" in minimize_kwargs["filter_kwargs"]:
-            raise ValueError(
-                "'hydrostatic_strain' must be passed through the --vectors-only option"
-            )
-        if "scalar_pressure" in minimize_kwargs["filter_kwargs"]:
+    if fully_opt or vectors_only:
+        # Use default filter unless filter function explicitly passed
+        fully_opt_dict = {"filter_func": filter_func} if filter_func else {}
+    else:
+        if filter_func:
```
oerc0122, line 244:
```suggestion
    elif filter_func:
```

oerc0122 (commented):
Looks reasonable. Just a couple of minor questions.

janus_core/cli/geomopt.py
alinelena: yes this is a tough one since cell_only will mean for sure angles too.. cell_vectors_only maybe?

oerc0122: `cell_lengths`? `lattice_parameters`?

janus_core/calculations/geom_opt.py
ElliottKasoar: I considered a try...except, but when tested the error it outputs `AttributeError: module 'ase.optimize' has no attribute 'test'`, which combined with the docstrings feels quite descriptive.

Anything you would add specifically?

pyproject.toml
ElliottKasoar: This is just from rebasing onto the `pin-ase` branch (#177).

Until the most recent tag, it was necessary to overwrite this line to get the `FrechetCellFilter` anyway, so this just maintains the previous behaviour until fixes from #178 can be implemented, otherwise tests will fail.

janus_core/calculations/geom_opt.py
oerc0122: It's very much a philosophical thing, but to my eye, the user shouldn't know or care when running Janus that they're running a `ase.optimize` optimizer. 

I think we should have something like `raise AttributeError(`f"No such optimizer ({optimizer})") from err`, the reason being that if we add other optimizers somewhere down the line, the error is always consistent internally and can be searched for and controlled by us.

janus_core/cli/geomopt.py
oerc0122: Ah, it's hard to see in GitHub that that line is inside the block (one reason I hate Python's white-space delimiting)

pyproject.toml
alinelena: I wanted to avoid to have it hardcoded to LBFGS somewhere...

janus_core/calculations/geom_opt.py
ElliottKasoar: I suppose my argument would be that at the moment, they have to care, since there are no current alternatives, but I see your point about future proofing, so I'll add something

janus_core/cli/geomopt.py
ElliottKasoar: Ah yes, I changed my mind about relying on geom_opt defaults. Fixed

janus_core/cli/geomopt.py
alinelena: cell_lengths has my vote.

janus_core/cli/geomopt.py
ElliottKasoar: It's implicit at this point. `fully_opt` basically means the filter function isn't `None`, and `hydrostatic_strain` is `False`

janus_core/cli/geomopt.py
ElliottKasoar: Opened as a new issue

pyproject.toml
ElliottKasoar: Sorry, I'm not sure I understand what you're proposing?

ElliottKasoar: Surely that's still hardcoded, just with less accessible help? What's the advantage?

Also, is this still relevant to the commented line change?

```
- ase = "^3.22.1"
+ ase = "3.22.1"
```

janus_core/calculations/geom_opt.py
ElliottKasoar: (Alternatively, `"LBFGS"` to look up the function, but then you lose IDE help etc. which makes setting extra kwargs more tricky, and I'm not sure what we gain, as we only check ase.optimize functions at the moment?)

PR #183, opened 2024-06-05
remove nans for real now

### LINKED ISSUE #182: NaNs in SinglePoint CLI output file
If I run `janus singlepoint --struct tests/data/benzene.xyz --arch mace_mp` the output file contains

```
stress="-inf -inf nan -inf -inf nan nan nan nan"
```

suggesting the data is not entirely being cleaned up.

On the Python side, it doesn't seem to be stored in the `struct.info` or `results` dictionaries, but NaNs are still present in `struct.calc.results`, so I suspect that is the source.

alinelena:
fixes #182

### DIFF
--- a/janus_core/calculations/single_point.py
+++ b/janus_core/calculations/single_point.py
@@ -308,7 +308,7 @@ class SinglePoint(FileNameMixin):
         self,
         results: CalcResults = None,
         properties: Collection[str] = (),
-        invalidate_calc: bool = False,
+        invalidate_calc: bool = True,
     ) -> None:
         """
         Remove NaN and inf values from results and calc.results dictionaries.
@@ -322,7 +322,7 @@ class SinglePoint(FileNameMixin):
         invalidate_calc : bool
             Remove calculator results if True. When True Atoms object loses
             its property methods and true values are in info and arrays.
-            Default is False.
+            Default is True.
         """
         results = results if results else {}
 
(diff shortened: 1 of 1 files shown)


ElliottKasoar (dismissed):
Seems to be fixed.

Only thing to note is I think stress is still duplicated in the output if the values are valid (e.g. for `NaCl.cif`) - `mace_mp_stress` and `stress`, which I assume is related.

It's a more minor issue, but if we did address it, it would probably be an alternative fix to this.

alinelena:
> Seems to be fixed.
> 
> Only thing to note is I think stress is still duplicated in the output if the values are valid (e.g. for `NaCl.cif`) - `mace_mp_stress` and `stress`, which I assume is related.
> 
> It's a more minor issue, but if we did address it, it would probably be an alternative fix to this.

duplication comes from the the fact I do not invalidate the calculator results... if I do that as default shall vanish. happy to do it.

alinelena:
> What about setting `write_results=False` when writing out? (Ideally as a default in the kwargs in case there really is a reason to write it)
> 
> Given that we’ve set the calculator and explicitly handle any results from it, I don’t think we should ever be losing any information that shouldn’t be set elsewhere (info or array).
> 
> I’d probably prefer to keep the results by default if we can.

i think now is fixed. at least I do not see duplicates by default. we do not loose any information.

oerc0122 (approved):

Issue #144, opened 2024-05-13
Split up helpers directory?


ElliottKasoar:
@federicazanca noted that [train.py](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/train.py) in [helpers](https://github.com/stfc/janus-core/tree/main/janus_core/helpers) doesn't exactly fit with other files in the directory, e.g. [utils.py](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/utils.py) or [janus_types.py](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/janus_types.py), which are used more as developer tools.

I originally placed train.py here because it seems closest to [mlip_calculators.py](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/mlip_calculators.py), as opposed to single point calculations etc. in the [calculations](https://github.com/stfc/janus-core/tree/main/janus_core/calculations) directory, but perhaps a third alternative for user-facing modules such as train.py and (to a lesser extent) mlip_calculators.py would be more appropriate?

alinelena:
yes good idea

PR #189, opened 2024-06-13
Add correlator with stress correlations

harveydevereux:
This PR adds an online correlation algorithm (multi-tau), with stress correlations as user observables. The correlations are,

$$C_{ab}(t) = \frac{1}{T-t}\sum_{t'=1}^{T-t}a(t)b(t+t'),$$

for some values $a$ and $b$, and time lags $t$. The lags (and timestep) are implied by the frequency the correlator is updated in ```run``` controled by ```correlation_parameters[.][3]``` (which could probably use being some TypedDict actually...).

A user may ask for a sequence of correlations by,

```python
correlation_kwargs={
            "correlations": [("s_xy", "s_xy")],
            "correlation_parameters": [(1, 10, 1, 1)],
        },
```

where ```s_xy``` is parsed into an ```Observable``` that can poll ```MolecularDynamics::dyn``` with a ```get``` method.

Correlations are written in ```yaml``` format for now a list of correlations (name, value, and lags).

New correlations can be added by providing an extension to ```Observable``` as is done for stress which connects user input to a function to get values from ```Union[Langevin, VelocityVerlet, ASE_NPT]```,

```python
class Stress(Observable):
    def __init__(self, component: str):
        self.component = component
        self._index = self._component_to_index(component, voigt=False)

    def get(self, md: Union[Langevin, VelocityVerlet, ASE_NPT]) -> float:
        return (
            md.atoms.get_stress(include_ideal_gas=True, voigt=False).flatten()[
                self._index
            ]
            / units.bar
        )

    def __str__(self) -> str:
	...
```

any combination of observables is a "valid" correlation.

- ```correlator.py``` could perhaps be split into multiple files

### DIFF
--- a/docs/source/apidoc/janus_core.rst
+++ b/docs/source/apidoc/janus_core.rst
@@ -175,7 +175,7 @@ janus\_core.helpers.post_process module
    :show-inheritance:
 
 janus\_core.helpers.correlator module
----------------------------------------
+-------------------------------------
 
 .. automodule:: janus_core.helpers.correlator
    :members:
@@ -185,7 +185,7 @@ janus\_core.helpers.correlator module
    :show-inheritance:
 
 janus\_core.helpers.observables module
----------------------------------------
+--------------------------------------
 
 .. automodule:: janus_core.helpers.observables
    :members:
--- a/docs/source/apidoc/janus_core.rst
+++ b/docs/source/apidoc/janus_core.rst
@@ -214,7 +214,7 @@ janus\_core.helpers.janus\_types module
    :undoc-members:
    :show-inheritance:
    :exclude-members: clear,copy,setdefault,items,keys,get,pop,popitem,fromkeys,update,values
-
+   :special-members:
 
 janus\_core.helpers.utils module
 --------------------------------
--- a/docs/source/apidoc/janus_core.rst
+++ b/docs/source/apidoc/janus_core.rst
@@ -174,6 +174,16 @@ janus\_core.helpers.post_process module
    :undoc-members:
    :show-inheritance:
 
+janus\_core.helpers.correlator module
+---------------------------------------
+
+.. automodule:: janus_core.helpers.correlator
+   :members:
+   :special-members:
+   :private-members:
+   :undoc-members:
+   :show-inheritance:
+
 janus\_core.helpers.train module
 --------------------------------
 
--- a/docs/source/apidoc/janus_core.rst
+++ b/docs/source/apidoc/janus_core.rst
@@ -184,6 +184,17 @@ janus\_core.helpers.correlator module
    :undoc-members:
    :show-inheritance:
 
+janus\_core.helpers.observables module
+---------------------------------------
+
+.. automodule:: janus_core.helpers.observables
+   :members:
+   :special-members:
+   :private-members:
+   :undoc-members:
+   :show-inheritance:
+
+
 janus\_core.helpers.train module
 --------------------------------
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -162,7 +162,9 @@ Built-in observables are collected within the ``janus_core.helpers.observables``
 
     Stress("xy", False)
 
-A new built-in observables can be implemented by a class with the method::
+A new built-in observables can be implemented by a class with the method:
+
+.. code-block:: python
 
    def __call__(self, atoms: Atoms, *args, **kwargs) -> float
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -150,3 +150,20 @@ Running these tests requires an additional flag to be passed to ``pytest``::
 Alternatively, using ``tox``::
 
     tox -e extra-mlips
+
+Adding a new Observable
+=======================
+
+Additional built-in observable quantities may be added for use within the ``janus_core.helpers.correlator.Correlation``s. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+
+Built-in observables are collected within ``janus_core.helpers.observables``. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas constribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
+
+    Stress("xy", False)
+
+A new built-in observables can be implemented by a class with the method::
+
+   def __call__(self, atoms: Atoms, *args, **kwargs) -> float
+
+The ``__call__`` should contain all the logic for obtaining some ``float`` value from an ``Atoms`` object, alongside optional positional arguments and kwargs. The args and kwargs are set by a user when specifying correlations for a ``janus_core.calculations.md.MolecularDynamics`` run. See also ``janus_core.helpers.janus_types.CorrelationKwargs``. These are set at the instantiation of the ``janus_core.calculations.md.MolecularDynamics`` object and are not modified. These could be used e.g. to specify an observable calculated only from one atom's data.
+
+``janus_core.helpers.observables.Stress`` includes a constructor to take a symbolic component, e.g. ``"xx"`` or ``"yz"``, and determine the index required from ``ase.Atoms.get_stress`` on instantiation for ease of use.
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -156,7 +156,9 @@ Adding a new Observable
 
 Additional built-in observable quantities may be added for use by the ``janus_core.helpers.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
 
-Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
+Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:
+
+.. code-block:: python
 
     Stress("xy", False)
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -156,7 +156,7 @@ Adding a new Observable
 
 Additional built-in observable quantities may be added for use by the ``janus_core.helpers.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
 
-Built-in observables are collected within ``janus_core.helpers.observables``. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas constribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
+Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
 
     Stress("xy", False)
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -154,7 +154,7 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-Additional built-in observable quantities may be added for use within the ``janus_core.helpers.correlator.Correlation``s. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+Additional built-in observable quantities may be added for use by the ``janus_core.helpers.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
 
 Built-in observables are collected within ``janus_core.helpers.observables``. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas constribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
 
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -474,23 +474,18 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
 
     def _parse_correlations(self):
         """Parse correlation kwargs into Correlations."""
-        if not self.correlation_kwargs:
-            self._correlations = None
-        elif len(self.correlation_kwargs) > 0:
-            self._correlations = []
-            for cor in self.correlation_kwargs:
-                self._correlations.append(Correlation(**cor))
+        if self.correlation_kwargs:
+            self._correlations = [Correlation(**cor) for cor in self.correlation_kwargs]
         else:
-            self._correlations = None
+            self._correlations = ()
 
     def _attach_correlations(self):
         """Attach all correlations to self.dyn."""
-        if self._correlations:
-            for i, _ in enumerate(self._correlations):
-                self.dyn.attach(
-                    partial(lambda i: self._correlations[i].update(self.dyn.atoms), i),
-                    self._correlations[i].update_frequency,
-                )
+        for i, _ in enumerate(self._correlations):
+            self.dyn.attach(
+                partial(lambda i: self._correlations[i].update(self.dyn.atoms), i),
+                self._correlations[i].update_frequency,
+            )
 
     def _write_correlations(self):
         """Write out the correlations."""
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -21,9 +21,16 @@ from ase.md.velocitydistribution import (
 )
 from ase.md.verlet import VelocityVerlet
 import numpy as np
+import yaml
 
 from janus_core.calculations.geom_opt import optimize
-from janus_core.helpers.janus_types import Ensembles, PathLike, PostProcessKwargs
+from janus_core.helpers.correlator import Correlation, option_to_observable
+from janus_core.helpers.janus_types import (
+    CorrelationKwargs,
+    Ensembles,
+    PathLike,
+    PostProcessKwargs,
+)
 from janus_core.helpers.log import config_logger, config_tracker
 from janus_core.helpers.post_process import compute_rdf, compute_vaf
 from janus_core.helpers.utils import FileNameMixin
@@ -105,6 +112,8 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
         Time between heating steps, in fs. Default is None, which disables heating.
     post_process_kwargs : Optional[PostProcessKwargs]
         Keyword arguments to control post-processing operations.
+    correlation_kwargs : Optional[CorrelationKwargs]
+        Keyword arguments to control on-the-fly correlations.
     log_kwargs : Optional[dict[str, Any]]
         Keyword arguments to pass to `config_logger`. Default is None.
     tracker_kwargs : Optional[dict[str, Any]]
@@ -173,6 +182,7 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
         temp_step: Optional[float] = None,
         temp_time: Optional[float] = None,
         post_process_kwargs: Optional[PostProcessKwargs] = None,
+        correlation_kwargs: Optional[CorrelationKwargs] = None,
         log_kwargs: Optional[dict[str, Any]] = None,
         tracker_kwargs: Optional[dict[str, Any]] = None,
         seed: Optional[int] = None,
@@ -253,6 +263,8 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
             Time between heating steps, in fs. Default is None, which disables heating.
         post_process_kwargs : Optional[PostProcessKwargs]
             Keyword arguments to control post-processing operations.
+        correlation_kwargs : Optional[CorrelationKwargs]
+            Keyword arguments to control on-the-fly correlations.
         log_kwargs : Optional[dict[str, Any]]
             Keyword arguments to pass to `config_logger`. Default is None.
         tracker_kwargs : Optional[dict[str, Any]]
@@ -289,6 +301,9 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
         self.temp_time = temp_time * units.fs if temp_time else None
         self.post_process_kwargs = (
             post_process_kwargs if post_process_kwargs is not None else {}
+        )
+        self.correlation_kwargs = (
+            correlation_kwargs if correlation_kwargs is not None else {}
         )
         self.log_kwargs = log_kwargs
         self.ensemble = ensemble
@@ -376,6 +391,8 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
             np.random.seed(seed)
             random.seed(seed)
 
+        self._parse_correlations()
+
     def _rotate_restart_files(self) -> None:
         """Rotate restart files."""
         if len(self.restart_files) > self.restarts_to_keep:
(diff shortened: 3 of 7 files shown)


janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,412 @@
...
+
+    @abstractmethod
+    # pylint: disable=C0103
+    def get(self, md: Union[Langevin, VelocityVerlet, ASE_NPT]) -> float:
+        """
+        Get an observable component from md.
+
+        Parameters
+        ----------
+        md : Union[Langevin, VelocityVerlet, ASE_NPT]
```
ElliottKasoar, line 228:
```suggestion
    def get(self, md: Union[Langevin, VelocityVerlet, ASE_NPT]) -> float:
        """
        Get an observable component from md.

        Parameters
        ----------
        md : Union[Langevin, VelocityVerlet, ASE_NPT]
```
I think we've used this elsewhere, but maybe it might be neater to use `MolecularDynamics` from `ase.md.md`, which would make it easier to extend? There's nothing inherent preventing this method working with other MD classes.

tests/test_correlator.py
```diff
@@ -0,0 +1,113 @@
...
+        stress_cor = cor["correlations"][0]["stress_xy-stress_xy"]
+        value, lags = stress_cor["value"], stress_cor["lags"]
+        assert len(value) == len(lags) == 10
+
+        stats = Stats(stats_path)
+        pxy = stats["Pxy"][1:]
+        direct = correlate(pxy, pxy)
+        assert np.mean(direct - value) < TOLERANCE
+    finally:
+        traj_path.unlink(missing_ok=True)
```
ElliottKasoar, line 111: This is fine, but for future reference pytest should (eventually) clean up files in `tmp_path` automatically.

ElliottKasoar (commented):
Thanks a lot for this @harveydevereux! It looks really useful, and I really appreciate the detailed description too. Apologies for taking so long to take a proper look.

I'm in the process of a longer code review, but a couple of initial thoughts:

Is there a way of generalising the observables through the SinglePoint class (or otherwise)? `md.atoms.get_stress` should just be using the MLIP calculator attached, so I think it ought to be able to slot in, and it would mean we could easily, if not immediately, add additional observables. Perhaps the additional options/flattening etc. makes this less straightforward than I'm imagining though.

Going slightly further, if we were to use this for velocity autocorrelation, would we expect this to match the [post processing VAF](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/post_process.py#L186) we can already calculate? It would be interesting to compare, if so, but that could be something to come back to.

harveydevereux:
> Thanks a lot for this @harveydevereux! It looks really useful, and I really appreciate the detailed description too. Apologies for taking so long to take a proper look.
> 
> I'm in the process of a longer code review, but a couple of initial thoughts:
> 
> Is there a way of generalising the observables through the SinglePoint class (or otherwise)? `md.atoms.get_stress` should just be using the MLIP calculator attached, so I think it ought to be able to slot in, and it would mean we could easily, if not immediately, add additional observables. Perhaps the additional options/flattening etc. makes this less straightforward than I'm imagining though.
> 
> Going slightly further, if we were to use this for velocity autocorrelation, would we expect this to match the [post processing VAF](https://github.com/stfc/janus-core/blob/main/janus_core/helpers/post_process.py#L186) we can already calculate? It would be interesting to compare, if so, but that could be something to come back to.

Thanks @ElliottKasoar I will take a look at the SinglePoint class, but it seems like a good way to go for observables.

On VAF this won't get the same results as VAF currently (I think) as it uses ```np.correlate``` with ```same``` rather than ```full```. Full gets the $C_{ab}$ I wrote above averaging across all points in the series considered as origins for the correlation. 

```
np.correlate(momenta[:, j, i], momenta[:, j, i], "same")
```

janus_core/calculations/md.py
```diff
@@ -442,6 +459,48 @@ def _restart_file(self) -> str:
...
+            for cor in range(len(self.correlation_kwargs["correlations"])):
+                obs_a, obs_b = map(
+                    option_to_observable, self.correlation_kwargs["correlations"][cor]
+                )
+                params = self.correlation_kwargs["correlation_parameters"][cor]
+                self._correlations.append(Correlation(obs_a, obs_b, *params))
+
+    def _attach_correlations(self):
+        """Attach all correlations to self.dyn."""
+        # pylint: disable=W0640
```
ElliottKasoar, line 482: Not one I've come across before, but is this definitely not a problem?

If I do 

```
x = [1,2,3]
y = []
for i, _ in enumerate(x):
    y.append(lambda: i*2)
```
then `y[0]() == y[1]() == y[2]() == 2`, so in this case is it not similar? The final `i` would be the only one used in the `lambda` function?

ElliottKasoar:
We should also think about how best to document this, although it may be best to iron out the details a bit first.

Ideally this would probably go in the full user guide, but for now perhaps @harveydevereux you could add a brief example of its use to the MD section of the README, since that'll be the starting point for the full docs.

It would also be great to have something for how you'd go about extending this to additional observables in the developer guide, but that can probably be spun off into a new issue (potentially through a worked example, similar to adding ALIGNN in #194)

janus_core/calculations/md.py
```diff
@@ -442,6 +459,48 @@ def _restart_file(self) -> str:
...
+            ):
+                raise ValueError("Mismatching sizes of correlations and parameters")
+
+            self._correlations = []
+            for cor in range(len(self.correlation_kwargs["correlations"])):
+                obs_a, obs_b = map(
+                    option_to_observable, self.correlation_kwargs["correlations"][cor]
+                )
+                params = self.correlation_kwargs["correlation_parameters"][cor]
+                self._correlations.append(Correlation(obs_a, obs_b, *params))
```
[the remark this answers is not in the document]
oerc0122: Could even do:
```python
            for i, (obs_a, obs_b) in enumerate(map(option_to_observable, self.correlation_kwargs["correlations"])):
                params = self.correlation_kwargs["correlation_parameters"][i]
                self._correlations.append(Correlation(obs_a, obs_b, *params))
```
but you're very much getting into unreadable one-liner territory.
Could `zip` the `correlation_parameters` to avoid the `enumerate` though which makes it cleaner
```suggestion
            for params, (obs_a, obs_b) in zip(
                self.correlation_kwargs["correlation_parameters"]
                map(
                    option_to_observable, self.correlation_kwargs["correlations"])
                ):
                self._correlations.append(Correlation(obs_a, obs_b, *params))
```
At which point the thing could become a comprehension

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,409 @@
...
+                if self._count_correlated[k, i] > 0:
+                    correlation[lag] = (
+                        self._correlation[k, i] / self._count_correlated[k, i]
+                    )
+                    lags[lag] = float(i) * float(self._window) ** k
+                    lag += 1
+        return (correlation[0:lag], lags[0:lag])
+
+
+# pylint: disable=R0903
```
oerc0122, line 176: Preferably use the long names of these issues so it's possible to see what you're trying to do without looking them up.
```suggestion
# pylint: disable=too-few-public-methods
```

janus_core/calculations/md.py
oerc0122, line 482:
```suggestion
        # pylint: disable=cell-var-from-loop
```

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,409 @@
...
+        matrix_voigt = {"xx": 0, "yy": 1, "zz": 2, "yz": 3, "xz": 4, "xy": 5}
+
+        if voigt:
+            return matrix_voigt[component]
+        if component in vector:
+            return vector[component]
+        return matrix[component]
+
+    @abstractmethod
+    # pylint: disable=C0103
```
oerc0122, line 218:
```suggestion
    # pylint: disable=invalid-name
```

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,409 @@
...
+
+        Returns
+        -------
+        int
+            Correlation update frequency.
+        """
+
+        return self._update_frequency
+
+    # pylint: disable=C0103
```
oerc0122, line 301:
```suggestion
    # pylint: disable=invalid-name
```

tests/test_correlator.py
```diff
@@ -0,0 +1,113 @@
...
+from janus_core.helpers.stats import Stats
+
+DATA_PATH = Path(__file__).parent / "data"
+MODEL_PATH = Path(__file__).parent / "models" / "mace_mp_small.model"
+runner = CliRunner()
+
+TOLERANCE = 1e-16
+
+
+# pylint: disable=C0103
```
oerc0122, line 22:
```suggestion
# pylint: disable=invalid-name
```

oerc0122 (commented):
Not much to add beyond Elliott's review. 

Just a couple of minor suggestions to avoid looking up what things mean.

janus_core/helpers/correlator.py
harveydevereux: good point, didn't realise that could be done for some reason, I'll add these in

tests/test_correlator.py
```diff
@@ -0,0 +1,113 @@
...
+        cor.update(val, val)
+    correlation, lags = cor.get()
+
+    direct = correlate(signal, signal, fft=False)
+    fft = correlate(signal, signal, fft=True)
+
+    assert len(correlation) == len(lags)
+    assert all(lags == range(points))
+    assert np.mean(direct - correlation) < TOLERANCE
+    assert np.mean(fft - correlation) < TOLERANCE
```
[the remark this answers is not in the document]
harveydevereux: approx is now used as this is clearly wrong as well

janus_core/helpers/janus_types.py
```diff
@@ -75,6 +75,15 @@ class PostProcessKwargs(TypedDict, total=False):
     vaf_output_file: Optional[PathLike]
 
 
+class CorrelationKwargs(TypedDict, total=False):
+    """Arguments for on-the-fly correlations."""
+
+    # Pairs of indexed observables, ('s_xy', 's_xy')
+    correlations: Sequence[tuple[str, str]]
+    # Blocks, points, window, and update frequency
+    correlation_parameters: Sequence[tuple[int, int, int, int]]
```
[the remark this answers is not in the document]
harveydevereux: Alongside the discussion on the actual names of these keys below,

I was thinking something like this so all values are clearer,

```
class CorrelationKwargs(TypedDict, total=False):
    observable_a: str
    observable_b: str
    blocks: int
    points: int
    window: int
    update_frequency: int
```

and

```
correlation_kwargs: Optional[Sequence[CorrelationKwargs]]
```

but then these are just a sequence of args I guess.

Not sure what else to write other than "observable_a". The idea is its calculating $\langle A(t)B(t+\tau)\rangle$. It could just be a.

janus_core/calculations/md.py
harveydevereux: yes, I think so, ```functools.partial``` may do it?

janus_core/calculations/md.py
```diff
@@ -453,6 +471,39 @@ def _restart_file(self) -> str:
...
         )
 
+    def _parse_correlations(self):
+        """Parse correlation kwargs into Correlations."""
+        if not self.correlation_kwargs:
+            self._correlations = None
+        elif len(self.correlation_kwargs) > 0:
+            self._correlations = []
+            for cor in self.correlation_kwargs:
+                self._correlations.append(Correlation(**cor))
```
oerc0122, line 482:
```suggestion
            self._correlations = [Correlation(**cor) for cor in self.correlation_kwargs]
```

janus_core/calculations/md.py
```diff
@@ -453,6 +471,39 @@ def _restart_file(self) -> str:
             f"res-{step}.xyz", f"T{self.temp}", prefix_override=self.restart_stem
         )
 
+    def _parse_correlations(self):
+        """Parse correlation kwargs into Correlations."""
+        if not self.correlation_kwargs:
+            self._correlations = None
+        elif len(self.correlation_kwargs) > 0:
```
oerc0122, line 479:
```suggestion
        if self.correlation_kwargs:
```

janus_core/calculations/md.py
```diff
@@ -453,6 +471,39 @@ def _restart_file(self) -> str:
...
+            self._correlations = []
+            for cor in self.correlation_kwargs:
+                self._correlations.append(Correlation(**cor))
+        else:
+            self._correlations = None
```
oerc0122, line 484: If instead of assigning `None` here, you assign `()` you can still exploit the `if self._correlations`, but don't have to as loops over it will find no elements, and we can reduce the number of checks and flatten some blocks.

janus_core/calculations/md.py
```diff
@@ -453,6 +471,39 @@ def _restart_file(self) -> str:
...
+
+    def _attach_correlations(self):
+        """Attach all correlations to self.dyn."""
+        if self._correlations:
+            for i, _ in enumerate(self._correlations):
+                self.dyn.attach(
+                    partial(lambda i: self._correlations[i].update(self.dyn.atoms), i),
+                    self._correlations[i].update_frequency,
+                )
```
oerc0122, line 493: e.g. here
```suggestion
        for i, _ in enumerate(self._correlations):
            self.dyn.attach(
                partial(lambda i: self._correlations[i].update(self.dyn.atoms), i),
                 self._correlations[i].update_frequency,
            )
```

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,287 @@
...
+        self._max_block_used = 0
+        self._min_dist = self._points / self._averaging
+
+        self._accumulator = np.zeros((self._blocks, 2))
+        self._count_accumulated = np.zeros(self._blocks).astype(int)
+        self._shift_index = np.zeros(self._blocks).astype(int)
+        self._shift = np.zeros((self._blocks, self._points, 2))
+        self._shift_not_null = np.zeros((self._blocks, self._points)).astype(bool)
+        self._correlation = np.zeros((self._blocks, self._points))
+        self._count_correlated = np.zeros((self._blocks, self._points)).astype(int)
```
oerc0122, line 55:
```suggestion
        self._accumulator = np.zeros((self._blocks, 2))
        self._count_accumulated = np.zeros(self._blocks, dtype=int)
        self._shift_index = np.zeros(self._blocks, dtype=int)
        self._shift = np.zeros((self._blocks, self._points, 2))
        self._shift_not_null = np.zeros((self._blocks, self._points), dtype=bool)
        self._correlation = np.zeros((self._blocks, self._points))
        self._count_correlated = np.zeros((self._blocks, self._points), dtype=int)
```

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,287 @@
...
+        return self._shift_not_null[block, p_i] and self._shift_not_null[block, p_j]
+
+    def get(self) -> tuple[Iterable[float], Iterable[float]]:
+        """
+        Obtain the correlation and lag times.
+
+        Returns
+        -------
+        tuple[Iterable[float], Iterable[float]]
+            The correlation and lags.
```
oerc0122, line 155: Under numpy docs you can address the tuple elements independently by a name like you would for parameters.

janus_core/helpers/janus_types.py
```diff
@@ -75,6 +75,35 @@ class PostProcessKwargs(TypedDict, total=False):
...
+class CorrelationKwargs(TypedDict, total=False):
+    """Arguments for on-the-fly correlations."""
+
+    a: tuple[Observable, dict]
+    b: tuple[Observable, dict]
+    name: str
+    blocks: int
+    points: int
+    averaging: int
+    update_frequency: int
```
oerc0122: May want to `#:` document these parameters

oerc0122 (changes requested):

oerc0122 (approved):

janus_core/calculations/md.py
```diff
@@ -713,6 +761,9 @@ def run(self) -> None:
         if self.post_process_kwargs:
             self._post_process()
 
+        if self.correlation_kwargs:
+            self._write_correlations()
```
ElliottKasoar, line 765: It's not really a problem, but why is this handled differently to parsing/attaching (checking for self.correlation_kwargs here, rather than a check within the function)?

janus_core/helpers/correlator.py
```diff
@@ -0,0 +1,298 @@
...
+        ----------
+        a : float
+            Newly observed value of left correland.
+        b : float
+            Newly observed value of right correland.
+        """
+        self._add(a, b, 0)
+
+    # pylint: disable=invalid-name
+    def _add(self, a: float, b: float, block: int):
```
ElliottKasoar, line 72: Would `_propagate` or similar be more descriptive name?

tests/test_correlator.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+    def user_observable_b(atoms: Atoms) -> float:
+        """User specified getter for correlation"""
+        return atoms.get_stress(include_ideal_gas=True, voigt=True)[-1] / bar
+
+    nve = NVE(
+        struct=single_point.struct,
+        temp=300.0,
+        steps=100,
```
ElliottKasoar, line 100: Do you need 100 steps to get a useful correlation? It would be nice if it were a little shorter if possible, but it's fine if the results are not distinguishable enough otherwise for the tolerance.

ElliottKasoar (changes requested):
Looking good!

The other comments I've added are pretty minor and all optional.

A few things that would be useful to clarify:

1. Do we not want a CLI interface for this?
2. Did we not want some built-in observables (e.g. `user_observable_a`/stress), which is probably a requirement for (1)
3. Can the use of this be documented in the README in some form (if we do (1) and (2), an example of the CLI with a description of the kwargs is probably sufficient)

Maybe this can all be a separate PR, but I don't think a user currently would discover or easily figure out how to use this as it is.

tests/test_correlator.py
harveydevereux: Given the test is now seeded we could choose say 10 steps for speed. For actual use probably more is needed for science, but this is just a test.

Originally it was because being unseeded the error value was unstable at << 100 steps.

harveydevereux:
> Looking good!
> 
> The other comments I've added are pretty minor and all optional.
> 
> A few things that would be useful to clarify:
> 
> 1. Do we not want a CLI interface for this?
> 2. Did we not want some built-in observables (e.g. `user_observable_a`/stress), which is probably a requirement for (1)
> 3. Can the use of this be documented in the README in some form (if we do (1) and (2), an example of the CLI with a description of the kwargs is probably sufficient)
> 
> Maybe this can all be a separate PR, but I don't think a user currently would discover or easily figure out how to use this as it is.

Definitely we document in this pr, I wanted to be sure the api/features were settled first, I think we are approaching that point *but* where do the (2) built-in ones go? I thought something like
```python
import stress from janus_core.helpers.correlator.observables
``` 
would not seem insane, but @alinelena has opinions, it would mean correlator becoming a folder

For CLI its interesting, do we want the user to be able to specify just built-ins? Or maybe have arbitrary code evaluated from a ```eval("some code")```, or somehow import python files

oerc0122:
> > Looking good!
> > The other comments I've added are pretty minor and all optional.
> > A few things that would be useful to clarify:
> > 
> > 1. Do we not want a CLI interface for this?
> > 2. Did we not want some built-in observables (e.g. `user_observable_a`/stress), which is probably a requirement for (1)
> > 3. Can the use of this be documented in the README in some form (if we do (1) and (2), an example of the CLI with a description of the kwargs is probably sufficient)
> > 
> > Maybe this can all be a separate PR, but I don't think a user currently would discover or easily figure out how to use this as it is.
> 
> Definitely we document in this pr, I wanted to be sure the api/features were settled first, I think we are approaching that point _but_ where do the (2) built-in ones go? I thought something like
> 
> ```python
> import stress from janus_core.helpers.correlator.observables
> ```
> 
> would not seem insane, but @alinelena has opinions, it would mean correlator becoming a folder
> 
> For CLI its interesting, do we want the user to be able to specify just built-ins? Or maybe have arbitrary code evaluated from a `eval("some code")`, or somehow import python files

`eval`ing code is not a good plan as if you're writing that on the command line why aren't you writing it in a python script? CLI will always be more restrictive than direct interfaces. We may want to build a correlation function factory which imports and constructs a standard `Correlator` class from a string by importing the appropriate libraries allowing a user to define their `Correlator` class or correlation function in that folder and import it dynamically.

In general they should be able to specify arguments in a way consistent with the rest of the code using the `typer` interfaces.

harveydevereux:
> > > Looking good!
> > > The other comments I've added are pretty minor and all optional.
> > > A few things that would be useful to clarify:
> > > 
> > > 1. Do we not want a CLI interface for this?
> > > 2. Did we not want some built-in observables (e.g. `user_observable_a`/stress), which is probably a requirement for (1)
> > > 3. Can the use of this be documented in the README in some form (if we do (1) and (2), an example of the CLI with a description of the kwargs is probably sufficient)
> > > 
> > > Maybe this can all be a separate PR, but I don't think a user currently would discover or easily figure out how to use this as it is.
> > 
> > 
> > Definitely we document in this pr, I wanted to be sure the api/features were settled first, I think we are approaching that point _but_ where do the (2) built-in ones go? I thought something like
> > ```python
> > import stress from janus_core.helpers.correlator.observables
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would not seem insane, but @alinelena has opinions, it would mean correlator becoming a folder
> > For CLI its interesting, do we want the user to be able to specify just built-ins? Or maybe have arbitrary code evaluated from a `eval("some code")`, or somehow import python files
> 
> `eval`ing code is not a good plan as if you're writing that on the command line why aren't you writing it in a python script? CLI will always be more restrictive than direct interfaces. We may want to build a correlation function factory which imports and constructs a standard `Correlator` class from a string by importing the appropriate libraries allowing a user to define their `Correlator` class or correlation function in that folder and import it dynamically.
> 
> In general they should be able to specify arguments in a way consistent with the rest of the code using the `typer` interfaces.

I'm aware it is a bad idea to do it that way, my point is this PR ill become quite large if we want to figure out all of that out here as well as implementing a correlator. The factory sounds like a good idea to do it but as I think you agree we move it to later work.

oerc0122:
> as I think you agree we move it to later work.

@harveydevereux Indeed, see the edit I made between you "quote replying" and posting.

ElliottKasoar:
> Definitely we document in this pr, I wanted to be sure the api/features were settled first, I think we are approaching that point _but_ where do the (2) built-in ones go? I thought something like
> 
> ```python
> import stress from janus_core.helpers.correlator.observables
> ```
> 
> would not seem insane, but @alinelena has opinions, it would mean correlator becoming a folder
> 
> For CLI its interesting, do we want the user to be able to specify just built-ins? Or maybe have arbitrary code evaluated from a `eval("some code")`, or somehow import python files

I would suggest a `janus_core.helpers.observables` module is probably fine for now. Helpers needs splitting up anyway (#144) so we can worry about subdirectories later if it makes sense.

I definitely agree that user-defined observables doesn't need to be addressed in this PR, and it may never be possible via the CLI. 

How I (and I think @alinelena) imagined the CLI was probably similar to what we do for functions in geom_opt.py, where we call `getattr(module, function_name)` to import a function based on its name from a module (e.g. `observables`).

For the CLI, this I'd imagine this would be passed through kwargs (`--filter-kwargs "{'obs_1': 'function_name', ...}"`, since otherwise it pollutes the main options too much.

We can come back to that though.

The main point was that without pre-defined functions, and the CLI that uses them, it's harder to write simple documentation.

If we want to get this merged and out of the way, perhaps @harveydevereux you could add something to the dev tutorial for how you approached writing the custom functions, which will remain useful.

docs/source/apidoc/janus_core.rst
```diff
@@ -174,6 +174,27 @@ janus\_core.helpers.post_process module
...
+
+.. automodule:: janus_core.helpers.correlator
+   :members:
+   :special-members:
+   :private-members:
+   :undoc-members:
+   :show-inheritance:
+
+janus\_core.helpers.observables module
+---------------------------------------
```
ElliottKasoar, line 188:
```suggestion
-------------------------------------

.. automodule:: janus_core.helpers.correlator
   :members:
   :special-members:
   :private-members:
   :undoc-members:
   :show-inheritance:

janus\_core.helpers.observables module
--------------------------------------
```

docs/source/developer_guide/tutorial.rst
```diff
@@ -150,3 +150,20 @@ Running these tests requires an additional flag to be passed to ``pytest``::
 Alternatively, using ``tox``::
 
     tox -e extra-mlips
+
+Adding a new Observable
+=======================
+
+Additional built-in observable quantities may be added for use within the ``janus_core.helpers.correlator.Correlation``s. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+
+Built-in observables are collected within ``janus_core.helpers.observables``. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas constribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
```
ElliottKasoar, line 159:
```suggestion
Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
```

docs/source/developer_guide/tutorial.rst
```diff
@@ -150,3 +150,20 @@ Running these tests requires an additional flag to be passed to ``pytest``::
 Alternatively, using ``tox``::
 
     tox -e extra-mlips
+
+Adding a new Observable
+=======================
+
+Additional built-in observable quantities may be added for use within the ``janus_core.helpers.correlator.Correlation``s. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
```
ElliottKasoar, line 157:
```suggestion
Additional built-in observable quantities may be added for use by the ``janus_core.helpers.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
```

ElliottKasoar (changes requested):
Looks great, thanks @harveydevereux!

Just a couple of minor typos/formatting suggestions for the docs

harveydevereux:
> Looks great, thanks @harveydevereux!
> 
> Just a couple of minor typos/formatting suggestions for the docs

Thanks for catching these @ElliottKasoar, I forgot to do the proof read...

ElliottKasoar (dismissed):

docs/source/developer_guide/tutorial.rst
```diff
@@ -150,3 +150,20 @@ Running these tests requires an additional flag to be passed to ``pytest``::
 Alternatively, using ``tox``::
 
     tox -e extra-mlips
+
+Adding a new Observable
+=======================
+
+Additional built-in observable quantities may be added for use by the ``janus_core.helpers.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+
+Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as::
```
oerc0122, line 159:
```suggestion
Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:

.. code-block:: python
```

And elsewhere for highlighting

janus_core/helpers/observables.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Parameters
+    ----------
+    component : str
+        Symbol for tensor components, xx, yy, etc.
+    include_ideal_gas : bool
+        Calculate with the ideal gas contribution.
+    """
+
+    def __init__(self, component: str, include_ideal_gas: bool = True):
```
oerc0122, line 21: Very specific setting, recommend making that kwarg only
```suggestion
    def __init__(self, component: str, *, include_ideal_gas: bool = True):
```

harveydevereux: Yes I think that is better thanks

oerc0122 (dismissed):

docs/source/developer_guide/tutorial.rst
```diff
@@ -150,3 +150,22 @@ Running these tests requires an additional flag to be passed to ``pytest``::
...
+
+Built-in observables are collected within the ``janus_core.helpers.observables`` module. For example the ``janus_core.helpers.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:
+
+.. code-block:: python
+
+    Stress("xy", False)
+
+A new built-in observables can be implemented by a class with the method::
+
+   def __call__(self, atoms: Atoms, *args, **kwargs) -> float
```
ElliottKasoar, line 167:
```suggestion
A new built-in observables can be implemented by a class with the method:

.. code-block:: python

   def __call__(self, atoms: Atoms, *args, **kwargs) -> float
```
To be consistent with the code block above

ElliottKasoar (approved):
Assuming I didn't break the tests, I think we should merge this. Thanks for all the work and discussion!

We can always address anything we missed when we decide how to implement a CLI etc.

oerc0122 (approved):
Agreed. LGTM.

PR #192, opened 2024-06-19
Promote CHGNET and M3GNET to full dependencies

### LINKED ISSUE #161: Add CI testing for extra MLIPs
Now that we have a few more tests for non-MACE MLIPs (calculator tests, and now EoS tests), we should add their tests into the Actions CI.

This requires adding the `--run-extra-mlips` flag for pytest, but will also require changing the installation to include the extra-mlips poetry group.

I'm still getting inconsistent errors for the EoS tests, so this would be a useful second check for that too.

ElliottKasoar:
Resolves #161

- Moves chgnet and m3gnet dependencies from optional to full dependencies, and removes pytest decorators, which prevent related tests running by default.
- Adds singlepoint tests for both

Note: Currently requires pinning dgl, as attempting to install the most recent version gives me an error: `ERROR: Could not find a version that satisfies the requirement dgl==2.2.1 (from versions: 0.1.0, 0.1.2, 0.1.3, 1.0.0, 1.0.1, 1.0.4, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 2.1.0)` (see issue linked in #161).


ElliottKasoar:
The README probably could do with updating too to reflect this, including a comment about setting the PyTorch dtype if using multiple MLIPs.

alinelena (approved):

PR #194, opened 2024-07-02
Add ALIGNN as optional MLIP

ElliottKasoar:
Adds optional integration with [ALIGNN](https://github.com/usnistgov/alignn), to be used as worked developer example for how to add a new MLIP to janus-core.

One other thing to consider, which could affect this implementation/the example slightly (although it would be a separate PR), is if we want to standardise the input model.

Currently only `device` is standardised, but if we wanted to do the same for the path to the model (which may be useful/essential for aiida-mlip, @federicazanca?), it would be sensible to do it now, before adding further MLIPs.

To do:

- [x] Add notes to docs explaining steps taken

Depends on #195, so may need a minor rebase.

### DIFF
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -1,3 +1,83 @@
 ========
 Tutorial
 ========
+
+Adding a new MLIP
+=================
+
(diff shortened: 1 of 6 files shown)


alinelena (dismissed):
may need rebase

ElliottKasoar:
(Rebased)

alinelena (approved):

PR #197, opened 2024-07-05
Downgrade seekpath for aiida-quantumespresso compatibility

### LINKED ISSUE #196: dft code for aiida incompatible with janus
If we need to do a training workflow using aiida-quantumespresso (dft optimisation with qe --> training with janus), we need to add aiida-quantumespresso to the dependencies of aiida-mlip (or at least have it installed at the same time as janus cause we need to use both). 
The problem is that janus and aiida-qe have some dependencies conflict:

```
And because janus-core (0.6.1b0) depends on seekpath (>=2.1.0,<3.0.0)
 and janus-core (0.6.2) depends on seekpath (>=2.1.0,<3.0.0), janus-core (>=v0.6.0b0,<0.7.0) requires seekpath (>=2.1.0,<3.0.0). 
And because aiida-quantumespresso (>=4.6.0,<5.0.0) requires seekpath (>=1.9.3,<2.0) (1), aiida-quantumespresso (>=4.6.0,<5.0.0) is incompatible with janus-core (>=v0.6.0b0,<0.7.0)
``` 

Can we change the version of seekpath in janus-core?
Should I use another dft code?
Other solutions?

ElliottKasoar:
Resolves #196

This doesn't seem to affect any phonopy functionality from initial testing e.g. from the [phonopy documentation](https://phonopy.github.io/phonopy/phonopy-module.html):

> Automatic selection of band paths using [SeeK-path](https://seekpath.readthedocs.io/en/latest/) is invoked by

>
>```
>phonon.auto_band_structure()
>```

Running 
```
janus phonons --struct tests/data/NaCl.cif --band --plot-to-file
```

still correctly generates `NaCl-auto_bands.svg` via `plot_band_structure`, and running

```
poetry add aiida-quantumespresso
```

also installs aiida-quantumespresso successfully.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ pyyaml = "^6.0.1"
 typer = "^0.9.0"
 typer-config = "^1.4.0"
 phonopy = "^2.23.1"
-seekpath = "^2.1.0"
+seekpath = "^1.9.7"
 spglib = "^2.3.0"
 torch-dftd = "^0.4.0"
 


federicazanca:
If we like to downgrade seekpath I'm fine with it, if we don't like it and want to find another solution we can try? (maybe tell aiida-qe to upgrade it but I think they use it quite a bit so it might just be easier for us to downgrade)

federicazanca (approved):

PR #201, opened 2024-07-10
Change pressure from bar to GPa

### LINKED ISSUE #186: Update/make pressure units consistent
Currently we mostly use bars, such as for NPT MD input and stats, and `scalar_pressure` for the `geomopt` CLI command.

However, GPa is probably preferable, and we should check we're consistent with how/when units correctly set.

For example, geometry optimization via Python does not set units for `scalar_pressure`, but perhaps we should, particularly as this may be a kwarg passed from other CLI options e.g. phonons.

ElliottKasoar:
Resolves #186

1. Changes units from bar to GPa in MD and geomopt (CLI)
2. Moves setting units for `scalar_pressure` when carrying out geometry optimization into calculation module. Previously this was only set for the geomopt CLI command, but this way it's set for anything that calls geomopt e.g. phonons, MD, EoS etc. and for the Python interface, so it's more consistent.

Is there a good way of signposting this? As I mention in #186, for any arbitrary `kwargs` passed to functions, ASE units should be assumed, so this is an exception.

When it's an explicit variable we make it clear, but is there an obvious way when it's passed via (potentially quite convoluted) kwargs e.g. `--minimize-kwargs "{'filter_kwargs': {'scalar_pressure' : x}}"`?

Also worth checking if I've missed obvious cases for other inconsistencies.

### DIFF
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Any, Callable, Optional, Union
 import warnings
 
-from ase import Atoms, filters
+from ase import Atoms, filters, units
 from ase.filters import FrechetCellFilter
 from ase.io import read, write
 import ase.optimize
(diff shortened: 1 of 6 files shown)


alinelena (approved):

alinelena:
@harveydevereux once this is in you will need to keep an eye... since there is a small change of units from bar to GPa as defaults. also is consistent now when passed to ase.

PR #208, opened 2024-07-11
Fix setting devices

### LINKED ISSUE #95: Add option for intel devices?
MACE seems to have [added support](https://github.com/ACEsuit/mace/pull/356) for the "xpu" device (at least for training) via the `intel_extension_for_pytorch` package.

I imagine this could be useful for anyone using the new Intel PVCs via [Dawn](https://www.hpc.cam.ac.uk/d-w-n), and elsewhere.

We can certainly add `xpu` as a valid device type straightforwardly. I think everything else can be handled by the individual MLIP calculators, so that may be all that's required?

ElliottKasoar:
Resolves #95

It's not in the main branch of MACE yet, but as mentioned in #95, there's no reason not to allow `xpu` in theory, so I've added it.

More importantly, in the process of extending the allowed devices, I realised that `device` was not passed to MACE-MP, or MACE-OFF, so this should now be fixed.

Also adds a test that the chosen `device` is valid, as the PyTorch error isn't particularly clear:

```
RuntimeError: don't know how to restore data location of torch.storage.UntypedStorage (tagged with invalid)
```

This should perhaps be three PRs, but they're interconnected and otherwise very minor...

### DIFF
--- a/janus_core/helpers/janus_types.py
+++ b/janus_core/helpers/janus_types.py
@@ -91,7 +91,7 @@ EoSNames = Literal[
 
 # Janus specific
 Architectures = Literal["mace", "mace_mp", "mace_off", "m3gnet", "chgnet"]
-Devices = Literal["cpu", "cuda", "mps"]
+Devices = Literal["cpu", "cuda", "mps", "xpu"]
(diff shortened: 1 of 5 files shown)


oerc0122 (approved):
If it does what it's meant to.

PR #210, opened 2024-07-12
correct units for pressure, previously were not set for geometry opti…

### LINKED ISSUE #209: unit for pressure set after filter set
https://github.com/stfc/janus-core/blob/c1653dd600aa6ee988f1f779e2121a0d26ee4acd/janus_core/calculations/geom_opt.py#L107

this will result in pressure being set after the filter was initialised.

alinelena:
…misation, also update the test for geom opt, now checks only the final energy. fixes #209

### DIFF
--- a/tests/test_geom_opt.py
+++ b/tests/test_geom_opt.py
@@ -2,7 +2,7 @@
 
 from pathlib import Path
 
-from ase.filters import FrechetCellFilter
+from ase.filters import FrechetCellFilter, UnitCellFilter
 from ase.io import read
 import pytest
 
@@ -15,21 +15,21 @@ DATA_PATH = Path(__file__).parent / "data"
 MODEL_PATH = Path(__file__).parent / "models" / "mace_mp_small.model"
 
 test_data = [
-    ("mace", "NaCl.cif", -27.046359959669214, {}),
-    ("mace", "NaCl.cif", -27.04636199814088, {"fmax": 0.001}),
-    ("mace", "NaCl.cif", -27.0463392211678, {"filter_func": FrechetCellFilter}),
+    ("mace", "NaCl.cif", -27.046349600581266, {}),
+    ("mace", "NaCl.cif", -27.046361983699768, {"fmax": 0.001}),
+    ("mace", "NaCl.cif", -27.04633922116779, {"filter_func": UnitCellFilter}),
     ("mace", "H2O.cif", -14.051389496520015, {"filter_func": None}),
     (
         "mace",
         "NaCl.cif",
-        -26.727162796978437,
+        -26.727162796978426,
         {
             "fmax": 0.001,
             "filter_func": FrechetCellFilter,
             "filter_kwargs": {"scalar_pressure": 5.0},
         },
     ),
-    ("mace", "NaCl.cif", -27.046353221978332, {"opt_kwargs": {"alpha": 100}}),
+    ("mace", "NaCl.cif", -27.04634943785021, {"opt_kwargs": {"alpha": 100}}),
 ]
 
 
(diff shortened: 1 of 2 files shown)


tests/test_geom_opt.py
```diff
@@ -21,13 +17,17 @@
 test_data = [
     ("mace", "NaCl.cif", -27.046359959669214, {}),
     ("mace", "NaCl.cif", -27.04636199814088, {"fmax": 0.001}),
-    ("mace", "NaCl.cif", -27.0463392211678, {"filter_func": UnitCellFilter}),
+    ("mace", "NaCl.cif", -27.0463392211678, {"filter_func": FrechetCellFilter}),
```
ElliottKasoar, line 20: I don't think we want this, as it's the default, so this doesn't really check it can be set?

tests/test_geom_opt.py
```diff
@@ -42,11 +42,8 @@ def test_optimize(architecture, struct_path, expected, kwargs):
         calc_kwargs={"model": MODEL_PATH},
     )
 
-    init_energy = single_point.run("energy")["energy"]
-
     struct = optimize(single_point.struct, **kwargs)
 
-    assert struct.get_potential_energy() < init_energy
```
ElliottKasoar, line 49: Why do we not want this?

oerc0122 (approved):
I agree with @ElliottKasoar 's comments and want those fixing first, but otherwise this looks ok.

tests/test_geom_opt.py
alinelena: because under pressure initial energy of the unrelaxed structure can be smaller than the final one...

tests/test_geom_opt.py
alinelena: do not we have 3.23 as minimum version? so we will not really have to check it since is there.

ElliottKasoar: It's not checking the availability of the functions from ASE, it was doing a very rough check that `filter_func` can be set as a different function in general.

tests/test_geom_opt.py
ElliottKasoar: Ah, true, but perhaps then the pressure test should be separated? It seems a shame to remove one of the relatively few physically meaningful tests we have for the other cases.

alinelena: > Ah, true, but perhaps then the pressure test should be separated? It seems a shame to remove one of the relatively few physically meaningful tests we have for the other cases.

I do not think we get any more info by checking the order if the final energy is correct.

ElliottKasoar: Not at the moment, no.

I think when I was writing these tests, we were looking for ways to reduce the arbitrary numbers as much as possible. In an ideal world we'd probably just have physical tests based on the parameters, which is where this came in, as a potential starting point for that.

For the tests that we're trying to distinguish, going much further in a sensible way probably isn't possible though, and you're right that on its own it's not adding anything.

ElliottKasoar (approved):
@alinelena it wouldn't let me suggest the change, so I've just pushed it... Feel free to revert if you don't like it, but essentially my point is 

```
("mace", "NaCl.cif", -27.046349600581266, {}),
("mace", "NaCl.cif", -27.04633922116779, {"filter_func": FrechetCellFilter})
```
are exactly the same test, but with different expected energies, which doesn't make sense.

Setting `UnitCellFilter` , which is a different test, wasn't giving a result different enough to be picked up by the tolerence of pytest.approx, so I also reduced the tolerance.

I.e. originally with these energies at the default tolerance, I can swap in either function and the tests pass. Now swapping them should cause them to fail.

oerc0122 (approved):

PR #213, opened 2024-07-15
Unify model path inputs for MLIPs

### LINKED ISSUE #202: Standardise model path keyword
As mentioned in #194, it's probably useful to standardise the (optional) path to the input model between MLIPs.

Currently only `device` is standardised, but given that for provenance (e.g. via aiida-mlip), we always want a way to specify it, and you'd expect it as an option in some form anyway, it would be useful to make it consistent.

It's probably worth aliasing the native names too, where appropriate, as we already do for `model`/`model_paths` with MACE.

ElliottKasoar:
Resolves #202

Slightly more complicated than I'd planned, but I think this works, and should be backwards compatible too.

Essentially this aims to:

- Ensure a path to the model can always be specified via `model_path`, which wasn't the case before, other than for MACE
- Allow parameters from each MLIP to be more or less swapped in for `model_path` e.g. `model_paths`, `model` etc
  - We could restrict this more e.g. now allowing labels (`"small"` to be passed as a `model_path`), but generally the MLIPs treat paths and labels in the same way, so I think it makes for us to as well
- Expands the calculator tests to ensure paths can be used, which requires adding a copy of each model
  - I think it's an important test, but we could potentially download them at test time if we don't want them in the repository


janus_core/helpers/mlip_calculators.py
```diff
@@ -6,17 +6,72 @@
...
+        if sum(kwarg in kwargs for kwarg in model_path_kwargs) > 1:
+            raise ValueError(
+                "Only one of 'model', 'model_paths', 'potential', and 'path' can be "
+                "specified"
+            )
+
+        # Set model_path from kwargs if any are specified
+        for kwarg in model_path_kwargs:
+            if kwarg in kwargs:
+                model_path = kwargs.pop(kwarg, None)
```
oerc0122, line 62: `kwargs` is a dictionary which means we can simplify this a lot using set operations.

`kwargs.keys()`  supports set operations, this means if we make `model_path_kwargs` a set too, we can check which kwargs are defined by:
```python
present = kwargs.keys() & {"model", "model_paths", "potential", "path"}
```
This leaves us with:
```suggestion
    present = kwargs.keys() & {"model", "model_paths", "potential", "path"}

    # Use model_path if specified, but check not also specified via kwargs
    if model_path and present:
        raise ValueError(
            "`model_path` cannot be used in combination with 'model', "
            "'model_paths', 'potential', or 'path'"
        )
    elif len(present) > 1:
        # Check at most one suitable kwarg is specified
        raise ValueError(
            "Only one of 'model', 'model_paths', 'potential', and 'path' can be "
            "specified"
        )
    elif present:
        # Set model_path from kwargs if any are specified
        model_path = kwargs[present[0]]
    else:
        return None
```
Alternatively you can check if `len(present) == 1` if you want to align, or set `model_path = None` if you want to have single-return.

janus_core/helpers/mlip_calculators.py
```diff
@@ -6,17 +6,72 @@
...
+
+    # Convert to path if file/directory exists
+    if isinstance(model_path, (Path, str)) and Path(model_path).expanduser().exists():
```
oerc0122, line 65: Why do you need to `expanduser` here?

tests/test_mlip_calculators.py
```diff
@@ -2,63 +2,137 @@
...
+
+MACE_OFF_PATH = Path(__file__).parent / "models" / "MACE-OFF23_small.model"
+
+M3GNET_DIR_PATH = Path(__file__).parent / "models" / "M3GNet-MP-2021.2.8-DIRECT-PES/"
+M3GNET_MODEL_PATH = (
+    Path(__file__).parent / "models" / "M3GNet-MP-2021.2.8-DIRECT-PES/model.pt"
+)
+M3GNET_POTENTIAL = load_model(path=M3GNET_DIR_PATH)
+
+CHGNET_PATH = Path(__file__).parent / "models" / "chgnet_0.3.0_e29f68s314m37.pth.tar"
```
oerc0122, line 21: Would it be cleaner to define:
Would it be cleaner to define:
```suggestion

MODEL_PATH = Path(__file__).parent / "models"

MACE_MP_PATH = MODEL_PATH / "mace_mp_small.model"

MACE_OFF_PATH = MODEL_PATH / "MACE-OFF23_small.model"

M3GNET_DIR_PATH = MODEL_PATH / "M3GNet-MP-2021.2.8-DIRECT-PES"
M3GNET_MODEL_PATH = M3GNET_DIR_PATH / "model.pt"
M3GNET_POTENTIAL = load_model(path=M3GNET_DIR_PATH)

CHGNET_PATH = MODEL_PATH / "chgnet_0.3.0_e29f68s314m37.pth.tar"
```

tests/test_mlip_calculators.py
```diff
@@ -2,63 +2,137 @@
...
+
+@pytest.mark.extra_mlips
+@pytest.mark.parametrize(
+    "architecture, device, kwargs",
+    [
+        ("alignn", "cpu", {}),
+        ("alignn", "cpu", {"model_path": "tests/models/v5.27.2024"}),
+        ("alignn", "cpu", {"model_path": "tests/models/v5.27.2024/best_model.pt"}),
+        ("alignn", "cpu", {"model": "alignnff_wt10"}),
+        ("alignn", "cpu", {"path": "tests/models/v5.27.2024"}),
```
oerc0122, line 108: Do these also need to go in the top model paths?

tests/test_mlip_calculators.py
```diff
@@ -2,63 +2,137 @@
...
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        {
+            "model_path": "tests/models/v5.27.2024/best_model.pt",
+            "model": "tests/models/v5.27.2024/best_model.pt",
+        },
+        {
+            "model_path": "tests/models/v5.27.2024/best_model.pt",
+            "path": "tests/models/v5.27.2024/best_model.pt",
```
oerc0122, line 131: And these?

oerc0122 (changes requested):

janus_core/helpers/mlip_calculators.py
ElliottKasoar: So that the path can be wrapped in quotes e.g. in the CLI `--model-path "~/.cache/mace/5f5yavf3"` didn't work, but `--model-path ~/.cache/mace/5f5yavf3` did, which didn't seem very intuitive, and may be annoying with special characters

tests/test_mlip_calculators.py
ElliottKasoar: They could do, but while they're optional, limited to a couple of tests, and intended partly as an example of adding a new MLIP, I think being explicit and somewhat self contained is probably nicer.

If you disagree I don't mind moving them

janus_core/helpers/mlip_calculators.py
oerc0122: I meant in this line in particular, does `Path` not resolve that correctly to check for existence. If not, that's interesting.

ElliottKasoar:
```
In [22]: Path("~/.cache/mace/5f5yavf3").expanduser().exists()
Out[22]: True

In [23]: Path("~/.cache/mace/5f5yavf3").exists()
Out[23]: False
```
Is this what you mean?

tests/test_mlip_calculators.py
oerc0122: I think at the very least we *may* want to use `MODEL_PATH` above if that's that correct path?

ElliottKasoar: Yes I was just thinking that, I'll add it in

janus_core/helpers/mlip_calculators.py
oerc0122: Yep. Weird, but I think that settles this.

tests/test_mlip_calculators.py
```diff
@@ -2,63 +2,137 @@
...
+
+@pytest.mark.extra_mlips
+@pytest.mark.parametrize(
+    "architecture, device, kwargs",
+    [
+        ("alignn", "cpu", {}),
+        ("alignn", "cpu", {"model_path": {MODEL_PATH} / "v5.27.2024"}),
+        ("alignn", "cpu", {"model_path": {MODEL_PATH} / "v5.27.2024/best_model.pt"}),
+        ("alignn", "cpu", {"model": "alignnff_wt10"}),
+        ("alignn", "cpu", {"path": {MODEL_PATH} / "v5.27.2024"}),
```
oerc0122, line 108: Do you mean the curly braces here? Won't that create a `set`?

ElliottKasoar: Well spotted, thanks! I initially thought it was going to be a f string, and didn't clean up properly

docs/source/developer_guide/tutorial.rst
```diff
@@ -50,35 +50,94 @@ This is done within the ``janus_core.helpers.mlip_calculators`` module, if ``arc
...
+
+Converting ``model_path`` into ``path`` is a minimum requirement, but we also aim to facilitate options native to the MLIP, including a default model, where possible:
+
+- If ``model_path`` refers to the path to the model file, as is expected by ``choose_calculator``, we define ``path`` as the parent directory of ``model_path``
+- If ``model_path`` refers to the directory of the model, closer to the typical use of ``AlignnAtomwiseCalculator``, we define ``path`` as ``model_path``
+- If ``model_path`` refers to a model label, similar to the MACE ``"small"`` models, we try loading the model using ALIGNN's ``get_figshare_model_ff``
+- If ``model_path`` is ``None``, we use the ALIGNN's ``default_path``
 
-As ``device`` is a parameter of ``choose_calculator``, it must either be passed explicitly, or added to the ``kwargs``, with the appropriate keyword expected by the calculator. When using CHGNet, for example, we instead set ``use_device=device``.
+Note: ``model_path`` will already be a ``pathlib.Path`` object, if the path exists.
```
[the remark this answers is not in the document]
ElliottKasoar: Yep, thanks!

janus_core/helpers/mlip_calculators.py
ElliottKasoar: Thanks, that's useful! It did feel quite convoluted.

ElliottKasoar: I made a couple of changes to pop the `kwarg` when we set `model_path`, to avoid passing it twice, and removing the `return None`, to make sure `model_path` is converted into a `Path` as well.

oerc0122 (approved):

PR #214, opened 2024-07-15
Test Windows and MacOS

ElliottKasoar:
I thought we had an issue, but can't find it...

- Adds Windows and MacOS (arm and intel, I think, based on [this list](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories))

These are experimental, as we can't currently guarantee support/functionality, but this makes it easier to see and fix issues, and hopefully we can get to a point where they are all supported fully.

Currently there seem to be at least a couple that we could address soon:

- Installing dgl for windows
- Setting the correct device for `chgnet` (on MacOS or any OS?)

(I don't suggest fixing them as part of this PR)

Unfortunately I think even "allowed" failures lose the green tick, which is annoying.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,17 +7,21 @@ jobs:
   tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
     runs-on: "${{ matrix.os }}"
+    continue-on-error: ${{ matrix.experimental }}
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
+        experimental: [false]
         include:
           - os: macos-latest
             python-version: "3.11"
+            experimental: true
           - os: windows-latest
             python-version: "3.11"
+            experimental: true
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


ElliottKasoar:
Errors on latest MacOS should hopefully be fixed in the next CHGNet release (https://github.com/CederGroupHub/chgnet/issues/181).

ElliottKasoar:
The Windows issue seems to be due to the availability of built distributions on PyPI. The most recent releases are:

[2.2.1](https://pypi.org/project/dgl/2.2.1/#files), which only seems to include Windows built distributions
[2.2.0](https://pypi.org/project/dgl/2.2.0/#files), which only seems to include MacOS built distributions
[2.1.0](https://pypi.org/project/dgl/2.1.0/#files), which only seems to include linux and MacOS
[2.0.0](https://pypi.org/project/dgl/2.0.0/#files), seems to include built distributions for Windows, MacOS and linux

So a potential solution to ensure compatibility might be reverting to 2.0.0.

Alternatives include:

- Follow their suggested installation, but this requires specifying the torch version, and will not work Winows/MacOS after 2.2.1: 
  - `pip install  dgl -f https://data.dgl.ai/wheels/torch-2.1/repo.html`

- Install from source, which requires different instructions for each OS
  - https://docs.dgl.ai/install/index.html#install-from-source

ElliottKasoar:
These will be tested on self-hosted runners e.g. via #334.

Issue #124, opened 2024-04-26
Refactor MD?


ElliottKasoar:
Wondering whether a base `MolecularDynamics` object should have a start/end temp, given `NVE`. Not sure what might be better though.

_Originally posted by @oerc0122 in https://github.com/stfc/janus-core/pull/116#discussion_r1574839775_

ElliottKasoar:
More generally, from the same PR:

> I think at some point this module may need to be refactored, so we may revisit some of the interactions e.g. between NPT and NVT_NH, but I think for now this looks good, and particularly having (mostly) settled on the file naming convention we want is perhaps the most important thing.

ElliottKasoar:
See also comments in #216 for potential ways to tidy up stats e.g. [here](https://github.com/stfc/janus-core/pull/216#discussion_r1679315082) and [here](https://github.com/stfc/janus-core/pull/216#discussion_r1679309646).

PR #216, opened 2024-07-16
Tidy MD stats variable names

### LINKED ISSUE #199: Update MD stat names
Some names related to the statistics calculated and written during MD still make some references to "log", e.g. [get_log_stats](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L463) and [get_log_header](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L446).

Describing these only in terms of "stats", and "log" referring only to the Python logging, would be clearer.

ElliottKasoar:
Resolves #199

Internal tidy up of function/variables names to avoid confusion with logging/stats file in MD.

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -129,10 +129,10 @@ class MolecularDynamics(FileNameMixin):  # pylint: disable=too-many-instance-att
     -------
     run()
         Run molecular dynamics simulation and/or heating ramp.
-    get_log_stats()
-        Get thermodynamical statistics to be written to molecular dynamics log.
-    get_log_header()
-        Get header string for molecular dynamics log.
+    get_stats()
+        Get thermodynamical statistics to be written to file.
+    get_stats_header()
+        Get header string for molecular dynamics statistics.
     """
 
     def __init__(  # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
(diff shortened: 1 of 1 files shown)


janus_core/calculations/md.py
```diff
@@ -454,26 +454,26 @@ def _restart_file(self) -> str:
         )
 
     @staticmethod
-    def get_log_header() -> str:
+    def get_stats_header() -> str:
```
oerc0122, line 457: Do we also want to provide `@property` accessors?

janus_core/calculations/md.py
```diff
@@ -518,7 +518,7 @@ def get_log_stats(self) -> str:
             density = 0.0
             pressure_tensor = np.zeros(6)
 
-        log_stats = (
+        stats = (
```
oerc0122, line 521: I think this might actually be better expressed as something like:
```python
stats = " ".join(f"{step:10d}",
                 f"{real_time.total_seconds():.3f}",
                 f"{time:13.2f}",
                 f"{e_pot:.8e}", ...
```
or
```
stats = " ".join(format(step, "10d"),
                 format(real_time.total_seconds(), ".3f"),
                 format(time, "13.2f"),
                 format(e_pot, "8e"), ...
```
But I'm not certain and not for this PR.

It does mean that for `pressure_tensor` you could do:
```python
*map(lambda x: format(x, "8e"), pressure_tensor)
```
I don't know.

janus_core/calculations/md.py
```diff
@@ -992,30 +992,30 @@ def __init__(
...
 
         Returns
         -------
         str
             Thermodynamical statistics to be written out.
         """
-        log_stats = MolecularDynamics.get_log_stats(self)
-        return log_stats + f" {self.temp}"
+        stats = MolecularDynamics.get_stats(self)
+        return stats + f" {self.temp}"
```
oerc0122, line 1005: Feels like there should be a nicer way to do this like a dict of {name: prop} which is iterated over to build the headers/values, but again, not for this PR.

oerc0122 (approved):
Just a couple of thoughts.

alinelena (approved):

janus_core/calculations/md.py
alinelena: is this addressed?

ElliottKasoar: Can we make this a property?

Of the options:

```
@staticmethod
@property
def _stats_header() -> str:

@property
@staticmethod
def _stats_header() -> str:
    
@property
def _stats_header(self) -> str:
```

If I try to access `self._stats_header`, first returns a property object, the second throws an error (`takes 0 positional arguments but 1 was given`), and the third works, but means we can't call `MolecularDynamics._stats_header`, which we use elsewhere.

ElliottKasoar: I may be missing something, but I think I'll leave it as beyond the scope of this anyway, since this was just relabelling, and potentially come back to it fir #124 or otherwise

Issue #206, opened 2024-07-11
Tidy/unify calculations options


ElliottKasoar:
At some point before a first release, we probably want to do a pass over all the main calculations, including:

- [x] Possibly converting `geom_opt` and `eos` into classes
- [x] Consider the interaction and need for `file_prefix` and `struct_name`
  - In some cases e.g. `eos`, I think they may actually do exactly the same thing
  - I'd propose potentially removing `struct_name` as an option completely, unless there's a clear benefit, to simplify the setting of file names
- [x] Check other settings are consistent between calculations
- [x] Ensure outputs from `ase.io.write` are consistent with singlepoint (i.e. results in `info`, they may or may not also be saved in `calc.results`)
- [x] Change default output extensions to .extxyz

PR #218, opened 2024-07-16
Fix setting calculator for list of Atoms

### LINKED ISSUE #217: Multiple calculators not independent
This is almost entirely hidden by the changes introduced by #176, but if `invalidate_calc` is set to `False` in single_point.py, then the results dictionary produced is incorrect:

```
from janus_core.calculations.single_point import SinglePoint
from ase.io import read

atoms_list = read("tests/data/benzene-traj.xyz", index=":")

sp = SinglePoint(atoms_list, struct_name="benzene")
sp.run()

print(sp.struct[0].calc.results["energy"])
print(sp.struct[1].calc.results["energy"])
```

Prints:

```
-74.80419118083256
-74.80419118083256
```

This appears to be because the same calculator object is shared between each structure.

ElliottKasoar:
Resolves #217

Not easily testable at the moment as `invalidate_calc` isn't settable, but I'll address that in a future PR for #206

### DIFF
--- a/janus_core/calculations/single_point.py
+++ b/janus_core/calculations/single_point.py
@@ -1,6 +1,7 @@
 """Prepare and perform single point calculations."""
 
 from collections.abc import Collection
+from copy import copy
 from pathlib import Path
 from typing import Any, Optional
 
@@ -219,7 +220,7 @@ class SinglePoint(FileNameMixin):  # pylint: disable=too-many-instance-attribute
 
         if isinstance(self.struct, list):
             for struct in self.struct:
-                struct.calc = calculator
+                struct.calc = copy(calculator)
             # Return single Atoms object if only one image in list
             if len(self.struct) == 1:
                 self.struct = self.struct[0]
--- a/janus_core/calculations/single_point.py
+++ b/janus_core/calculations/single_point.py
@@ -1,7 +1,7 @@
 """Prepare and perform single point calculations."""
 
 from collections.abc import Collection
-from copy import copy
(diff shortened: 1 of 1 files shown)


ElliottKasoar:
I think the test failures may be from the new mace update to 0.3.6 - [this action](https://github.com/ElliottKasoar/janus-core/actions/runs/9910557037/) for a commit on main passed previously, but fails on re-run, and main tests fail for me locally now after updating my dependencies:

```
  • Updating lightning-utilities (0.11.3.post0 -> 0.11.5)
  • Updating monty (2024.5.24 -> 2024.7.12)
  • Updating cachetools (5.3.3 -> 5.4.0)
  • Updating pynvml (11.5.0 -> 11.5.3)
  • Updating mace-torch (0.3.5 -> 0.3.6)
  • Updating phonopy (2.26.3 -> 2.26.6)
```

janus_core/calculations/single_point.py
```diff
@@ -219,7 +220,7 @@ def set_calculator(
 
         if isinstance(self.struct, list):
             for struct in self.struct:
-                struct.calc = calculator
+                struct.calc = copy(calculator)
```
oerc0122, line 223: Does this want to be a copy or a deepcopy?

ElliottKasoar: Good question, this seemed to be sufficient for the results, but perhaps a deepcopy would be safest just in case.

oerc0122 (approved):

PR #221, opened 2024-07-17
Update Atoms info from calculated results

ElliottKasoar:
Initial work aiming to address part of #206:

> -  Ensure outputs from ase.io.write are consistent with singlepoint (i.e. results in info, they may or may not also be saved in calc.results)

Originally, I thought this would be implemented as a custom `write` function that also moved results to info, then called `ase.io.write`, but we want this even if we're not writing to file.

There's probably a better name than `output_atoms`, but the idea is a generalised version of what was previously done in single_point - move `Atoms.calc.results` to `Atoms.info` or `Atoms.arrays`.

To do:

- [ ] More testing (possibly directly on the utils function) to ensure this behaves as expected, but so far seems to work with minimal changes
- [ ] Extend to other calculations (currently only done for single_point and geom_opt, to give an idea of the changes required)
- [ ] Update all CLIs (e.g. dealing with `invalidate_calc`, either as new option or via `write_kwargs`, since we don't return an Atoms object


janus_core/helpers/utils.py
```diff
@@ -229,3 +231,87 @@ def dict_remove_hyphens(dictionary: dict) -> dict:
...
+
+
+def results_to_info(
+    struct: Atoms,
+    *,
+    properties: Collection[Properties] = (),
+    invalidate_calc: bool = True,
+) -> None:
+    """
+    Copy or move MLIP calculated results to info.
```
oerc0122, line 243: What does it mean to be in "info"?

ElliottKasoar: The `Atoms.info` dict. I can make it clearer in the various places I reference it

oerc0122:
Following discussions, recommend developing separate script to automate generation of different MLIP records for database. That way files will only be read/written once and extra data appended to info object as it goes. 

In terms of that, current implementation is fine provided non-arch key is also added so ASE automatically recognises the last run's parameters (this may want to be disabled in aforementioned script) as well as the tagged key. 

Also suggest adding `arch` to the info block under the same grounds.

ElliottKasoar:
With the last set of changes, this now:

- Includes an `"arch"` info label with the same tag as the labelled properties (`[arch]_energy`) etc.
- Defaults to both including both the `info` and original `calc.results` dicts
  - I don't especially like this, but losing the ability to use built in functions like `Atoms.get_stress` etc. by default seems really quite inconvenient at the moment

The current default of copying results to the info dict is still not obviously the right choice for me personally, but I think @alinelena disagrees.

janus_core/calculations/single_point.py
```diff
@@ -386,11 +306,15 @@ def run(
...
 
         for prop in properties:
-            if prop not in ["energy", "forces", "stress"]:
+            if prop not in get_args(Properties):
                 raise NotImplementedError(
                     f"Property '{prop}' cannot currently be calculated."
                 )
 
+        # If none specified, get all valid properties
+        if len(properties) == 0:
```
oerc0122, line 315: Could this be `not properties`?

janus_core/helpers/utils.py
```diff
@@ -229,3 +231,89 @@ def dict_remove_hyphens(dictionary: dict) -> dict:
...
+
+    if struct.calc:
+        # Set default architecture from calculator name
+        arch = struct.calc.parameters["arch"]
+        struct.info["arch"] = arch
+
+        for key in properties & struct.calc.results.keys():
+            tag = f"{arch}_{key}"
+            value = struct.calc.results[key]
+            if key == "forces":
```
oerc0122, line 266: Are the ASE compatible labels already in there?

janus_core/calculations/single_point.py
ElliottKasoar: Yes, probably. This was mostly a like for like replacement for `if "energy" in properties or len(properties) == 0:` etc.

I've added a similar change in #225, where I'm doing more refactoring of SinglePoint, but happy to do it here if it makes more sense.

janus_core/helpers/utils.py
ElliottKasoar: At a point where we'd want to copy/move them to `Atoms.info`, they'd be in `Atoms.calc.results`, if that's what you're asking.

The dictionary is automatically populated by running calculations e.g.

```
In [26]: x.calc.results
Out[26]: {}

In [27]: x.get_potential_energy()
^[[AOut[27]: -27.035127799332745

In [28]: x.calc.results
Out[28]:
{'energy': -27.035127799332745,
 'free_energy': -27.035127799332745,
 'node_energy': array([-0.86285876, -0.32487255, -0.86285876, -0.32487255, -0.86285876,
        -0.32487255, -0.86285876, -0.32487255]),
 'forces': array([[ 2.08166817e-17,  0.00000000e+00,  1.19298552e-20],
        [ 2.08166817e-17,  6.93889390e-18,  2.08504865e-17],
        [ 1.73472348e-17, -8.67361738e-19, -7.80625564e-18],
        [-6.93889390e-18, -1.73472348e-18,  5.20417043e-18],
        [ 1.73472348e-17,  1.04083409e-17, -1.04083409e-17],
        [-6.93889390e-18,  0.00000000e+00,  6.93889390e-18],
        [-8.67361738e-19,  2.60208521e-18, -5.06858339e-18],
        [-1.73472348e-18,  6.93889390e-18, -1.38777878e-17]]),
 'stress': array([-4.78327600e-03, -4.78327600e-03, -4.78327600e-03,  1.08000967e-19,
        -2.74004242e-19, -2.04504710e-19])}
```

With `write_results=True` for `ase.io.write`, these are then read back in with a generic `SinglePointCalculator` that make the same results still accessible:

```
In [35]: write(images=x, write_results=True, filename="test.xyz")

In [36]: y = read("test.xyz")

In [37]: y.calc
Out[37]: <ase.calculators.singlepoint.SinglePointCalculator at 0x7f9468830810>

In [38]: y.calc.results
Out[38]:
{'forces': array([[ 0.,  0.,  0.],
        [ 0.,  0.,  0.],
        [ 0., -0., -0.],
        [-0., -0.,  0.],
        [ 0.,  0., -0.],
        [-0.,  0.,  0.],
        [-0.,  0., -0.],
        [-0.,  0., -0.]]),
 'energy': -27.035127799332745,
 'free_energy': -27.035127799332745,
 'stress': array([-4.78327600e-03, -4.78327600e-03, -4.78327600e-03,  1.08000967e-19,
        -2.74004242e-19, -2.04504710e-19])}
```

oerc0122: Sounds good. Just wanted to check.

janus_core/helpers/utils.py
```diff
@@ -229,3 +231,89 @@ def dict_remove_hyphens(dictionary: dict) -> dict:
...
+
+    if set_info:
+        for image in images:
+            results_to_info(
+                image, properties=properties, invalidate_calc=invalidate_calc
+            )
+    else:
+        # Label architecture even if not copying results to info
+        for image in images:
+            if image.calc:
```
oerc0122, line 315: Is this likely to be called without an `image.calc`? I assume the answer is yes.

janus_core/helpers/utils.py
```diff
@@ -229,3 +231,89 @@ def dict_remove_hyphens(dictionary: dict) -> dict:
...
+                image.info["arch"] = image.calc.parameters["arch"]
+
+    if write_results:
+        write(images=images, write_results=not invalidate_calc, **kwargs)
```
oerc0122, line 319: Does `invalidate_calc` have a default? I would say it's confusing that this is explicitly `not invalidate_calc`, I can't work out why intuitively.

oerc0122 (dismissed):
All looks good, just want to be sure of a couple of things.

janus_core/helpers/utils.py
ElliottKasoar: Not particularly likely, actually.

None of the cases where it's called at the moment should be False here, since attaching a calculator is generally the first thing we do before anything else, and there's no reason not to use `ase.io.write` (or write at all) if we haven't calculated anything yet.

Arguably this could be a value error `if not image.calc`, but I think I wrote it like this to make it generally a safe swap for `ase.io.write`, even if practically it shouldn't be needed usually.

janus_core/helpers/utils.py
ElliottKasoar: `invalidate_calc` defaults to False now, meaning we keep the `calc.results` dictionary for each images normally.

If `invalidate_calc is True`, then there's no point setting `write(write_results=True)`, as there will be no results to write for any of the images.

You could argue for a separate option if `invalidate_calc is False`, which I did have originally.

However, one issue with this is that the options become confusing, due to the clashes with the `write_results` flag that triggers `write`, which actually made it impossible to pass `write_results` to `write` in some situations at the moment.

That's not a good enough reason on its own, but I then felt that the point of these changes more generally is to keep the Atoms object and written files consistent, so if we write out an image that had a results dict, it makes sense that we can read it in with the results dict, which requires `write(write_results=True)`.

ElliottKasoar: `not invalidate_calc` is now only a default, so it can be explicitly overwritten if the user wants (although if `invalidate_calc` is True, this won't do anything)

ElliottKasoar:
Relatively minor change to swap unpacking `write_kwargs` when it's passed to `output_structs` to passing it as the dict, as otherwise there could be keyword clashes e.g. from `write_results`.

janus_core/calculations/single_point.py
```diff
@@ -236,15 +237,11 @@ def _get_potential_energy(self) -> MaybeList[float]:
...
             energies = [struct.get_potential_energy() for struct in self.struct]
-            for struct, energy in zip(self.struct, energies):
-                struct.info[tag] = energy
             return energies
 
         energy = self.struct.get_potential_energy()
-        self.struct.info[tag] = energy
         return energy
 
     def _get_forces(self) -> MaybeList[ndarray]:
```
oerc0122, line 247: Do these want to be `MaybeSequence` if we're allowing `Sequence`s?

janus_core/helpers/utils.py
```diff
@@ -229,3 +236,100 @@ def dict_remove_hyphens(dictionary: dict) -> dict:
...
+    """
+    # Separate kwargs for output_structs from kwargs for ase.io.write
+    # This assumes values passed via kwargs have priority over passed parameters
+    write_kwargs = write_kwargs if write_kwargs else {}
+    if "set_info" in write_kwargs:
+        set_info = write_kwargs.pop("set_info")
+    if "properties" in write_kwargs:
+        properties = write_kwargs.pop("properties")
+    if "invalidate_calc" in write_kwargs:
+        invalidate_calc = write_kwargs.pop("invalidate_calc")
```
oerc0122, line 317: [pop](https://docs.python.org/3/library/stdtypes.html#dict.pop) allows defaults.
```suggestion
    set_info = write_kwargs.pop("set_info", set_info)
    properties = write_kwargs.pop("properties", properties)
    invalidate_calc = write_kwargs.pop("invalidate_calc", invalidate_calc)
```

tests/test_utils.py
```diff
@@ -46,3 +60,81 @@ def test_dict_remove_hyphens():
...
+    struct.get_stress()
+
+    # Check all expected keys are in results
+    assert all(key in struct.calc.results for key in results_keys)
+
+    # Check results and MLIP-labelled keys are not in info or arrays
+    assert len(results_keys & struct.info.keys()) == 0
+    assert len(results_keys & struct.arrays.keys()) == 0
+    assert len(label_keys & struct.info.keys()) == 0
+    assert len(label_keys & struct.arrays.keys()) == 0
```
oerc0122, line 101: Could use `not` rather than `len==0` but same thing really.

tests/test_utils.py
```diff
@@ -46,3 +60,81 @@ def test_dict_remove_hyphens():
...
+        properties=properties,
+        invalidate_calc=invalidate_calc,
+        write_kwargs=write_kwargs,
+    )
+
+    # Check results keys depend on invalidate_calc
+    if invalidate_calc:
+        assert len(results_keys & struct.calc.results.keys()) == 0
+    else:
+        assert all(key in struct.calc.results for key in results_keys)
```
oerc0122, line 115: Could just check set equality or subset.

janus_core/calculations/single_point.py
ElliottKasoar: `Atoms` can be a sequence, but we still only return their forces as a list, so I think this is still accurate at the moment.

janus_core/helpers/utils.py
ElliottKasoar: Good point, thanks

oerc0122 (dismissed):
If this works, I'm happy with it, other comments are mostly aesthetic.

tests/test_utils.py
ElliottKasoar: Ah cool, thanks!

oerc0122 (approved):

PR #225, opened 2024-07-22
Tidy calculations

ElliottKasoar:
Depends on #224 and #221 (will need rebasing)

Aims to address part of #206:

> - Consider the interaction and need for file_prefix and struct_name
>   - In some cases e.g. eos, I think they may actually do exactly the same thing
>   - I'd propose potentially removing struct_name as an option completely, unless there's a clear benefit, to simplify the setting of file names
 > - Check other settings are consistent between calculations

New changes are all in the final commit: ([Tidy calculations](https://github.com/stfc/janus-core/commit/1fd66b16e33a2598dd7ed809e802356a68139e2f)):

- Removes `struct_name`, since `file_prefix` performs essentially the same purpose (`FileNameMixin` should be updated similarly, although we may want to add in something for an optional filepath somehow)
- Refactors `SinglePoint` to be more consistent with other calculations, so all kwargs passed on init normally, but can also be overwritten when calling `run`
- Refactors phonons to clarify order of function calls, and includes `run` function that takes a similar sequence of calculations to the CLI interface.

### DIFF
--- a/janus_core/calculations/eos.py
+++ b/janus_core/calculations/eos.py
@@ -24,8 +24,6 @@ class EoS(FileNameMixin):
     ----------
     struct : Atoms
         Structure to calculate equation of state for.
-    struct_name : Optional[str]
-        Name of structure. Default is None.
     min_volume : float
         Minimum volume scale factor. Default is 0.95.
     max_volume : float
@@ -59,7 +57,6 @@ class EoS(FileNameMixin):
     def __init__(  # pylint: disable=too-many-arguments,too-many-locals
         self,
         struct: Atoms,
-        struct_name: Optional[str] = None,
         min_volume: float = 0.95,
         max_volume: float = 1.05,
         n_volumes: int = 7,
@@ -81,8 +78,6 @@ class EoS(FileNameMixin):
         ----------
         struct : Atoms
             Structure.
-        struct_name : Optional[str]
-            Name of structure. Default is None.
         min_volume : float
             Minimum volume scale factor. Default is 0.95.
         max_volume : float
@@ -160,7 +155,7 @@ class EoS(FileNameMixin):
         self.logger = config_logger(**log_kwargs)
         self.tracker = config_tracker(self.logger, **tracker_kwargs)
 
-        FileNameMixin.__init__(self, struct, struct_name, file_prefix)
+        FileNameMixin.__init__(self, struct, None, file_prefix)
 
         self.write_kwargs.setdefault(
             "filename",
--- a/janus_core/calculations/eos.py
+++ b/janus_core/calculations/eos.py
@@ -125,8 +125,8 @@ class EoS(FileNameMixin):
         lattice_scalars : NDArray[float64]
             Lattice scalars of generated structures.
         """
-        [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs] = none_to_dict(
-            [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs]
+        (minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs) = none_to_dict(
+            (minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs)
         )
 
         self.struct = struct
--- a/janus_core/calculations/eos.py
+++ b/janus_core/calculations/eos.py
@@ -125,6 +125,10 @@ class EoS(FileNameMixin):
         lattice_scalars : NDArray[float64]
             Lattice scalars of generated structures.
         """
+        [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs] = none_to_dict(
+            [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs]
+        )
+
         self.struct = struct
         self.min_volume = min_volume
         self.max_volume = max_volume
@@ -132,17 +136,13 @@ class EoS(FileNameMixin):
         self.eos_type = eos_type
         self.minimize = minimize
         self.minimize_all = minimize_all
+        self.minimize_kwargs = minimize_kwargs
         self.write_results = write_results
         self.write_structures = write_structures
-        self.file_prefix = file_prefix
-
-        [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs] = none_to_dict(
-            [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs]
-        )
-        self.minimize_kwargs = minimize_kwargs
         self.write_kwargs = write_kwargs
         self.log_kwargs = log_kwargs
 
+        # Validate parameters
         if not isinstance(struct, Atoms):
             if isinstance(struct, Sequence) and isinstance(struct[0], Atoms):
                 raise NotImplementedError(
@@ -150,17 +150,6 @@ class EoS(FileNameMixin):
                     "object at a time currently"
                 )
             raise ValueError("`struct` must be an ASE Atoms object")
-
-        log_kwargs.setdefault("name", __name__)
-        self.logger = config_logger(**log_kwargs)
-        self.tracker = config_tracker(self.logger, **tracker_kwargs)
-
-        FileNameMixin.__init__(self, struct, None, file_prefix)
-
-        self.write_kwargs.setdefault(
-            "filename",
-            self._build_filename("generated.extxyz").absolute(),
-        )
 
         if (
             (self.minimize or self.minimize_all)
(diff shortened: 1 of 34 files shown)


janus_core/calculations/eos.py
```diff
@@ -129,35 +124,24 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-locals
         lattice_scalars : NDArray[float64]
             Lattice scalars of generated structures.
         """
+        [minimize_kwargs, write_kwargs, log_kwargs, tracker_kwargs] = none_to_dict(
```
oerc0122, line 127: Using square brackets here looks like *shudder* MATLAB. Don't need these.

janus_core/calculations/md.py
```diff
@@ -282,13 +275,33 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-locals,too-many-sta
             Random seed used by numpy.random and random functions, such as in Langevin.
             Default is None.
         """
+        [
```
oerc0122, line 278: Prefer round braces as this is to me tuple unpacking with a fixed set or an operation rather than a variable list.

janus_core/calculations/single_point.py
```diff
@@ -228,6 +227,44 @@ def set_calculator(
...
+    def properties(self, value: MaybeSequence[Properties]) -> None:
+        """
+        Setter for `properties`.
+
+        Parameters
+        ----------
+        value : MaybeSequence[Properties]
+            Physical properties to be calculated.
+        """
+        self._properties = value
```
oerc0122, line 252: Probably want the setting to be the last thing done? Not sure how this plays with certain failure modes.

janus_core/calculations/single_point.py
```diff
@@ -278,60 +315,47 @@ def _get_stress(self) -> MaybeList[ndarray]:
...
-
-        # If none specified, get all valid properties
-        if len(properties) == 0:
-            properties = get_args(Properties)
-
-        write_kwargs = write_kwargs if write_kwargs else {}
+        # Parameters can be overwritten, otherwise default to values from instantiation
+        properties = self.properties
+        write_results = write_results if write_results else self.write_results
+        write_kwargs = write_kwargs if write_kwargs else self.write_kwargs
```
oerc0122, line 345: Do thse need to be `if write_xxx is not None` because if I pass in `False` I presumably want to override what's in `self`?

janus_core/cli/phonons.py
```diff
@@ -238,10 +230,23 @@ def phonons(
...
+
+    calcs = []
+    if bands:
+        calcs.append("bands")
+    if thermal:
+        calcs.append("thermal")
+    if dos:
+        calcs.append("dos")
+    if pdos:
+        calcs.append("pdos")
```
oerc0122, line 244: Just for fun, some horrible metaprogramming which should never really be done:
```python
        calcs = [key for key in ("bands", "thermal", "dos", "pdos") if locals()[key]]
```

janus_core/calculations/single_point.py
ElliottKasoar: Ah yes, I realised this when I wrote this originally, but forgot and changed it to this form today... Thanks, I'll fix it

ElliottKasoar: Do we actually want an alternative to be able to explicitly set `None` too?

I think I saw someone set a global `marker = obj()`, to be a better check against (`if x is not marker` but I'm not sure if there are neater ways.

oerc0122: Following discussion the proposal seems to be to remove parameter setting from `run`, simplify the calculation, and require setting of relevant data by the user as property setting or through `__init__`.

janus_core/calculations/single_point.py
ElliottKasoar: True, probably slightly safer, fixed

janus_core/helpers/utils.py
```diff
@@ -191,23 +163,21 @@ def spacegroup(
     )
 
 
-def none_to_dict(dictionaries: list[Optional[dict]]) -> list[dict]:
+def none_to_dict(dictionaries: Sequence[Optional[dict]]) -> Generator[dict]:
```
oerc0122, line 166: Generator strictly takes 3 args (yield, return, receives), so this should be:
```suggestion
def none_to_dict(dictionaries: Sequence[Optional[dict]]) -> Generator[dict, None, None]:
```

janus_core/helpers/utils.py
```diff
@@ -191,23 +163,21 @@ def spacegroup(
...
-        List of dictionaries that be be None.
+    dictionaries : Sequence[dict]
+        Sequence of dictionaries that be be None.
 
     Returns
     -------
-    list[dict]
-        Dictionaries set to {} if previously None.
+    Generator[dict]
+        Generator of dictionaries set to {} if previously None.
```
oerc0122, line 178:
```rst
Yields
------
dict
    Input dictionary or ``{}`` if empty or `None`.

File: tests/test_utils.py [hunk 8]
```diff
@@ -138,3 +139,24 @@ def test_output_structs(
...
+    dicts = list(none_to_dict(dicts))
+    for dictionary in dicts:
+        assert dictionary is not None
+
+    assert dicts[0] == {}
+    assert dicts[1]["a"] == 1
+    assert dicts[2] == {}
+    assert dicts[3]["b"] == 2
+    assert dicts[3]["a"] == 11
+    assert dicts[4] == {}
```
oerc0122, line 162:
```suggestion
@pytest.mark.parametrize(
    "dicts_in",
    [
        [None, {"a": 1}, {}, {"b": 2, "a": 11}, None],
        (None, {"a": 1}, {}, {"b": 2, "a": 11}, None),
    ],
)
def test_none_to_dict(dicts_in):
    """Test none_to_dict removes Nones from sequence, and preserves dictionaries."""
    dicts = list(none_to_dict(dicts))
    for dictionary in dicts:
        assert dictionary is not None

    assert dicts[0] == {}
    assert dicts[1] == dicts_in[1]
    assert dicts[2] == dicts_in[2]
    assert dicts[3] == dicts_in[3]
    assert dicts[4] == {}
```

oerc0122 (changes requested):

tests/test_utils.py
ElliottKasoar: Is this supposed to be `dicts = list(none_to_dict(dicts_in))`?

What's the benefit/aim generally?

janus_core/helpers/utils.py
ElliottKasoar: Even though it's not from a `yield` statement?

oerc0122: Hm... good point. I'm not sure... Maybe we should just change it to a yield from? I don't know.

ElliottKasoar: > Even though it's not from a `yield` statement?

numpydoc does indeed complain about no Returns section, so yep changed to `yield from`

tests/test_utils.py
oerc0122: Yes. Just to avoid having the manual checks of every element. Just simplifying the logical info.

ElliottKasoar: Fair enough, I sort of like having it closer to the way we use it, but it's the same thing really, so yeah probably neater

janus_core/calculations/geom_opt.py
```diff
@@ -69,6 +69,13 @@ class GeomOpt(FileNameMixin):  # pylint: disable=too-many-instance-attributes
         Logger if log file has been specified.
     tracker : Optional[OfflineEmissionsTracker]
         Tracker if logging is enabled.
+
+    Methods
+    -------
+    set_optimizer()
+        Set optimizer for geometry optimization.
+    run()
+        Run geometry optimization.
```
oerc0122, line 78: Does this raise duplicate definition warnings when building docs?

ElliottKasoar: Duplicate of the individual method docstrings?

They're optional, but this is just a list of the [methods for the class](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring), rather than the individual docstrings for each method.

oerc0122: Often sphinx complains if things are fully documented in docstrings and also listed in `Methods` or `Attributes` blocks.

ElliottKasoar: Ah ok, it doesn't seem to be a problem in this case. Our single_point module already has a few listed: https://stfc.github.io/janus-core/apidoc/janus_core.html#module-janus_core.calculations.single_point

alinelena (dismissed):

janus_core/calculations/geom_opt.py
oerc0122: Nope, doesn't raise an issue. All good. Weird though given I've had to "fix" so many of these in other docs. Maybe "Methods" doesn't do it in the same way "Attributes" does or maybe it depends on the Sphinx theme.

oerc0122 (approved):

PR #226, opened 2024-07-24
add sevennet support

alinelena:


pyproject.toml
```diff
@@ -46,6 +46,9 @@ codecarbon = "^2.5.0"
 optional = true
 [tool.poetry.group.extra-mlips.dependencies]
 alignn = "^2024.5.27"
+sevennet = "^0.9.1"
```
ElliottKasoar, line 49: Whether or not we manage to use extras nicely for the current default MLIPs (such that `pip install janus-core` installs MACE, M3GNet and CHGNet by default, but `pip install janus-core[mace]` would only install MACE), perhaps we should use extras for these optional MLIPs, rather than bundling them all together.

I haven't tested this, but in this case I think we might want something like:

```
alignn = { version = "^2024.5.27", optional = true }
sevennet = { version = "^0.9.1", optional = true }


[tool.poetry.extras]
alignn= ["alignn"]
sevennet = ["sevennet "]
```

We'd need to update the tutorial to note this too.

See: https://python-poetry.org/docs/pyproject/#extras (note the behaviour differs slightly to the current implementation via groups)

docs/source/developer_guide/tutorial.rst
```diff
@@ -85,7 +85,8 @@ Converting ``model_path`` into ``path`` is a minimum requirement, but we also ai
 - If ``model_path`` is ``None``, we use the ALIGNN's ``default_path``
 
 .. note::
-    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists.
+    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists. you may want to cast it back to a string
+     ```str(model_path)``` if a string is needed
```
alinelena, line 89:
```suggestion
     ``str(model_path)`` if a string is needed
```

docs/source/developer_guide/tutorial.rst
```diff
@@ -85,7 +85,8 @@ Converting ``model_path`` into ``path`` is a minimum requirement, but we also ai
 - If ``model_path`` is ``None``, we use the ALIGNN's ``default_path``
 
 .. note::
-    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists.
+    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists. you may want to cast it back to a string
```
oerc0122, line 88:
```suggestion
    ``model_path`` will already be a :class:`pathlib.Path` object, if the path exists. you may want to cast it back to a string
```
Should give you a nice link here.

docs/source/developer_guide/tutorial.rst
```diff
@@ -121,28 +127,39 @@ For ``tests.test_mlip_calculators``, ``architecture``, ``device`` and accepted f
     )
     def test_extra_mlips(architecture, device, kwargs):
 
+not all models may support empty paths, so for some you may want to remove the ``("alignn", "cpu", {})`` test.
+
+Load models - failure
+^^^^^^^^^^^^^^^^^^^^^
+
 It is also useful to test that ``model_path``, and ``model`` or and the "standard" MLIP calculator parameter (``path``) cannot be defined simultaneously::
```
oerc0122, line 135: Does this highlight properly? May be better as:
```suggestion
It is also useful to test that ``model_path``, and ``model`` or and the "standard" MLIP calculator parameter (``path``) cannot be defined simultaneously

.. code-block:: python
```

docs/source/developer_guide/tutorial.rst
```diff
@@ -85,7 +85,8 @@ Converting ``model_path`` into ``path`` is a minimum requirement, but we also ai
 - If ``model_path`` is ``None``, we use the ALIGNN's ``default_path``
 
 .. note::
-    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists.
+    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists. you may want to cast it back to a string
+     ``str(model_path)`` if a string is needed
```
ElliottKasoar, line 89:
```suggestion
    ``model_path`` will already be a ``pathlib.Path`` object, if the path exists.
    Some MLIPs do not support this, so you may be required to cast it back to a string (``str(model_path)``).
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -203,6 +206,27 @@ def choose_calculator(
...
+            raise ValueError(
+                "Please specify `model_path` there is no "
+                f"default model for {architecture}"
+            )
+        if isinstance(model_path, Path):
+            model = str(model_path)
+        elif isinstance(model_path, str):
+            model = model_path
+        else:
+            model = None
```
oerc0122, line 224: What does `model = None` do here? `raise`?

tests/test_mlip_calculators.py
```diff
@@ -103,9 +105,12 @@ def test_invalid_device(architecture):
     [
         ("alignn", "cpu", {}),
         ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024"}),
-        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024/best_model.pt"}),
+        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024" / "best_model.pt"}),
```
oerc0122, line 108: Looking at it, I have a feeling
```
MODEL_PATH / "v5.27.2024"
```
wants to become a module level constant.

docs/source/developer_guide/tutorial.rst
```diff
@@ -121,28 +127,39 @@ For ``tests.test_mlip_calculators``, ``architecture``, ``device`` and accepted f
     )
     def test_extra_mlips(architecture, device, kwargs):
 
+not all models may support empty paths, so for some you may want to remove the ``("alignn", "cpu", {})`` test.
```
ElliottKasoar, line 130:
```suggestion
.. note::
    Not all models support an empty (default) model path, so the equivalent test to``("alignn", "cpu", {})`` may need to be removed, or moved to the tests described in `Load models - failure`_.
```

pyproject.toml
```diff
@@ -32,20 +32,23 @@ chgnet = "0.3.8"
 dgl = "2.1.0"
 mace-torch = "0.3.6"
 matgl = "1.1.2"
-numpy = "^1.26.4"
+numpy = "1.26.4"
```
oerc0122, line 35: Any reason for unpinning these in particular? Particularly with numpy, recommend doing that in a separate PR.

janus_core/helpers/mlip_calculators.py
alinelena: yes inside the sevennet

pyproject.toml
oerc0122: You've just enabled 2.0 by unpinning this.

tests/test_mlip_calculators.py
```diff
@@ -123,16 +128,32 @@ def test_extra_mlips(architecture, device, kwargs):
...
+        {
+            "architecture": "sevennet",
+            "model_path": SEVENNET_PATH,
+            "model": SEVENNET_PATH,
         },
         {
-            "model_path": MODEL_PATH / "v5.27.2024/best_model.pt",
-            "path": MODEL_PATH / "v5.27.2024/best_model.pt",
+            "architecture": "sevennet",
+            "model_path": "",
```
ElliottKasoar, line 152: Maybe worth adding a test without model_path as a kwarg at all?

tests/test_mlip_calculators.py
```diff
@@ -123,16 +128,32 @@ def test_extra_mlips(architecture, device, kwargs):
...
         },
     ],
 )
-def test_extra_mlips_invalid(kwargs):
+def test_alignff_invalid(kwargs):
```
ElliottKasoar, line 156:
```suggestion
def test_extra_mlips_invalid(kwargs):
```
I think this got lost in a rebase?

pyproject.toml
alinelena: reverted

tests/test_mlip_calculators.py
alinelena: done

docs/source/developer_guide/tutorial.rst
```diff
@@ -121,28 +127,39 @@ For ``tests.test_mlip_calculators``, ``architecture``, ``device`` and accepted f
...
 It is also useful to test that ``model_path``, and ``model`` or and the "standard" MLIP calculator parameter (``path``) cannot be defined simultaneously::
 
     @pytest.mark.extra_mlips
     @pytest.mark.parametrize(
         "kwargs",
         [
             {
+                "architecture": "alignn",
                 "model_path": "tests/models/v5.27.2024/best_model.pt",
                 "model": "tests/models/v5.27.2024/best_model.pt",
```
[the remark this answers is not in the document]
alinelena: this was done already.. odd I see it only now

tests/test_mlip_calculators.py
alinelena: that will fail... which part of the code will test?

janus_core/helpers/mlip_calculators.py
```diff
@@ -203,6 +206,27 @@ def choose_calculator(
 
         calculator = AlignnAtomwiseCalculator(path=path, device=device, **kwargs)
 
+    elif architecture == "sevennet":
+        from sevenn.sevennet_calculator import SevenNetCalculator
+
+        __version__ = "0.0.0"
+
+        if model_path is None or model_path.name == "":
```
ElliottKasoar, line 214: If I set `model_path = "."` this will also throw an error, I think

```
In [6]: x
Out[6]: PosixPath('.')

In [7]: x.name
Out[7]: ''
```

But you can imagine a user might want to do this, so I think we might need a different test.

I still think testing `if model_path == ""` within `_set_model_path` makes more sense, as this differentiates between "" and ".", and we can probably assume "" isn't valid for any MLIP.

tests/test_mlip_calculators.py
ElliottKasoar: Fail as in throw the wrong error, or not throw an error?

I mean we cuold check not specifying `model_path` (or similar) also throws an error, which is different to setting it to an invalid value (`""`), which throws a potentially different error.

tests/test_mlip_calculators.py
```diff
@@ -102,10 +106,13 @@ def test_invalid_device(architecture):
...
-        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024"}),
-        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024/best_model.pt"}),
+        ("alignn", "cpu", {"model_path": ALIGNN_PATH}),
+        ("alignn", "cpu", {"model_path": ALIGNN_PATH / "best_model.pt"}),
         ("alignn", "cpu", {"model": "alignnff_wt10"}),
-        ("alignn", "cpu", {"path": MODEL_PATH / "v5.27.2024"}),
+        ("alignn", "cpu", {"path": ALIGNN_PATH} ),
+        ("sevennet", "cpu", {"model": SEVENNET_PATH}),
+        ("sevennet", "cpu", {"path": SEVENNET_PATH}),
+        ("sevennet", "cpu", {"model_path": SEVENNET_PATH}),
```
ElliottKasoar, line 115: We probably should test the string labels too (https://github.com/MDIL-SNU/SevenNet/tree/main?tab=readme-ov-file#training) - one of `"base"`, `"fine_tune"`, and `"sevennet-0"`? I think these should work as is.

tests/test_mlip_calculators.py
```diff
@@ -102,10 +106,13 @@ def test_invalid_device(architecture):
     "architecture, device, kwargs",
     [
         ("alignn", "cpu", {}),
-        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024"}),
-        ("alignn", "cpu", {"model_path": MODEL_PATH / "v5.27.2024/best_model.pt"}),
+        ("alignn", "cpu", {"model_path": ALIGNN_PATH}),
+        ("alignn", "cpu", {"model_path": ALIGNN_PATH / "best_model.pt"}),
         ("alignn", "cpu", {"model": "alignnff_wt10"}),
-        ("alignn", "cpu", {"path": MODEL_PATH / "v5.27.2024"}),
+        ("alignn", "cpu", {"path": ALIGNN_PATH} ),
```
ElliottKasoar, line 112:
```suggestion
        ("alignn", "cpu", {"path": ALIGNN_PATH}),
```
I think this is all the pre-commit is complaining about

pyproject.toml
```diff
@@ -39,13 +39,16 @@ typer-config = "^1.4.0"
 phonopy = "^2.23.1"
 seekpath = "^1.9.7"
 spglib = "^2.3.0"
-torch-dftd = "^0.4.0"
+torch-dftd = "0.4.0"
```
oerc0122, line 42: Are we also certain we want to unpin torch? Presumably it was pinned for a reason.

oerc0122: My brain is working backwards. This is now *strictly* pinned to `0.4.0`. (As was numpy being strictly pinned) for some reason I thought this was specifying minimum version now

docs/source/developer_guide/tutorial.rst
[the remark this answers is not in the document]
ElliottKasoar: This is updating the guide to match the tests, not the tests themselves

pyproject.toml
ElliottKasoar: For reference, I think this is only used by [mace](https://github.com/ACEsuit/mace/blob/main/mace/calculators/foundations_models.py). Pinning, as it now is, is probably safest.

alinelena: yes correct... more precisely mace_mp for mace naked we need a sum calculator from ase... probably similar for others...

janus_core/helpers/mlip_calculators.py
alinelena: fixed by refactoring

tests/test_mlip_calculators.py
alinelena: mute point now since labels are supported.

oerc0122 (dismissed):

tests/test_mlip_calculators.py
ElliottKasoar: Did you add a test for this?

Now that there's a default, I think we want to test that it works both with `kwargs = {"model_path": "sevennet-0"}` and `kwargs={}` (which should hopefully do the same).

tests/test_mlip_calculators.py
```diff
@@ -113,6 +113,8 @@ def test_invalid_device(architecture):
         ("sevennet", "cpu", {"model": SEVENNET_PATH}),
         ("sevennet", "cpu", {"path": SEVENNET_PATH}),
         ("sevennet", "cpu", {"model_path": SEVENNET_PATH}),
+        ("sevennet", "cpu", {}),
+        ("sevennet", "cpu", {"sevennet-0"}),
```
ElliottKasoar, line 117: Same for this, I think?

alinelena: yap, should be solved now.

tests/test_single_point.py
```diff
@@ -293,6 +293,7 @@ def test_mlips(arch, device, expected_energy):
     ("alignn", "cpu", -11.148092269897461, {}),
     ("sevennet", "cpu", -27.061979293823242, {"model_path": SEVENNET_PATH}),
     ("sevennet", "cpu", -27.061979293823242, {}),
+    ("sevennet", "cpu", -27.061979293823242, {'SevenNet-0_11July2024'}),
```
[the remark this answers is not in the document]
alinelena: done

tests/test_mlip_calculators.py
alinelena: done now

tests/test_single_point.py
```diff
@@ -286,17 +289,23 @@ def test_mlips(arch, device, expected_energy):
...
+    ("sevennet", "cpu", -27.061979293823242, {}),
+    ("sevennet", "cpu", -27.061979293823242, {"model": "SevenNet-0_11July2024"}),
+]
 
 
 @pytest.mark.extra_mlips
-@pytest.mark.parametrize("arch, device, expected_energy", test_extra_mlips_data)
-def test_extra_mlips(arch, device, expected_energy):
-    """Test single point energy using ALIGNN-FF calculator."""
+@pytest.mark.parametrize("arch, device, expectEd_energy, kwargs", test_extra_mlips_data)
```
ElliottKasoar, line 301:
```suggestion
@pytest.mark.parametrize("arch, device, expected_energy, kwargs", test_extra_mlips_data)
```

tests/test_single_point.py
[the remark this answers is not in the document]
ElliottKasoar: I think actually it might have unpacked into the right position, but I like this being more explicit anyway, thanks

PR #227, opened 2024-07-30
Update documentation

ElliottKasoar:
Resolves most of #6

- Moves examples from README into docs user guide
  - Single point example (+ Python alternative) left
- Add getting started to README/user guide

Loosely depends on #225 for minor changes to CLI, so we should check through this once that is finalised.


alinelena:
I would move in examples also as much as possible from tutorials... but for that we need to be sure we can run then (hence having our own runner for that)... there is a jupyter execute that we may want to use...

ElliottKasoar:
> I would move in examples also as much as possible from tutorials... but for that we need to be sure we can run then (hence having our own runner for that)... there is a jupyter execute that we may want to use...

I thought we wanted to keep them somewhat separate, particularly as we use a few additional structure files etc, which would make this repository more bloated?

I'm open to alternatives, but I thought we may just want to build very simple examples to add to the python interface section here (plus more detail about each class/parameters), but keep the longer form tutorials separate.

(Either way, I think it's maybe worth doing as a new PR)

docs/source/getting_started/getting_started.rst
```diff
@@ -0,0 +1,35 @@
...
+Dependencies
+------------
+
+``janus-core`` dependencies currently include:
+
+- Python >= 3.9
+- ASE >= 3.23
+- chgnet = 0.3.8
+- mace-torch = 0.3.6
+- matgl = 1.1.2
```
alinelena, line 14: can we add alignn and sevennet?

docs/source/index.rst
```diff
@@ -15,6 +15,7 @@ Welcome to janus-core's documentation!
    :caption: Contents:
 
    user_guide/index
+   getting_started/getting_started
    developer_guide/index
```
alinelena, line 19:
```suggestion
   getting_started/getting_started
   user_guide/index
   developer_guide/index
```

docs/source/user_guide/command_line.rst
```diff
@@ -0,0 +1,276 @@
...
+.. code-block:: yaml
+
+    struct: "NaCl.cif"
+    properties:
+    - "energy"
+    out: "NaCl-results.extxyz"
+    arch: mace_mp
+    model-path: medium
+    calc-kwargs:
+    dispersion: True
```
alinelena, line 220:
```suggestion
      dispersion: True
```

docs/source/user_guide/command_line.rst
```diff
@@ -0,0 +1,276 @@
...
+
+Foundational models can also be fine-tuned, by including the ``foundation_model`` option in your configuration file, and using ``--fine-tune`` option:
+
+.. code-block:: bash
+
+    janus train --mlip-config /path/to/fine/tuning/config.yml --fine-tune
+
+
+
+Calculate MACE descriptors
```
alinelena, line 255:
```suggestion
Calculate descriptors
```

docs/source/user_guide/command_line.rst
```diff
@@ -0,0 +1,276 @@
...
+--------------------------
+
```
alinelena, line 257: add a note only MACE descriptors are supported at the moment.

docs/source/user_guide/command_line.rst
```diff
@@ -0,0 +1,276 @@
+======================
+Command line interface
+======================
+
```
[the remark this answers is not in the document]
ElliottKasoar: The autocomplete is mentioned in the `--help` output I've added, so hopefully that's ok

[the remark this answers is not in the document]
alinelena: i will be explicit... rather than assuming people will read the help...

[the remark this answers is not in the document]
ElliottKasoar: Fair enough, added as a tip

docs/source/developer_guide/tutorial.rst
```diff
@@ -14,21 +14,29 @@ The following steps can then be taken, using `ALIGNN-FF <https://github.com/usni
...
+
+    [tool.poetry.extras]
+    alignnff = ["alignn"]
+    sevennet = ["sevenn", "torch_geometric"]
 
 Poetry will automatically resolve dependencies of the MLIP, if present, to ensure consistency with existing dependencies.
 
-These dependencies can then be installed by running:
+.. note::
+    In the case of ``sevennet``, it was necessary to add ``torch_geometric`` as an additional dependency, but in most cases this should not be required
```
oerc0122, line 31: It might be best to avoid using it as an example if it's a special case then.

ElliottKasoar: I mostly added this as well as `alignn` because it's a special case (although partly also since it's useful to show multiple MLIPs for the `--extras` installation).

Most of this tutorial is focused on alignn, but I thought it's probably useful to show how, if the MLIP doesn't come with everything it needs, you'd add that too.

ElliottKasoar: Just to add this isn't unique to sevennet, unfortunately. We'd need something similar for m3gnet (matgl) if it was an extra too, so I definitely think an example is useful

alinelena: yes and in theory all the potentials will be extras... and we will have [all] that pulls them all... I am in two minds about making mace as default... I am inclining towards it... but happy to hear options.

README.md
```diff
@@ -1,244 +1,143 @@
...
 [![DOI][doi-badge]][doi-link]
 
-# janus-core
-
-![logo][logo]
-
 Tools for machine learnt interatomic potentials
 
-## Features in development
+## Getting started
```
alinelena, line 16: I will add link to the getting started which is in the docs... and of course similar stuff about development, and usage.

I will also add a section something like, janus can be used command line or via python scripting or jupyter notebooks and add link to relevant parts in docs...

docs/source/developer_guide/tutorial.rst
alinelena: but this is not about documentations per se...

docs/source/user_guide/python.rst
```diff
@@ -0,0 +1,5 @@
+================
+Python interface
+================
+
+Tutorials illustrating the use of currently available calculations can be found in the `janus-tutorials <https://github.com/stfc/janus-tutorials>`_ repository.
```
alinelena, line 5: maybe will be more desciptive? list the examples?

docs/source/user_guide/command_line.rst
```diff
@@ -0,0 +1,360 @@
...
+This allows the cell vectors and angles to be optimized, as well as the atomic positions, at constant volume, and sets the ``alpha``, the initial guess for the Hessian, to 100 for the optimizer function.
+
+For all options, run ``janus geomopt --help``.
+
+
+Molecular dynamics
+------------------
+
+Run an NPT molecular dynamics simulation (using the `MACE-MP <https://github.com/ACEsuit/mace-mp>`_ "small" force-field) at 300K and 1 bar for 1000 steps (1 ps):
```
oerc0122, line 156:
```suggestion
Run an NPT molecular dynamics simulation (using the `MACE-MP <https://github.com/ACEsuit/mace-mp>`_ "small" force-field) at 300K and 1 bar for 1000 steps (1 ps, default time-step is 1 fs):
```

oerc0122 (approved):

PR #229, opened 2024-07-30
Add table writer to utils for standard format tables

oerc0122:
Table dumper for data for ASCII and CSV forms.

### DIFF
--- a/janus_core/helpers/utils.py
+++ b/janus_core/helpers/utils.py
@@ -2,8 +2,9 @@
 
 from abc import ABC
 from collections.abc import Collection, Sequence
+from io import StringIO
 from pathlib import Path
-from typing import Optional, get_args
+from typing import Any, Literal, Optional, TextIO, get_args
 
 from ase import Atoms
 from ase.io import write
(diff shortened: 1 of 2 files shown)


alinelena (dismissed):

janus_core/helpers/utils.py
```diff
@@ -337,3 +338,168 @@ def output_structs(
...
+    --------
+    >>> data = write_table(fmt="ascii", single=(1, 2, 3),
+    ...                    double=(2,4,6), double_units="THz")
+    >>> print(*data, sep="", end="")
+    # single | double [THz]
+    1 2
+    2 4
+    3 6
+    >>> data = write_table(fmt="csv", a=(3., 5.), b=(11., 12.),
+    ...                    units={'a': 'eV', 'b': 'Ha'})
```
ElliottKasoar, line 401: Do we definitely want two ways to specify the units? It's a bit confusing to me, and it's not obvious that `_units` overwrites the units dict.

I don't mind too much either way, a quick comment might be sufficient

janus_core/helpers/utils.py
```diff
@@ -337,3 +338,186 @@ def output_structs(
...
+    fmt: Literal["ascii", "csv"],
+    file: Optional[TextIO] = None,
+    units: Optional[dict[str, str]] = None,
+    formats: Optional[dict[str, str]] = None,
+    **columns,
+) -> Optional[StringIO]:
+    """
+    Dump a table in a standard format.
+
+    Table columns are passed as kwargs. With the column header being
```
ElliottKasoar, line 353:
```suggestion
    Table columns are passed as kwargs, with the column header being
```

ElliottKasoar (commented):
I think this needs rebasing, and an incredibly minor suggestion, but otherwise looks good

ElliottKasoar (approved):

PR #231, opened 2024-07-31
Ci

alinelena:


ElliottKasoar:
We will reuse similar settings (path etc.) for regression tests, but this isn't needed for now

PR #237, opened 2024-08-01
[skip ci] [do not merge]update readme with install info

alinelena:


ElliottKasoar:
I think this is mostly covered in #227, although you're right that `python3 -m pip` is sensible to add

alinelena:
> I think this is mostly covered in #227, although you're right that `python3 -m pip` is sensible to add

you are right I did not check it before... be sure you move all the changes from this in 227 and close this one...

PR #243, opened 2024-08-02
Rename geomopt flags

### LINKED ISSUE #180: Rename `vectors_only`
`vectors_only` isn't the clearest name to me as to what that means. The help explains it, but I think `cell_only` would possibly be better since the atomic positions are also vectors.

_Originally posted by @oerc0122 in https://github.com/stfc/janus-core/pull/179#discussion_r1624393815_

ElliottKasoar:
Resolves #180

- Renames `--vectors-only` to `--opt-cell-lengths` to reduce confusion that this is in addition to optimizing cell positions.
- Renames `--fully-opt` to `--opt-cell-fully` for consistency (if we ever add something like `--opt-cell-angles`, this would be equivalent to both being true, which feels reasonably consistent)

Still open to further suggestions

### DIFF
--- a/janus_core/cli/geomopt.py
+++ b/janus_core/cli/geomopt.py
@@ -180,7 +180,7 @@ def geomopt(
         Default is False.
     filter_func : Optional[str]
         Name of filter function from ase.filters or ase.constraints, to apply
-        constraints to atoms. If using --vectors only or --opt-cell-fully, defaults to
+        constraints to atoms. If using --opt-cell-lengths or --opt-cell-fully, defaults to
         `FrechetCellFilter` if available, otherwise `ExpCellFilter`.
     pressure : float
         Scalar pressure when optimizing cell geometry, in GPa. Passed to the filter
--- a/janus_core/cli/geomopt.py
+++ b/janus_core/cli/geomopt.py
@@ -36,7 +36,7 @@ app = Typer()
 def _set_minimize_kwargs(
     minimize_kwargs: dict[str, Any],
     traj: Optional[str],
-    vectors_only: bool,
+    opt_cell_lengths: bool,
     pressure: float,
 ) -> None:
     """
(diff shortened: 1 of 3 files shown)


alinelena (approved):

janus_core/cli/geomopt.py
```diff
@@ -171,19 +172,20 @@ def geomopt(
...
         Whether to optimize cell vectors, as well as atomic positions, by setting
         `hydrostatic_strain` in the filter function. Default is False.
-    fully_opt : bool
+    opt_cell_fully : bool
         Whether to fully optimize the cell vectors, angles, and atomic positions.
         Default is False.
     filter_func : Optional[str]
         Name of filter function from ase.filters or ase.constraints, to apply
-        constraints to atoms. If using --vectors only or --fully-opt, defaults to
+        constraints to atoms. If using --vectors only or --opt-cell-fully, defaults to
```
oerc0122, line 183:
```suggestion
        constraints to atoms. If using --opt-cell-lengths or --opt-cell-fully, defaults to
```

oerc0122 (commented):
With that fix, I'm happy

janus_core/cli/geomopt.py
ElliottKasoar: Well spotted, thanks

oerc0122 (dismissed):

alinelena (dismissed):

janus_core/cli/geomopt.py
```diff
@@ -171,19 +172,20 @@ def geomopt(
...
         `hydrostatic_strain` in the filter function. Default is False.
-    fully_opt : bool
+    opt_cell_fully : bool
         Whether to fully optimize the cell vectors, angles, and atomic positions.
         Default is False.
     filter_func : Optional[str]
         Name of filter function from ase.filters or ase.constraints, to apply
-        constraints to atoms. If using --vectors only or --fully-opt, defaults to
+        constraints to atoms. If using --opt-cell-lengths or --opt-cell-fully, defaults to
         `FrechetCellFilter` if available, otherwise `ExpCellFilter`.
```
ElliottKasoar, line 184:
```suggestion
        constraints to atoms. If using --opt-cell-lengths or --opt-cell-fully, defaults
        to `FrechetCellFilter` if available, otherwise `ExpCellFilter`.
```

alinelena (approved):

oerc0122 (approved):

PR #244, opened 2024-08-02
Delete duplicate struct from CLI summary

### LINKED ISSUE #242: Inputs duplicated in summary
Some calculations, such as `geomopt`, [save the structure in the `inputs`](https://github.com/stfc/janus-core/blob/main/janus_core/cli/geomopt.py#L264), which can output weird binaries as it's writing out the Atoms object.

As we already handle the structure separately, we should just be able to delete the object from the inputs dict.

ElliottKasoar:
Resolves #242

Removes "struct" from the `inputs` dict when saving the CLI summary, as this is exactly what `save_struct_calc` is designed to do anyway, and if the `struct` is actually a trajectory, we end up with a duplicate, with binary info printed.

### DIFF
--- a/janus_core/cli/utils.py
+++ b/janus_core/cli/utils.py
@@ -141,6 +141,9 @@ def save_struct_calc(
     calc_kwargs : dict[str, Any]]
         Keyword arguments to pass to the calculator.
     """
+    # Remove duplicate struct if already in inputs:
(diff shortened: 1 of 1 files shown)


oerc0122 (approved):
Looks to do what it says on the tin

alinelena (approved):

PR #256, opened 2024-08-08
Update MLIP installations

### LINKED ISSUE #255: Update MLIP installations
> yes and in theory all the potentials will be extras... and we will have [all] that pulls them all... I am in two minds about making mace as default... I am inclining towards it... but happy to hear options.
>
> _Originally posted by @alinelena in https://github.com/stfc/janus-core/pull/227#discussion_r1709307046_

I'm also open to alternatives implementations, but I think we're generally agreed that we want an option to install all MLIPs.

If everyone is happy to return to MACE as the single default, that's ok with me, and then we can move everything else into individual, all, and potentially any other subsets that might be useful?

ElliottKasoar:
Resolves #255

Moves chgnet and m3gnet to optional dependencies again, but still tests against these in the CI/with tox.

I need to double check all the installation mechanisms, but this should mean pip/poetry install janus-core should only install MACE, and with extras we can install each MLIP individually, or install all together.

To do:

- [x] Update docs to reflect these changes (probably after #227 is merged)
- [x] Test against all MLIPs

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ or to install all supported MLIPs:
 python3 -m pip install janus-core[all]
 ```
 
+Individual `extras` are listed in [Getting Started](https://stfc.github.io/janus-core/getting_started/getting_started.html#installation), as well as in [pyproject.toml](pyproject.toml) under `[tool.poetry.extras]`.
+
 
 ### Further help
 
--- a/README.md
+++ b/README.md
@@ -20,9 +20,9 @@ Tools for machine learnt interatomic potentials
 
 - Python >= 3.9
 - ASE >= 3.23
-- chgnet = 0.3.8
 - mace-torch = 0.3.6
-- matgl = 1.1.2
+- chgnet = 0.3.8 (optional)
+- matgl = 1.1.3 (optional)
 - sevenn = 0.9.3 (optional)
 - alignn = 2024.5.27 (optional)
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -26,28 +26,33 @@ documentation = "https://stfc.github.io/janus-core/"
...
+chgnet = {version = "0.3.8", optional = true}
+dgl = { version = "2.1.0", optional = true } # Pin due to matgl installation issues
+matgl = { version = "1.1.3", optional = true}
 sevenn = { version = "0.9.3", optional = true }
+torchdata = {version = "0.7.1", optional = true} # Pin due to dgl issue
 torch_geometric = { version = "^2.5.3", optional = true }
 
 [tool.poetry.extras]
+all = ["alignn", "chgnet", "matgl", "dgl", "torchdata", "sevenn", "torch_geometric"]
 alignnff = ["alignn"]
```
ElliottKasoar, line 53: Do we want to change this to just "alignn", or change the calculator architecture name to match this, so we're consistent , @alinelena?

alinelena: yes fine to change it to alignn... why do we have torchdata, geometris and dgl as options... I would expect them to be deps of specific targets, sevenn and matgl/chgnet?

README.md
```diff
@@ -46,6 +46,20 @@ To get all the latest changes, `janus-core` can also be installed from GitHub:
 python3 -m pip install git+https://github.com/stfc/janus-core.git
 ```
 
+By default, MACE is the only MLIP installed.
```
alinelena, line 49: I will add a line on how the user can get the full list of available mlips.

pyproject.toml
ElliottKasoar: Cool ok

Those other dependencies aren't actual options you'd select, the extra options are `all`, `alignn`, `chgnet`, `m3gnet` and `sevennet`, the others are just dependencies within those that we have to specify to fix installation issues (either because they're missing, or we need to pin them)

all = ["alignn", "chgnet", "matgl", "dgl", "torchdata", "sevenn", "torch_geometric"]
alignn = ["alignn"]
chgnet = ["chgnet"]
m3gnet = ["matgl", "dgl", "torchdata"]
sevennet = ["sevenn", "torch_geometric"]

README.md
ElliottKasoar: I think I might have done this slightly differently to what you meant earlier, but hopefully this is ok?

alinelena (approved):

PR #265, opened 2024-08-12
Add carbon emissions for training

### LINKED ISSUE #264: Add carbon tracking for training
Currently carbon tracking is only done for calculations, but it should be relatively straight forward to also add this to training.

ElliottKasoar:
Resolves #264

- Also adds CLI log/summary in the process
  - This seemed to break some tests, as any tests that followed the training (e.g. test_utils, or any Python tests if run in a different order) threw `FileNotFoundError`s, looking for `logs/[model].log`. There may be better solutions but removing these from the `FileHandler`s from the root logger seems to fix this

Depends on #258 (will need rebasing)

### DIFF
--- a/janus_core/cli/train.py
+++ b/janus_core/cli/train.py
@@ -6,6 +6,8 @@ from typing import Annotated
 from typer import Option, Typer
 import yaml
 
+from janus_core.cli.types import LogPath, Summary
+from janus_core.cli.utils import carbon_summary, end_summary, start_summary
 from janus_core.helpers.train import train as run_train
 
 app = Typer()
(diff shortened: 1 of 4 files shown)


janus_core/cli/train.py
```diff
@@ -52,4 +61,15 @@ def train(
     elif "foundation_model" in config:
         raise ValueError("Please include the `--fine-tune` option for fine-tuning")
 
-    run_train(mlip_config)
+    inputs = {"mlip_config": str(mlip_config), "fine_tune": fine_tune}
+
+    # Save summary information before training begins
+    start_summary(command="train", summary=summary, inputs=inputs)
```
oerc0122, line 67: Would this be better implemented as a context? Not for this PR though

oerc0122 (approved):

janus_core/cli/train.py
ElliottKasoar: Potentially. There is a [context manager](https://mlco2.github.io/codecarbon/usage.html#context-manager), and function decorators too, which would be simpler, but I didn't think they integrated well with saving to the log, or having individual tasks when we have calculations that do multiple things, when I originally implemented this. Worth revisiting at some point

PR #271, opened 2024-08-13
Remove ANSI sequences from stdout in tests

### LINKED ISSUE #270: ANSI sequences in CLI help
codecarbon now includes rich, which typer uses by default to format outputs, including [ANSI escape sequences for colour](https://typer.tiangolo.com/tutorial/printing/#color).

This is causing tests to fail, as these sequences are in between plain text that we check for.

As suggested by @oerc0122, we can remove these sequences for the tests. See: https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python/14693789#14693789

ElliottKasoar:
Resolves #270

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -72,6 +72,9 @@ def decode_ansi(output: str) -> str:
     """
     Decode output, which may include rich formatting with ANSI sequences to be removed.
 
+    Based on:
+    https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python/14693789#14693789
+
     Parameters
     ----------
     output : str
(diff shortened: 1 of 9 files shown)


tests/utils.py
```diff
@@ -65,3 +66,35 @@ def assert_log_contains(
 
     assert all(inc in messages for inc in includes)
     assert all(exc not in messages for exc in excludes)
+
+
+def decode_ansi(output: str) -> str:
```
alinelena, line 71: add the stackoverflow link into the comment?

oerc0122, line 71: I wouldn't say it's "decode"ing the ansi, I'd call it "strip_ansi_codes" or similar

ElliottKasoar: I meant it as decoding strings with ansi in them, but yeah that's probably clearer

Issue #279, opened 2024-08-14
MD filenames missing variables


ElliottKasoar:
Changes introduced recently (#225, I think) mean that MD output files are missing the ensemble, temperature, etc. For example, `NaCl-final.extxyz`, rather than `NaCl-npt-T300.0-p0.0-final.extxyz`

ElliottKasoar:
I think this is only in the CLI, and comes from setting it to match the singlepoint "prefix": https://github.com/stfc/janus-core/blob/main/janus_core/cli/md.py#L337.

(This made sense for other calculations, and meant it's straight forward to use the filename, rather than having to use the chemical formula, as we do otherwise)

PR #280, opened 2024-08-14
Set consistent log and summary file names

### LINKED ISSUE #132: md summary and log shall follow same naming scheme as the rest of the files

ElliottKasoar:
Resolves #132

As mentioned in the issue, sharing the summary is relatively straightforward, but extra changes were needed to use the `FileNameMixin` to set the log filename to be consistent with other output files.

Since the default filename moves into the Python, I added an additional flag that controls the logging, so a filename isn't required. 

Loosely depends on fixing #279, for correct documentation, and tests will then require updating following a fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,9 +156,11 @@ Options:
   --write-kwargs DICT  Keyword arguments to pass to ase.io.write when saving
                        results. Must be passed as a dictionary wrapped in
                        quotes, e.g. "{'key' : value}".  [default: "{}"]
-  --log PATH           Path to save logs to.  [default: singlepoint.log]
-  --summary PATH       Path to save summary of inputs and start/end time.
-                       [default: singlepoint_summary.yml]
+  --log PATH           Path to save logs to. Default is inferred from the name
+                       of the structure file.
+  --summary PATH       Path to save summary of inputs, start/end time, and
+                       carbon emissions. Default is inferred from the name of
+                       the structure file.
   --config TEXT        Configuration file.
   --help               Show this message and exit.
 ```
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -75,9 +75,11 @@ prints the following:
       --write-kwargs DICT  Keyword arguments to pass to ase.io.write when saving
                            results. Must be passed as a dictionary wrapped in
                            quotes, e.g. "{'key' : value}".  [default: "{}"]
-      --log PATH           Path to save logs to.  [default: singlepoint.log]
-      --summary PATH       Path to save summary of inputs and start/end time.
-                           [default: singlepoint_summary.yml]
+      --log PATH           Path to save logs to. Default is inferred from the name
+                           of the structure file.
+      --summary PATH       Path to save summary of inputs, start/end time, and
+                           carbon emissions. Default is inferred from the name of
+                           the structure file.
       --config TEXT        Configuration file.
       --help               Show this message and exit.
 
@@ -92,7 +94,7 @@ Perform a single point calcuation (using the `MACE-MP <https://github.com/ACEsui
     janus singlepoint --struct tests/data/NaCl.cif --arch mace_mp --model-path small
 
 
-This will calculate the energy, stress and forces and save this in ``NaCl-results.extxyz``, in addition to generating a log file, ``singlepoint.log``, and summary of inputs, ``singlepoint_summary.yml``.
+This will calculate the energy, stress and forces and save this in ``NaCl-results.extxyz``, in addition to generating a log file, ``NaCl-log.yml``, and summary of inputs, ``NaCl-summary.yml``.
 
 Additional options may be specified. For example:
 
@@ -127,7 +129,7 @@ Perform geometry optimization (using the `MACE-MP <https://github.com/ACEsuit/ma
     janus geomopt --struct tests/data/H2O.cif --arch mace_mp --model-path small
 
 
-This will optimize the atomic positions and save the resulting structure in ``H2O-opt.extxyz``, in addition to generating a log file, ``geomopt.log``, and summary of inputs, ``geomopt_summary.yml``.
+This will optimize the atomic positions and save the resulting structure in ``H2O-opt.extxyz``, in addition to generating a log file, ``H20-log.yml``, and summary of inputs, ``H20-summary.yml``.
 
 Additional options may be specified. This shares most options with ``singlepoint``, as well as a few additional options, such as:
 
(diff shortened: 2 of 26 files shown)


README.md
```diff
@@ -156,9 +156,11 @@ Options:
...
                        results. Must be passed as a dictionary wrapped in
                        quotes, e.g. "{'key' : value}".  [default: "{}"]
-  --log PATH           Path to save logs to.  [default: singlepoint.log]
-  --summary PATH       Path to save summary of inputs and start/end time.
-                       [default: singlepoint_summary.yml]
+  --log PATH           Path to save logs to. Default is inferred from the name
+                       of the structure file.
+  --summary PATH       Path to save summary of inputs, start/end time, and
+                       carbon emissions. Default is inferred from the name of
+                       the structure file.
```
harveydevereux, line 163: Don't these now also infer from the ensemble as well when md is done, ```NaCl-npt-T300.0-p1.0-log.yml```, and not in e.g. pure geomopt, ```H20-log.yml```. Do we need to comment on that here?

Obviously would effect a lot of docstrings in the other diff's so I won't comment individually

janus_core/calculations/descriptors.py
```diff
@@ -104,18 +110,23 @@ def __init__(
...
+        # Configure logging
+        if attach_logger:
+            self.log_kwargs.setdefault(
+                "filename",
+                self._build_filename("log.yml").absolute(),
+            )
+        self.log_kwargs.setdefault("name", __name__)
+        self.logger = config_logger(**self.log_kwargs)
+        self.tracker = config_tracker(self.logger, **tracker_kwargs)
+
```
harveydevereux, line 129: Is there some way to avoid repeating this pattern in the various classes that use it? I suppose it'll end up similar to the ```FileNameMixin``` and there will have to be some repetition.

tests/test_md_cli.py
```diff
@@ -69,54 +69,72 @@ def test_md(ensemble, tmp_path):
...
-        assert lines[0].split(" | ")[16] == "Target T [K]\n"
-        init_temp = float(lines[1].split()[16])
-        final_temp = float(lines[-1].split()[16])
-        assert init_temp == 300.0
-        assert final_temp == pytest.approx(final_temp)
+    final_path = Path("./NaCl-final.extxyz").absolute()
+    stats_path = Path("./NaCl-stats.dat").absolute()
+    traj_path = Path("./NaCl-traj.extxyz").absolute()
+    log_path = Path("./NaCl-log.yml").absolute()
+    summary_path = Path("./NaCl-summary.yml").absolute()
```
harveydevereux, line 78: Are we also testing the cases when the ensemble parameters and temperatures get added in to the log and summary?

Maybe also appropriate in ```tests/test_md.py```.

janus_core/calculations/descriptors.py
oerc0122: `FileNameMixin` could probably absorb this as a method taking `attach_logger` as an argument. Or some other mixin `LoggerMixin`

janus_core/calculations/md.py
```diff
@@ -267,13 +270,15 @@ def __init__(
             Keyword arguments to control post-processing operations.
         correlation_kwargs : Optional[list[CorrelationKwargs]]
             Keyword arguments to control on-the-fly correlations.
+        seed : Optional[int]
```
oerc0122, line 273: Why has seed been added here?

tests/test_descriptors_cli.py
```diff
@@ -22,57 +22,65 @@ def test_help():
...
-        [
-            "descriptors",
-            "--struct",
-            DATA_PATH / "NaCl.cif",
-            "--out",
-            out_path,
-            "--log",
+    out_path = Path("./NaCl-descriptors.extxyz").absolute()
+    log_path = Path("./NaCl-log.yml").absolute()
+    summary_path = Path("./NaCl-summary.yml").absolute()
```
oerc0122, line 29: Why are these no longer `tmp_path`'d?

README.md
ElliottKasoar: Technically not until #279 is fixed, but yes it should do, good point. I think the docstrings for the MD CLI are the only one that needs changing, or am I missing something?

tests/test_md_cli.py
ElliottKasoar: These will change once #279 is fixed

janus_core/calculations/md.py
ElliottKasoar: I just moved it for consistency

tests/test_descriptors_cli.py
ElliottKasoar: It's testing the minimum inputs for `descriptors`, so all outputs are their default locations, as we do with the other calculations.

For this PR, the log and summary paths are the main point, as we can't save them to `tmp_path` without specifying their full paths, which means we don't test that they use the same prefix as other outputs.

Having done that, I thought doing the same for `out_path` made sense, for similar reasons.

ElliottKasoar: Similarly, I think not having any tests without `tmp_path` is the reason we didn't catch #279

README.md
harveydevereux: Ah ok yes I was slightly confused by what's coming and when. 

And I think you are right it should just be for the md cli for this comment alongside the readme, I think as that is the only time we get the ```npt-TXXX-pXXX``` etc

ElliottKasoar:
I suggest #283 is reviewed before this, as it also address the logging suggestion, and this will need various conflicts resolving.

docs/source/user_guide/command_line.rst
```diff
@@ -92,7 +94,7 @@ Perform a single point calcuation (using the `MACE-MP <https://github.com/ACEsui
     janus singlepoint --struct tests/data/NaCl.cif --arch mace_mp --model-path small
 
 
-This will calculate the energy, stress and forces and save this in ``NaCl-results.extxyz``, in addition to generating a log file, ``singlepoint.log``, and summary of inputs, ``singlepoint_summary.yml``.
+This will calculate the energy, stress and forces and save this in ``NaCl-results.extxyz``, in addition to generating a log file, ``NaCl-log.yml``, and summary of inputs, ``NaCl-summary.yml``.
```
alinelena, line 97: I will keep the opration... NaCl-singlepoint-log.yml and so on

docs/source/user_guide/command_line.rst
```diff
@@ -166,8 +168,8 @@ This will generate several output files:
 - The structure trajectory every 100 steps, written to ``NaCl-npt-T300.0-p1.0-traj.extxyz``
 - The structure to be able to restart the dynamics every 1000 steps, written to ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``
 - The final structure written to ``NaCl-npt-T300.0-p1.0-final.extxyz``
-- A log of the processes carried out, written to ``md.log``
-- A summary of the inputs and start/end time, written to ``md_summary.yml``.
+- A log of the processes carried out, written to ``NaCl-npt-T300.0-p1.0-log.yml``
```
alinelena, line 171: here I will not add md, since the ensemble makes it pointless

ElliottKasoar: Unless you actively object to it, I've kept `md` in because for the log, we determine the name in `BaseCalculation`, so having different logic for md makes things more complicated than I think they need to be.

oerc0122 (approved):

alinelena:
if keeps logic simple I am fine with keeping md

PR #281, opened 2024-08-15
Fix MD CLI file prefix

### LINKED ISSUE #279: MD filenames missing variables

ElliottKasoar:
Resolves #279

This essentially reverts the behaviour to pre-#225 for MD, so rather than using the name of the input file, the chemical formula is used, which means the ensemble and variables are not ignored when constructing file names.

In the longer term, I would quite like to return consistency to file names (with this change, MD becomes the only CLI calculation that uses the chemical formula as a default).

One option for this is converting `SinglePoint`'s read and attach calculator functionality into general utility functions, and allowing any calculation to take a file input, rather than only ASE Atoms.

This would mean the other CLI commands could bypass `SinglePoint`, and would have complete access to filenames, if given, to be used in their file prefix default.

### DIFF
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -162,10 +162,10 @@ Run an NPT molecular dynamics simulation (using the `MACE-MP <https://github.com
 
 This will generate several output files:
 
-- Thermodynamical statistics every 100 steps, written to ``NaCl-npt-T300.0-p1.0-stats.dat``
-- The structure trajectory every 100 steps, written to ``NaCl-npt-T300.0-p1.0-traj.extxyz``
-- The structure to be able to restart the dynamics every 1000 steps, written to ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``
-- The final structure written to ``NaCl-npt-T300.0-p1.0-final.extxyz``
+- Thermodynamical statistics every 100 steps, written to ``Cl4Na4-npt-T300.0-p1.0-stats.dat``
+- The structure trajectory every 100 steps, written to ``Cl4Na4-npt-T300.0-p1.0-traj.extxyz``
+- The structure to be able to restart the dynamics every 1000 steps, written to ``Cl4Na4-npt-T300.0-p1.0-res-1000.extxyz``
+- The final structure written to ``Cl4Na4-npt-T300.0-p1.0-final.extxyz``
 - A log of the processes carried out, written to ``md.log``
 - A summary of the inputs and start/end time, written to ``md_summary.yml``.
 
@@ -202,9 +202,9 @@ Run an NVT heating simultation from 20K to 300K in steps of 20K, with 10fs at ea
 
 The produced final, statistics, and trajectory files will indicate the heating range:
 
-- ``NaCl-nvt-T20.0-T300.0-final.extxyz``
-- ``NaCl-nvt-T20.0-T300.0-stats.dat``
-- ``NaCl-nvt-T20.0-T300.0-traj.extxyz``
+- ``Cl4Na4-nvt-T20.0-T300.0-final.extxyz``
+- ``Cl4Na4-nvt-T20.0-T300.0-stats.dat``
+- ``Cl4Na4-nvt-T20.0-T300.0-traj.extxyz``
 
 The final structure file will include the final structure at each temperature point (20K, 40K, ..., 300K).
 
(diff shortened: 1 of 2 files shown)


docs/source/user_guide/command_line.rst
```diff
@@ -162,10 +162,10 @@ Run an NPT molecular dynamics simulation (using the `MACE-MP <https://github.com
...
 This will generate several output files:
 
-- Thermodynamical statistics every 100 steps, written to ``NaCl-npt-T300.0-p1.0-stats.dat``
-- The structure trajectory every 100 steps, written to ``NaCl-npt-T300.0-p1.0-traj.extxyz``
-- The structure to be able to restart the dynamics every 1000 steps, written to ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``
-- The final structure written to ``NaCl-npt-T300.0-p1.0-final.extxyz``
+- Thermodynamical statistics every 100 steps, written to ``Cl4Na4-npt-T300.0-p1.0-stats.dat``
+- The structure trajectory every 100 steps, written to ``Cl4Na4-npt-T300.0-p1.0-traj.extxyz``
+- The structure to be able to restart the dynamics every 1000 steps, written to ``Cl4Na4-npt-T300.0-p1.0-res-1000.extxyz``
+- The final structure written to ``Cl4Na4-npt-T300.0-p1.0-final.extxyz``
```
alinelena, line 168: I am confused... the original behaviour was not to use chemical formula but the stem of the input structure file. this is what we shall have... chemical formula as printed by ase has little meaning.

ElliottKasoar: > Either way, I think we may need changes to the Python MD class, as we need a way for the Mixin/filename builder to know the stem, and this isn't something we can easily set after instantiation either.

Actually, we probably can bypass this, but it would be quite hacky...

I was thinking that, given only an instantiated MD object, you can't tell whether each file has been set via the builder or by the user.

However, within the CLI module, we do know if each file is specified. so we don't need to work it out.

So, assuming the user doesn't set `file_prefix`, for each filename not given by the user, we could replace the formula with the file stem.

alinelena: why can't we just use the stem as file_prefix if the user pases none and then let the individual calculations add their suffix and extension?

ElliottKasoar: That's exactly what the current code in main does, so the cause of the bug, because that's not how file prefix works. 

It did briefly originally act in that way, but the idea of the prefix was to replace anything shared between different outputs, so it overwrites the rest of the suffix parameters, leaving only unique endings e.g. `-stats`, `-traj` etc., which is why we currently end up with `NaCl-traj` etc.

alinelena: then I am confused... why a prefix will overwrite parts of the suffix? that tells me the prefix is not working as a prefix. you may want to extend the filenames then to have prefix+random+suffix+extension...

ElliottKasoar: At the time it was partly that `struct_name` did exactly what you're suggesting `file_prefix` should do, but also because I consider all of the labelling of the simulation to be a prefix, it just happens to contain more information for MD than other calculations.

I wanted a way to say "I know what these files will be, I don't need/want hard-to-read file names with all the extra stuff", and produce outputs that are consistent with all the other calculations - `file_prefix-output_type.extension`.

I appreciate it can be useful to have the information in the file name too, but it is all saved in the summary, log and/or Atoms.info already, so I don't think we should always have to include it.

alinelena: is not about producing hard to read files... The ensemble-T pattern, or step-res, are part of the same logic and they need to be supported. 
I will give you an example of a typical workflow
nacl.
0 - geometry minimisation, 1 - heat to 200K in npt, get final structure, 2 - run npt for 1000 steps, 3 - compute average cell, 4 - run nvt in this scenario a default run will make this impossible... unless user sets up prefix files to do what we already do.

ElliottKasoar: The restart files all include the step still, regardless of the prefix. They end in `-res-step.extxyz`. (Interestingly, they do actually use the stem currently, so I need to check how, but pretty sure it's because they're built on the fly, so have access to the updated prefix).

I'm not sure I understand what makes that workflow impossible?

Assuming we set `file_prefix = NaCl`, so we don't worry about NaCl vs Cl4Na4 for now:

0. `janus geomopt --struct NaCl.cif` produces `NaCl-opt.extxyz` (or we do it in MD, doesn't matter too much)
1. `janus md --ensemble nvt --struct NaCl-opt.extxyz --temp....` produces `NaCl-final.extxyz`, `NaCl-stats.dat`, and ``NaCl-traj.extxyz`
2. `janus md --ensemble npt --restart --struct NaCl-final.extxyz --temp....` produces a new `NaCl-final.extxyz`, but the trajectory and stats files should just be appended
3. Use the results from (2)
4. `janus md --ensemble nvt --restart --struct NaCl-final.extxyz` again, produces a new `NaCl-final.extxyz`, but the trajectory and stats files should just be appended

I'm probably missing something but what doesn't work?

We may lose the intermediate `-final` files, but they may actually go into the trajectory anyway.

Or do you mean because we lose the information about the simulation ensemble etc? This only happens if the user explicitly sets `--file-prefix NaCl` for all of the above steps.

If they don't, they'll all include ensemble, T, etc., so at the point where a user is deciding what `file-prefix` is, surely they can add in any relevant information for each step?

alinelena: if you use file-prefix you are on your own so any issues are of your own making... so that is fine.

if one does not use file prefix. we need to be clear what happens. 

for an nvt at T300K shall produce NaCl-nvt-T300K-stats.dat -traj.extxyz, -log. -summary.yml

alinelena: my objection here to use of a random stoichiometric dformula for prefix.

ElliottKasoar: So you're ok with a user defined file prefix overriding the rest? That's the main thing I'm not keen on changing...

In terms of the formula, again I agree that in the case of the CLI, the stem is nicer, but in cases where we don't have that, like on the Python side (where anything other than `SinglePoint` only gets the `Atoms` object), is there something better to use?

On the CLI side, I think the two solutions I proposed still hold:

1. Bring back struct_name for MD, so we have a way of passing the stem to replace the formula
2. The change I mention in the PR description, where we allow files to be passed directly to calculations, which again would give knowledge of the stem within the class

In either case we use the stem, either as `struct_name`, which only replaced the part before `-nvt...`, or by replacing the formula with the stem when we have it, as we do for `SinglePoint` at the moment, but applied to all calculations

ElliottKasoar:
Closing in favour of #283, which implements the change I proposed, following discussions with @alinelena.

PR #283, opened 2024-08-19
Add file inputs to all calculations

### LINKED ISSUE #279: MD filenames missing variables

ElliottKasoar:
Resolves #279

- Moves parsing of `struct`/`struct_path` and attaching calculators to new base class, which all calculations inherit from
  - Provides access to file stems within (hence resolving #279)
  - Also means we don't need to do everything via SinglePoint

To do:

- [x] Add tests 
  - Mostly covered via CLIs, which all use `struct_path`, but we can update MD, for example, to ensure #279 is actually resolved
  - `test_filenamemixin` needs updating
- [x] Check docs are up to date
  - API list needs updating

### DIFF
--- a/janus_core/calculations/descriptors.py
+++ b/janus_core/calculations/descriptors.py
@@ -16,7 +16,7 @@ from janus_core.helpers.janus_types import (
     MaybeSequence,
     PathLike,
 )
-from janus_core.helpers.utils import FileNameMixin, none_to_dict
+from janus_core.helpers.utils import none_to_dict
 
 
 class Descriptors(BaseCalculation):
@@ -159,7 +159,6 @@ class Descriptors(BaseCalculation):
             raise ValueError("Please attach a calculator to `struct`.")
 
         # Set output file
-        FileNameMixin.__init__(self, self.struct, self.struct_path, None)
         self.write_kwargs.setdefault(
             "filename",
             self._build_filename("descriptors.extxyz").absolute(),
--- a/janus_core/calculations/descriptors.py
+++ b/janus_core/calculations/descriptors.py
@@ -7,19 +7,47 @@ from ase import Atoms
 from ase.io import write
 import numpy as np
 
-from janus_core.helpers.janus_types import ASEWriteArgs, MaybeSequence
-from janus_core.helpers.log import config_logger, config_tracker
+from janus_core.calculations.base import BaseCalculation
+from janus_core.helpers.janus_types import (
+    Architectures,
+    ASEReadArgs,
+    ASEWriteArgs,
+    Devices,
+    MaybeSequence,
+    PathLike,
+)
 from janus_core.helpers.utils import FileNameMixin, none_to_dict
 
 
-class Descriptors(FileNameMixin):
+class Descriptors(BaseCalculation):
     """
     Prepare and calculate MLIP descriptors for structures.
 
     Parameters
     ----------
-    struct : MaybeSequence[Atoms]
-        Structure(s) to calculate descriptors for.
+    struct : Optional[MaybeSequence[Atoms]]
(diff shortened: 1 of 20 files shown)


janus_core/calculations/descriptors.py
```diff
@@ -78,39 +123,59 @@ def __init__(
...
+            write_kwargs,
+        ) = none_to_dict(
+            (
+                read_kwargs,
+                calc_kwargs,
+                log_kwargs,
+                tracker_kwargs,
+                write_kwargs,
+            )
         )
```
oerc0122, line 141: Duplicate?

janus_core/calculations/geom_opt.py
```diff
@@ -187,13 +215,32 @@ def __init__(
...
+            log_kwargs=log_kwargs,
+            tracker_kwargs=tracker_kwargs,
+        )
+
+        if not self.struct.calc:
+            raise ValueError("Please attach a calculator to `struct`.")
 
         # Set output file
-        FileNameMixin.__init__(self, self.struct, None)
+        FileNameMixin.__init__(self, self.struct, self.struct_path, None)
```
oerc0122, line 243: Will this move into the base calculation?

janus_core/calculations/descriptors.py
ElliottKasoar: read_kwargs stays as a duplicate now, so I can set the default index for each calculation, but otherwise these should now only by the kwargs not set within the base calc

janus_core/calculations/geom_opt.py
ElliottKasoar: There may be a nicer/more general way to pass additional, but implemented a potential fix, good idea

ElliottKasoar:
Also note, this changes the default read index from 0 to -1 for calculations that can't read everything, which is consistent with ASE's default if you don't specify an index.

janus_core/helpers/utils.py
```diff
@@ -257,6 +276,157 @@ def results_to_info(
...
+    # If set_calc not specified, set to True if reading file or calculators not attached
+    if set_calc is None:
+        if struct_path:
+            set_calc = True
+
+        if isinstance(struct, Atoms):
+            set_calc = True if struct.calc is None else False
+
+        if isinstance(struct, Sequence):
+            set_calc = True if any(image.calc is None for image in struct) else False
```
oerc0122, line 412:
```suggestion
        if isinstance(struct, Atoms):
            set_calc = struct.calc is None

        if isinstance(struct, Sequence):
            set_calc = any(image.calc is None for image in struct)
```

tests/test_md_cli.py
```diff
@@ -34,39 +34,80 @@ def test_md_help():
...
+    elif ensemble == "nve":
+        file_prefix = "NaCl-nve-T300.0-"
+    elif ensemble == "npt":
+        file_prefix = "NaCl-npt-T300.0-p0.0-"
+    elif ensemble == "nvt-nh":
+        file_prefix = "NaCl-nvt-nh-T300.0-"
+    elif ensemble == "nph":
+        file_prefix = "NaCl-nph-T300.0-p0.0-"
+    else:
+        raise NotImplementedError("Ensemble not currently supported")
```
oerc0122, line 51: This kind of chained `if` may be better as a `dict`?

oerc0122 (commented):
Just a couple of stylistic suggestions.

oerc0122 (approved):

PR #285, opened 2024-08-21
Adds Velocity, ShearStress, StressDiagonal, observables

harveydevereux:
This adds a Velocity observable capable of averaging over supplied components ```x, y, z``` and atom lists. The behaviour is abstracted so also for stress we can obtain shear stress correlations $\langle \sigma_{xy}\sigma_{xy}\rangle+\langle \sigma_{yz}\sigma_{yz}\rangle+\langle \sigma_{zx}\sigma_{zx}\rangle$ with the built in ```ShearStress === Stress(['xy', 'yz', 'zx'])```

- ci tests now test against post-process vaf, dependent on changes to it #284 .

- I used the ShearStress observable to get experimental viscosity for LiF which could be an example perhaps along with the VAF.

- [x] still need to fix the docs/ update the developer guide.
- [x] ~~perhaps we could have a helper function to create Velocity-Velocity correlations from a structure (say, creating correlation_kwargs for different atom types automatically). As it stands ```Observable``` does not attach to a particular atoms object and so does not have this information at construction. Except implicitly via ```atoms: list[int]```.~~
- [x] Velocity can take a ```SliceLike``` or ```list[int]```. So open-ended ranges can be specified i.e. ```slice(0, None)```, but also non-monotonic, decreasing or whatever sequences can be specified that cannot be represented as a slice.

bitmap

### DIFF
--- a/docs/source/apidoc/janus_core.rst
+++ b/docs/source/apidoc/janus_core.rst
@@ -278,6 +278,7 @@ janus\_core.processing.observables module
    :private-members:
    :undoc-members:
    :show-inheritance:
+   :inherited-members:
 
 janus\_core.processing.post\_process module
 -------------------------------------------
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -210,6 +210,5 @@ nitpick_ignore = [
     ("py:class", "Context"),
     ("py:class", "Path"),
     ("py:obj", "dimension"),
-    ("py:obj", "allowed_components"),
     ("py:obj", "janus_core.processing.observables.Stress.value_count")
 ]
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -203,9 +203,13 @@ nitpick_ignore = [
     ("py:class", "Architectures"),
     ("py:class", "Devices"),
     ("py:class", "MaybeSequence"),
+    ("py:class", "SliceLike"),
     ("py:class", "PathLike"),
     ("py:class", "Atoms"),
     ("py:class", "Calculator"),
     ("py:class", "Context"),
     ("py:class", "Path"),
+    ("py:obj", "dimension"),
+    ("py:obj", "allowed_components"),
+    ("py:obj", "janus_core.processing.observables.Stress.value_count")
 ]
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -209,6 +209,4 @@ nitpick_ignore = [
     ("py:class", "Calculator"),
     ("py:class", "Context"),
     ("py:class", "Path"),
-    ("py:obj", "dimension"),
-    ("py:obj", "janus_core.processing.observables.Stress.value_count")
 ]
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -187,23 +187,23 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-A ``janus_core.processing.observables.Observable`` abstracts obtaining a quantity derived from ``Atoms``. They may be used as kernels for input into analysis such as a correlation.
+A :class:`janus_core.processing.observables.Observable` abstracts obtaining a quantity derived from ``Atoms``. They may be used as kernels for input into analysis such as a correlation.
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable`` and are implemented within the ``janus_core.processing.observables`` module.
+Additional built-in observable quantities may be added for use by the :class:`janus_core.processing.correlator.Correlation` class. These should extend :class:`janus_core.processing.observables.Observable` and are implemented within the :py:mod:`janus_core.processing.observables` module.
 
-The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object. When used as part of a ``janus_core.processing.correlator.Correlation``, each value will be correlated and the results averaged.
+The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object. When used as part of a :class:`janus_core.processing.correlator.Correlation`, each value will be correlated and the results averaged.
 
-As an example of building a new ``Observable`` consider the ``janus_core.processing.observables.Stress`` built-in. The following steps may be taken:
+As an example of building a new ``Observable`` consider the :class:`janus_core.processing.observables.Stress` built-in. The following steps may be taken:
 
 1. Defining the observable.
 ---------------------------
 
-The stress tensor may be computed on an atoms object using ``Atoms.get_stress``. A user may wish to obtain a particular component, or perhaps only compute the stress on some subset of ``Atoms``. For example during a ``janus_core.calculations.md.MolecularDynamics`` run a user may wish to correlate only the off-diagonal components (shear stress), computed across all atoms.
+The stress tensor may be computed on an atoms object using ``Atoms.get_stress``. A user may wish to obtain a particular component, or perhaps only compute the stress on some subset of ``Atoms``. For example during a :class:`janus_core.calculations.md.MolecularDynamics` run a user may wish to correlate only the off-diagonal components (shear stress), computed across all atoms.
 
 2. Writing the ``__call__`` method.
 -----------------------------------
 
-In the call method we can use the base ``janus_core.processing.observables.Observable``'s optional atom selector ``atoms_slice`` to first define the subset of atoms to compute the stress for:
+In the call method we can use the base :class:`janus_core.processing.observables.Observable`'s optional atom selector ``atoms_slice`` to first define the subset of atoms to compute the stress for:
 
 .. code-block:: python
 
@@ -223,7 +223,7 @@ Next the stresses may be obtained from:
             / units.GPa
         )
 
-Finally, to facilitate handling components in a symbolic way, ``janus_core.processing.observables.ComponentMixin`` exists to parse ``str`` symbolic components to ``int`` indices by defining a suitable mapping. For the stress tensor (and the format of ``Atoms.get_stress``) a suitable mapping is defined in ``janus_core.processing.observables.Stress``'s ``__init__`` method:
+Finally, to facilitate handling components in a symbolic way, :class:`janus_core.processing.observables.ComponentMixin` exists to parse ``str`` symbolic components to ``int`` indices by defining a suitable mapping. For the stress tensor (and the format of ``Atoms.get_stress``) a suitable mapping is defined in :class:`janus_core.processing.observables.Stress`'s ``__init__`` method:
 
 .. code-block:: python
 
@@ -242,7 +242,7 @@ Finally, to facilitate handling components in a symbolic way, ``janus_core.proce
             },
         )
 
-This then concludes the ``__call__`` method for ``janus_core.processing.observables.Stress`` by using ``janus_core.processing.observables.ComponentMixin``'s
+This then concludes the ``__call__`` method for :class:`janus_core.processing.observables.Stress` by using :class:`janus_core.processing.observables.ComponentMixin`'s
 pre-calculated indices:
 
 .. code-block:: python
@@ -263,9 +263,9 @@ Since usually total system stresses are required we can define two built-ins to 
     StressHydrostatic = Stress(components=["xx", "yy", "zz"])
     StressShear = Stress(components=["xy", "yz", "zx"])
 
-Where by default ``janus_core.processing.observables.Observable``'s ``atoms_slice`` is ``slice(0, None, 1)``, which expands to all atoms in an ``Atoms``.
+Where by default :class:`janus_core.processing.observables.Observable`'s ``atoms_slice`` is ``slice(0, None, 1)``, which expands to all atoms in an ``Atoms``.
 
-For comparison the ``janus_core.processing.observables.Velocity`` built-in's ``__call__`` not only returns atom velocity for the requested components, but also returns them for every tracked atom i.e:
+For comparison the :class:`janus_core.processing.observables.Velocity` built-in's ``__call__`` not only returns atom velocity for the requested components, but also returns them for every tracked atom i.e:
 
 .. code-block:: python
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -187,32 +187,52 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable``. The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object:
+Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable``. The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object, and potentially ``value_count`` may need to be overloaded indicating the number of values returned by ``__call__``. The number of values returned will depend on the dimension of the underlying ``Observable``, and potentially the number of atoms resulting from the ``Observable``'s slice. These values are correlated individually and averaged.
+
+As an example of building a new ``Observable`` consider the ``janus_core.processing.observables.Stress`` and ``janus_core.processing.observables.Velocity`` built-ins. ``janus_core.processing.observables.Stress`` returns the required stress components as calculated by ``Atoms.get_stress`` called upon the tracked atoms. Therefore ``__call__`` returns one ``float`` per component correlated. The call method applies the atom slice and returns the desired stress components.
 
 .. code-block:: python
 
-    @abstractmethod
-    def __call__(self, atoms: Atoms) -> list[float]
+    def __call__(self, atoms: Atoms) -> list[float]:
+        sliced_atoms = atoms[self.atoms_slice]
+        sliced_atoms.calc = atoms.calc
+        stresses = (
+            sliced_atoms.get_stress(
+                include_ideal_gas=self.include_ideal_gas, voigt=True
+            )
+            / units.GPa
+        )
+        return stresses[self._indices]
 
-These values are returned as a ``list[float]`` representing the dimensions of the observed value which may be returned for a slice of atoms individually. ``value_count`` should be overloaded to return the expected length returned by ``__call__``. This is so enough correlators may be spawned to track the correlated the values. Note the final correlation will be the average of the correlations across these dimensions and atom counts.
-
-For example the ``janus_core.processing.observables.Stress`` built-in returns the required stress components as calculated by ``Atoms.get_stress`` called upon the tracked atoms. Therefore ``__call__`` returns one ``float`` per component correlated. For example an observable representing the ``xy`` and ``zy`` components of stress computed across all odd-index atoms in some ``Atoms`` object can be constructed as follows:
+For example an observable representing the ``xy`` and ``zy`` components of stress computed across all odd-index atoms in some ``Atoms`` object can be constructed as follows:
 
 .. code-block:: python
 
     s = Stress(components=["xy", "zy"], atoms_slice=(0, None, 2))
 
-In this case ``s(atoms)`` will return 2 values.
+In this case ``s(atoms)`` will return 2 values, no matter how many atoms are sliced from ``atoms_slice``.
 
-The ``janus_core.processing.observables.Velocity`` built-in's ``__call__`` not only returns atom velocity for the requested dimensions, but also returns them for every tracked atom. ``value_count`` is overloaded to reflect this. Therefore given the observable
+However, the ``janus_core.processing.observables.Velocity`` built-in's ``__call__`` not only returns atom velocity for the requested dimensions, but also returns them for every tracked atom, i.e:
+
+.. code-block:: python
+
+       def __call__(self, atoms: Atoms) -> list[float]:
+           return atoms.get_velocities()[self.atoms_slice, :][:, self._indices].flatten()
+
+so ``value_count`` is overloaded to reflect this.
+
+.. code-block:: python
+
+       def value_count(self, n_atoms: int | None = None) -> int:
+           return selector_len(self.atoms_slice, n_atoms) * self.dimension
+
+Unlike ``janus_core.processing.observables.Stress`` when we create a ``janus_core.processing.observables.Velocity`` observable like so,
 
 .. code-block:: python
 
     v = Velocity(components=["x", "y", "z"], atoms_slice=(0, None, 2))
 
-The value of ``v(atoms)`` will have the length ``3 * len(atoms)//2``. The 3 dimensions for each odd-indexed atom.
-
-New built-in observables are collected within the ``janus_core.processing.observables`` module. Special cases may also be defined for ease of use:
+the value of ``v(atoms)`` will have the length ``3 * len(atoms)//2``. The 3 dimensions of velocity for each odd-indexed atom. New built-in observables may be written by implementing new ``__call__`` methods to return the desired quantities. These should be collected within the ``janus_core.processing.observables`` module. Special cases may also be defined for ease of use, for example:
 
 .. code-block:: python
 
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -187,54 +187,87 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable``. The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object, and potentially ``value_count`` may need to be overloaded indicating the number of values returned by ``__call__``. The number of values returned will depend on the dimension of the underlying ``Observable``, and potentially the number of atoms resulting from the ``Observable``'s slice. These values are correlated individually and averaged.
+A ``janus_core.processing.observables.Observable`` abstracts obtaining a quantity derived from ``Atoms``. They may be used as kernels for input into analysis such as a correlation.
 
-As an example of building a new ``Observable`` consider the ``janus_core.processing.observables.Stress`` and ``janus_core.processing.observables.Velocity`` built-ins. ``janus_core.processing.observables.Stress`` returns the required stress components as calculated by ``Atoms.get_stress`` called upon the tracked atoms. Therefore ``__call__`` returns one ``float`` per component correlated. The call method applies the atom slice and returns the desired stress components.
+Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable`` and are implemented within the ``janus_core.processing.observables`` module.
+
+The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object. When used as part of a ``janus_core.processing.correlator.Correlation``, each value will be correlated and the results averaged.
+
+As an example of building a new ``Observable`` consider the ``janus_core.processing.observables.Stress`` built-in. The following steps may be taken:
+
+1. Defining the observable.
+---------------------------
+
+The stress tensor may be computed on an atoms object using ``Atoms.get_stress``. A user may wish to obtain a particular component, or perhaps only compute the stress on some subset of ``Atoms``. For example during a ``janus_core.calculations.md.MolecularDynamics`` run a user may wish to correlate only the off-diagonal components (shear stress), computed across all atoms.
+
+2. Writing the ``__call__`` method.
+-----------------------------------
+
+In the call method we can use the base ``janus_core.processing.observables.Observable``'s optional atom selector ``atoms_slice`` to first define the subset of atoms to compute the stress for:
 
 .. code-block:: python
 
     def __call__(self, atoms: Atoms) -> list[float]:
         sliced_atoms = atoms[self.atoms_slice]
+        # must be re-attached after slicing for get_stress
         sliced_atoms.calc = atoms.calc
-        stresses = (
+
+Next the stresses may be obtained from:
+
+.. code-block:: python
+
+    stresses = (
             sliced_atoms.get_stress(
                 include_ideal_gas=self.include_ideal_gas, voigt=True
             )
             / units.GPa
         )
-        return stresses[self._indices]
 
-For example an observable representing the ``xy`` and ``zy`` components of stress computed across all odd-index atoms in some ``Atoms`` object can be constructed as follows:
+Finally, to facilitate handling components in a symbolic way, ``janus_core.processing.observables.ComponentMixin`` exists to parse ``str`` symbolic components to ``int`` indices by defining a suitable mapping. For the stress tensor (and the format of ``Atoms.get_stress``) a suitable mapping is defined in ``janus_core.processing.observables.Stress``'s ``__init__`` method:
+
+.. code-block:: python
+
+        ComponentMixin.__init__(
+            self,
+            components={
+                "xx": 0,
+                "yy": 1,
+                "zz": 2,
+                "yz": 3,
+                "zy": 3,
+                "xz": 4,
+                "zx": 4,
+                "xy": 5,
+                "yx": 5,
+            },
+        )
+
+This then concludes the ``__call__`` method for ``janus_core.processing.observables.Stress`` by using ``janus_core.processing.observables.ComponentMixin``'s
+pre-calculated indices:
+
+.. code-block:: python
+
+    return stesses[self._indices]
+
+The combination of the above means a user may obtain, say, the ``xy`` and ``zy`` stress tensor components over odd-indexed atoms by calling the following observable on an ``Atoms``:
 
 .. code-block:: python
 
     s = Stress(components=["xy", "zy"], atoms_slice=(0, None, 2))
 
-In this case ``s(atoms)`` will return 2 values, no matter how many atoms are sliced from ``atoms_slice``.
 
-However, the ``janus_core.processing.observables.Velocity`` built-in's ``__call__`` not only returns atom velocity for the requested dimensions, but also returns them for every tracked atom, i.e:
-
-.. code-block:: python
-
-       def __call__(self, atoms: Atoms) -> list[float]:
-           return atoms.get_velocities()[self.atoms_slice, :][:, self._indices].flatten()
-
-so ``value_count`` is overloaded to reflect this.
-
-.. code-block:: python
-
-       def value_count(self, n_atoms: int | None = None) -> int:
-           return selector_len(self.atoms_slice, n_atoms) * self.dimension
-
-Unlike ``janus_core.processing.observables.Stress`` when we create a ``janus_core.processing.observables.Velocity`` observable like so,
-
-.. code-block:: python
-
-    v = Velocity(components=["x", "y", "z"], atoms_slice=(0, None, 2))
-
-the value of ``v(atoms)`` will have the length ``3 * len(atoms)//2``. The 3 dimensions of velocity for each odd-indexed atom. New built-in observables may be written by implementing new ``__call__`` methods to return the desired quantities. These should be collected within the ``janus_core.processing.observables`` module. Special cases may also be defined for ease of use, for example:
+Since usually total system stresses are required we can define two built-ins to handle the shear and hydrostatic stresses like so:
 
 .. code-block:: python
 
     StressHydrostatic = Stress(components=["xx", "yy", "zz"])
     StressShear = Stress(components=["xy", "yz", "zx"])
+
+Where by default ``janus_core.processing.observables.Observable``'s ``atoms_slice`` is ``slice(0, None, 1)``, which expands to all atoms in an ``Atoms``.
+
+For comparison the ``janus_core.processing.observables.Velocity`` built-in's ``__call__`` not only returns atom velocity for the requested components, but also returns them for every tracked atom i.e:
+
+.. code-block:: python
+
+    def __call__(self, atoms: Atoms) -> list[float]:
+        return atoms.get_velocities()[self.atoms_slice, :][:, self._indices].flatten()
--- a/docs/source/developer_guide/tutorial.rst
+++ b/docs/source/developer_guide/tutorial.rst
@@ -187,20 +187,34 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
-
-Built-in observables are collected within the ``janus_core.processing.observables`` module. For example the ``janus_core.processing.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:
+Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable``. The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object:
 
 .. code-block:: python
 
-    Stress("xy", False)
+    @abstractmethod
+    def __call__(self, atoms: Atoms) -> list[float]
 
-A new built-in observables can be implemented by a class with the method:
+These values are returned as a ``list[float]`` representing the dimensions of the observed value which may be returned for a slice of atoms individually. ``value_count`` should be overloaded to return the expected length returned by ``__call__``. This is so enough correlators may be spawned to track the correlated the values. Note the final correlation will be the average of the correlations across these dimensions and atom counts.
+
+For example the ``janus_core.processing.observables.Stress`` built-in returns the required stress components as calculated by ``Atoms.get_stress`` called upon the tracked atoms. Therefore ``__call__`` returns one ``float`` per component correlated. For example an observable representing the ``xy`` and ``zy`` components of stress computed across all odd-index atoms in some ``Atoms`` object can be constructed as follows:
 
 .. code-block:: python
 
-   def __call__(self, atoms: Atoms, *args, **kwargs) -> float
+    s = Stress(components=["xy", "zy"], atoms_slice=(0, None, 2))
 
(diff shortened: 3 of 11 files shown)


janus_core/helpers/correlator.py
```diff
@@ -209,6 +212,8 @@ def __init__(
 
         Parameters
         ----------
+        n_atoms : int
+            Number of possible atoms to track.
         a : tuple[Observable, tuple, dict]
             Getter for a and kwargs.
```
ElliottKasoar, line 218: This isn't new but this (and `b`) needs updating to `Union[Observable, tuple[Observable, tuple, dict]]`.

The docstring probably could be clarified more too.

It might just be me, but I read "getter" as a function that might return the class, whereas this seems to be expecting the observable itself to be given?

It's probably also worth clarifying the tuple and dict optional components e.g. whether only one is allowed (it looks like it from the tests `"a": Velocity(["x", "y", "z"], na),`) and what each actually corresponds to.

janus_core/helpers/correlator.py
```diff
@@ -196,6 +198,7 @@ class Correlation:
 
     def __init__(
         self,
+        n_atoms: int,
```
ElliottKasoar, line 201: Given that the order of these isn't particularly intuitive, might it make sense to start with `*,`, to force passing them as key words?

janus_core/helpers/janus_types.py
```diff
@@ -179,3 +131,122 @@ class EoSResults(TypedDict, total=False):
     bulk_modulus: float
     v_0: float
     e_0: float
+
+
+# pylint: disable=too-few-public-methods
+class Observable:
```
ElliottKasoar, line 137: This is probably complex enough now that it should be in its own module, or a module related to correlations/observables.

janus_core/helpers/janus_types.py
```diff
@@ -179,3 +131,122 @@ class EoSResults(TypedDict, total=False):
...
+        """
+        if self.atoms:
+            if isinstance(self.atoms, list):
+                return len(self.atoms)
+            if isinstance(self.atoms, int):
+                return 1
+
+            start = self.atoms.start
+            stop = self.atoms.stop
+            step = self.atoms.step
```
ElliottKasoar, line 228: I'm probably missing something, but where are these attributes from?

janus_core/helpers/correlator.py
```diff
@@ -178,6 +178,8 @@ class Correlation:
 
     Parameters
     ----------
+    n_atoms : int
+        Number of possible atoms to track.
```
ElliottKasoar, line 182: It's not entirely clear to me from this how this works?

Would you not expect all atoms to be tracked by default, and otherwise would you not want to specify something through something like a mask, rather than just a number (the first n atoms?).

I may be misunderstanding its purpose though.

ElliottKasoar (commented):
Looks like a good start, thanks @harveydevereux!

I suspect I knew a lot of the answers to my questions when we discussed this when you started, but maybe it's helpful coming to it with fresh eyes, so we can make things as clear as possible.

janus_core/helpers/correlator.py
harveydevereux: This annoys me but I don't see a way around. It is not the mask, but merely to apply a mask that may have an unspecified number of atoms that are resolved to the number of atoms in md. 

The correlation needs to know how many correlators to spawn (i.e. from the mask combined with an actual atoms object) for averaging over atoms. And these need to be in a set order for correct averaging over both dimensions, x, y, x and atoms.

I have added "masks" to the observables, i.e. the Velocity observable, in the same ways as post_process using ```SliceLike``` which allows for ```unspecified endings``` (```start, stop, step``` = ```5, None, 1```). This leaves them potentially not tied to a specific max atom count (ignoring bounds issues). They could have one eliminating this, but then the user will have to specify say an atoms object or count when creating a ```Velocity``` observable to set that. Perhaps that is ok.

Currently this value, ```n_atoms```, is passed when the Correlation is created, or in other words when it is attached to a MD object that has some atoms, and in particular a definite ```n_atoms```. So that the mask can be resolved into a specific number of correlators to be spawned. We could just spawn ```n_atoms``` always but that may be inefficient in some cases.

janus_core/helpers/janus_types.py
harveydevereux: I think it is yes, but I was hesitant. What do we call it? Since it will then form part of a user facing api t might be confusing to have ```janus_core.helpers.observables``` and ```janus_core.helpers.observable``` and the like.

janus_core/helpers/janus_types.py
harveydevereux: See above, this is the "mask" application. Currently only ```Velocity``` actually has a non-none value possible for ```self.atoms```. It could be a general kwarg for Observable, but then why would ```Stress``` allow atoms to be passed? Perhaps if you want stress computed over a subset of atoms, but I am unsure if that is doable easily in ase.

janus_core/helpers/correlator.py
harveydevereux: I will clarify those points.

On the "getter" though this exists solely to keep (after restructuring I have done here) the ability for a user to pass an arbitrary (argument conforming) function to create a custom Observable without defining a class. Maybe it is just clutter but a better name can be found. My thinking is it is a "getter of the observed value". Any suggestions welcome.

janus_core/helpers/janus_types.py
ElliottKasoar: > See above, this is the "mask" application. Currently only `Velocity` actually has a non-none value possible for `self.atoms`. 

I see, thanks. This is `atoms : Union[list[int], SliceLike]`, right?

One minor note is that I'd generally assume `atoms` would be an atoms object/list, so we could consider a less ambiguous name.

More significantly, lists and tuples don't have built-in `start`, `stop` or `step` attributes. Is this handled somewhere, or will this throw errors?

> It could be a general kwarg for Observable, but then why would `Stress` allow atoms to be passed? Perhaps if you want stress computed over a subset of atoms, but I am unsure if that is doable easily in ase.

I can imagine in general the ability to calculate any observable for an arbitrary subset could be useful, but if it's not trivial I think it makes sense to keep it case by case for now, as and when we have a useful application.

janus_core/helpers/janus_types.py
ElliottKasoar: Any reason not to just put it in `janus_core.helpers.observables` itself?

harveydevereux: > Any reason not to just put it in `janus_core.helpers.observables` itself?

I seem to recall some circular import issues, I will check

harveydevereux: The issue is I want to use ```StartStopStep``` and ```SliceLike``` in ```observables.py```, imported from ```janus_types.py```.

But also use ```Observable``` in ```janus_types.py``` imported from ```observable.py``` for ```CorrelationKwargs```, its a circle.

```
ImportError: cannot import name 'SliceLike' from partially initialized module 'janus_core.helpers.janus_types' (most likely due to a circular import) (/home/vol04/scarf1214/janus-core_devel/janus_core/helpers/janus_types.py)
```

It can be solved by double-defining ```StartStopStep``` and ```SliceLike``` in ```observables.py```... or maybe there is a better python solution I do not know?

```CorrelationKwargs``` is the only ```janus_type``` that requires another janus object in its definition. This is since ```Observable``` has increased in complexity to support the atom and dimension averaging automatically.

Comment (ElliottKasoar) [reply janus_core/helpers/janus_types.py:137]: I suppose one option could be to also move `CorrelationKwargs`, but I think that ought to be in `janus_types.py`, so it's not a particularly nice solution.

Assuming `SliceLike` is only used as a type hint, what about about gating the import with `TYPE_CHECKING`, or does it throw errors?

Comment (ElliottKasoar) [reply janus_core/helpers/janus_types.py]: Ok nice, that's probably the best option.

You can also add `from __future__ import annotations`, which means you can get rid of the quotes (ruff will then probably also ask you to convert `Union` to `x | y` etc.)

File: janus_core/helpers/observables.py [hunk 6]
```diff
@@ -1,54 +1,227 @@
...
+        -------
+        int
+            Atom count averaged over.
+        """
+        if self.atoms:
+            if isinstance(self.atoms, list):
+                return len(self.atoms)
+            if isinstance(self.atoms, int):
+                return 1
+
```
harveydevereux, line 100: Does this handle Tuple? Can we make a more general handler also with post_process as well

janus_core/helpers/correlator.py
harveydevereux: On ```Union[Observable, tuple[Observable, tuple, dict]]```. The idea was to support an bare Observable,

```Python
"a": Velocity(["x", "y", "z"], na)
```

but also an Observable that may have additional args and kwargs

```Python
"a": (SomeObservable(["x", "y", "z"], na), (arg1, arg2, ...), {"some kwarg": 0.1, ...}
```

Thinking about it though these are more complicated and not more useful than designing an Observable class that stores some args and kwargs on initialisation and applying them in ```__call__```, as they would need to be anyway. So I think they can also be removed. We could then have ```__call__(self, atoms: Atoms) -> list[float]``` too.

The only way I can see this being useful is if MD kept track of some args and kwargs for each Observable (that vary over a simulation) and passed them into ```Correlation.update(self, atoms: Atoms)``` but I cannot think of a use case.

It came from having the ability for a user to write a function (as apposed to a class deriving from ```Observable```) and pass that as an ```Observable``` where args and kwargs are useful.

janus_core/helpers/observables.py
```diff
@@ -1,72 +1,305 @@
...
+        Initialise an observable with a given dimensionality.
+
+        Parameters
+        ----------
+        dimension : int
+            The dimension of the observed data.
+        """
+        self._dimension = dimension
+
+    def __call__(self, atoms: Atoms) -> list[float]:
```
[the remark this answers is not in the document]
harveydevereux: yes it should, updating shortly

janus_core/processing/observables.py
```diff
@@ -1,72 +1,307 @@
...
+            The component symbols to check.
+
+        Raises
+        ------
+        ValueError
+            If any component is invalid.
+        """
+        for component in components:
+            if component not in self.allowed_components:
+                component_names = list(self._components.keys())
```
oerc0122, line 143:
```suggestion
```
`str.join` works on `set`s and indeed any iterable.

janus_core/processing/observables.py
```diff
@@ -1,72 +1,307 @@
...
+    def _indices(self) -> list[int]:
+        """
+        Get indices associated with self._components.
+
+        Returns
+        -------
+        list[int]
+            The indices for each self._components.
+        """
+        return [self._components[c] for c in self.components]
```
oerc0122, line 125: Is this just `list(self._components.values())`?

janus_core/processing/observables.py
```diff
@@ -1,72 +1,307 @@
...
+        ----------
+        components : str
+            The component symbols to check.
+
+        Raises
+        ------
+        ValueError
+            If any component is invalid.
+        """
+        for component in components:
```
oerc0122, line 141:
```suggestion
        for component in self.allowed_components.keys() - components.keys():
```
All keys not in allowed components.

janus_core/processing/observables.py
```diff
@@ -1,72 +1,307 @@
...
+            sliced_atoms.get_stress(
+                include_ideal_gas=self.include_ideal_gas, voigt=True
+            )
             / units.GPa
-        )
+        )[self._indices]
+
+
+StressDiagonal = Stress(components=["xx", "yy", "zz"])
+ShearStress = Stress(components=["xy", "yz", "zx"])
```
oerc0122, line 233: Don't like that this isn't `StressShear` to go with others. May also want `StressHydrostatic` rather than `Diagonal`?

janus_core/processing/observables.py
```diff
@@ -1,72 +1,307 @@
...
+        """
+        ComponentMixin.__init__(self, components={"x": 0, "y": 1, "z": 2})
+        self._set_components(components)
+
+        Observable.__init__(self, len(components))
+
+        if atoms_slice:
+            self.atoms_slice = atoms_slice
+        else:
+            self.atoms_slice = slice(0, None, 1)
```
oerc0122, line 273:
```suggestion
        self.atoms_slice = atoms_slice if atoms_slice else slice(0, None, 1)
```

janus_core/processing/observables.py
harveydevereux: those are both good ideas,```StressHydrostatic``` sounds good, unless someone can think of better I'll go with it

janus_core/processing/observables.py
```diff
@@ -138,9 +138,8 @@ def _set_components(self, components: list[str]):
         ValueError
             If any component is invalid.
         """
-        for component in components:
+        for component in self.allowed_components.keys() - components.keys():
             if component not in self.allowed_components:
-                component_names = list(self._components.keys())
                 raise ValueError(
                     f"'{component}' invalid, must be '{', '.join(component_names)}'"
                 )
```
oerc0122, line 145: Blindly applying this isn't right.

You no longer need to check the `if`, as that's done by the operation. `component_names` is not just undefined and need to refer to the components which are not defined. 

```python
if any(self.allowed_components.keys() - components.keys()):
    raise ValueError(
                    f"'{','.join(self.allowed_components.keys() - components.keys())}' invalid, must be in '{', '.join(self.allowed_components)}'"
    )

Comment (harveydevereux) [reply janus_core/processing/observables.py]: yes, unfortunately I did check the changes locally but then added them all together anyway... the code was odd due to the line length but this can be solved

File: janus_core/processing/observables.py (hunk 9 above)
Comment (harveydevereux) [reply janus_core/processing/observables.py]: this may lead to duplicate indices. e.g. stress supports both "xy" and "yx" as ```self._components```. ```self.components``` are the particular (say stress) components an observable "tracks" hence the need to access those particular values from the dict.

I've changed this so there is no ```allowed_components``` property, just a ```_allowed_components``` member to track this. Hopefully makes this clearer.

File: janus_core/processing/observables.py (hunk 13 above)
Comment (oerc0122) [reply janus_core/processing/observables.py]: It's my problem really, I click "suggestion" because I'm too lazy to match the formatting and then it renders with "add this" even when it's invalid code. I should change the type to `Python`

File: janus_core/helpers/janus_types.py [hunk 14]
```diff
@@ -28,7 +21,6 @@
 PathLike = Union[str, Path]
 StartStopStep = tuple[Optional[int], Optional[int], int]
 SliceLike = Union[slice, range, int, StartStopStep]
-
```
ElliottKasoar, line 31: Very minor but looks spacing looks a bit odd here now

janus_core/helpers/utils.py
```diff
@@ -407,3 +412,50 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
...
+
+    Returns
+    -------
+    StartStopStep
+        Standardized `SliceLike` as `start`, `stop`, `step` triplet.
+    """
+    if isinstance(index, int):
+        if index == -1:
+            return (index, None, 1)
+        return (index, index + 1, 1)
```
ElliottKasoar, line 434: Is it obvious why this makes sense?

I'd assumed it would be more like converting `range(index)`, or maybe just that specific value (`start == stop`), but I may be missing something

oerc0122: Python stops one early. A slice of `x[1:1]` is a zero-element slice, so to extract a single index we need `x[1:2]`.

For `index == -1`, `x[-1:0]` ie `-1+1`, that is also zero-element.

janus_core/processing/observables.py
```diff
@@ -1,72 +1,292 @@
...
-class Stress:
+        Raises
+        ------
+        ValueError
+            If any component is invalid.
+        """
+        if any(components - self._allowed_components.keys()):
+            raise ValueError(
+                f"'{components-self.allowed_components.keys()}'"
+                " invalid, must be '{', '.join(self._components)}'"
```
oerc0122, line 132:
```suggestion
                f" invalid, must be '{', '.join(self._components)}'"
```

janus_core/helpers/utils.py
ElliottKasoar: Ah ok so I think I was mixing up a couple of things.

I know Python uses half-open intervals, but for some reason I was thinking part of the point of  our `StartStopStep` was to do more like what we discussed a long time ago for the MD temperature ramp, where users may want to define both end points and we construct the range for them.

I see what you mean though.

I'm not sure the `-1` case works when it's used elsewhere though e.g. trying to use `slicelike_len_for`:

```

In [70]: x = [0,1,2,3,4,5,6,7,8,9,10]
In [71]: sliceable_length = len(x)
In [72]: slc = -1

In [73]: start, stop, step = slicelike_to_startstopstep(slc)
    ...: if stop is None:
    ...:     stop = sliceable_length
    ...: print(len(range(start, stop, step)))
12
```

I may be misusing `slicelike_len_for` though

janus_core/processing/observables.py
```diff
@@ -1,72 +1,292 @@
...
         return (
-            atoms.get_stress(include_ideal_gas=self.include_ideal_gas, voigt=True)[
-                self._index
-            ]
+            sliced_atoms.get_stress(
+                include_ideal_gas=self.include_ideal_gas, voigt=True
+            )
             / units.GPa
-        )
+        )[self._indices]
```
oerc0122, line 217: I think this is a bit messy and confusing, personally would prefer something more like:
```python
        stresses = sliced_atoms.get_stress(
                include_ideal_gas=self.include_ideal_gas, voigt=True
            ) / units.GPa
        return stresses[self._indices]
```

janus_core/processing/observables.py
```diff
@@ -1,72 +1,292 @@
...
+            Atom count to expand atoms_slice.
+
+        Returns
+        -------
+        int
+            The number of values returned by __call__.
+        """
+        if isinstance(self.atoms_slice, list):
+            return len(self.atoms_slice) * self.dimension
+        return slicelike_len_for(self.atoms_slice, self.n_atoms) * self.dimension
```
oerc0122, line 292: It might be best to move the list case into `slicelike_len_for` so we can have a clean interface here. Maybe `selector_len` or something in that case?

janus_core/helpers/utils.py
oerc0122: No, that looks to be a bug in `slicelike_len_for` due to the fact it has `-1:sliceable_len` i.e. `-1, 0, ... sliceable_len` == `len+2 == 12`

harveydevereux: I've updated ``slicelike_len_for``, now ``select_len`` (cf below) and this is also a test case. ``list`` is also handled with it as well as integers

janus_core/processing/observables.py
harveydevereux: done, but this also made me realise slicelike_len_for is not defined due to the import guarding (ultimately a circular import issue from before). @ElliottKasoar I have moved ```CorrelationKwargs``` into ```correlator.py``` so there is no longer an issue with ```Observable``` and ```SliceLike``` circular imports.

janus_core/processing/observables.py
harveydevereux: I've split the index onto the return line as above, it is clearer.

docs/source/developer_guide/tutorial.rst
```diff
@@ -187,20 +187,34 @@ Alternatively, using ``tox``::
...
 
-Built-in observables are collected within the ``janus_core.processing.observables`` module. For example the ``janus_core.processing.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:
+.. code-block:: python
+
+    @abstractmethod
+    def __call__(self, atoms: Atoms) -> list[float]
+
+These values are returned as a ``list[float]`` representing the dimensions of the observed value which may be returned for a slice of atoms individually. ``value_count`` should be overloaded to return the expected length returned by ``__call__``. This is so enough correlators may be spawned to track the correlated the values. Note the final correlation will be the average of the correlations across these dimensions and atom counts.
```
ElliottKasoar, line 197: > These values are returned as a ``list[float]`` representing the dimensions of the observed value which may be returned for a slice of atoms individually.

It's not entirely clear to me what this means

docs/source/developer_guide/tutorial.rst
```diff
@@ -187,20 +187,34 @@ Alternatively, using ``tox``::
...
+
+For example the ``janus_core.processing.observables.Stress`` built-in returns the required stress components as calculated by ``Atoms.get_stress`` called upon the tracked atoms. Therefore ``__call__`` returns one ``float`` per component correlated. For example an observable representing the ``xy`` and ``zy`` components of stress computed across all odd-index atoms in some ``Atoms`` object can be constructed as follows:
 
 .. code-block:: python
 
-    Stress("xy", False)
+    s = Stress(components=["xy", "zy"], atoms_slice=(0, None, 2))
 
-A new built-in observables can be implemented by a class with the method:
+In this case ``s(atoms)`` will return 2 values.
```
ElliottKasoar, line 205: This could be expanded, maybe with an actual example of the use with `atoms`. E.g. naively:

```
In [31]: s = Stress(components=["xy", "zy"], atoms_slice=(0, None, 2))

In [32]: a
Out[32]: Atoms(symbols='NaClNaClNaClNaCl', pbc=True, cell=[5.64, 5.64, 5.64], spacegroup_kinds=..., calculator=MACECalculator(...))

In [33]: s(a)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[33], line 1
----> 1 s(a)

File ~/PSDI/janus-core/janus_core/processing/observables.py:208, in Stress.__call__(self, atoms)
    194 def __call__(self, atoms: Atoms) -> list[float]:
    195     """
    196     Get the stress components.
    197
   (...)
    206         The stress components in GPa units.
    207     """
--> 208     sliced_atoms = atoms[self.atoms_slice]
    209     sliced_atoms.calc = atoms.calc
    210     stresses = (
    211         sliced_atoms.get_stress(
    212             include_ideal_gas=self.include_ideal_gas, voigt=True
    213         )
    214         / units.GPa
    215     )

File ~/PSDI/venv_janus_11_new/lib/python3.11/site-packages/ase/atoms.py:1151, in Atoms.__getitem__(self, i)
   1149 atoms.arrays = {}
   1150 for name, a in self.arrays.items():
-> 1151     atoms.arrays[name] = a[i].copy()
   1153 atoms.constraints = conadd
   1154 return atoms

IndexError: arrays used as indices must be of integer (or boolean) type
```

(I get a similar, but different error for a trajectory I read in, rather than just one `Atoms` structure)

docs/source/developer_guide/tutorial.rst
```diff
@@ -187,20 +187,34 @@ Alternatively, using ``tox``::
...
 =======================
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should extend ``janus_core.processing.observables.Observable``. The abstract method ``__call__`` should be implemented to obtain the values of the observed quantity from an ``Atoms`` object:
 
-Built-in observables are collected within the ``janus_core.processing.observables`` module. For example the ``janus_core.processing.observables.Stress`` observable allows a user to quickly setup a given correlation of stress tensor components (with and without the ideal gas contribution). An observable for the ``xy`` component is obtained without the ideal gas contribution as:
+.. code-block:: python
+
+    @abstractmethod
+    def __call__(self, atoms: Atoms) -> list[float]
```
ElliottKasoar, line 195: I think walking through how you actually go about constructing this function might be useful, too. This could be based on one of the existing methods, or maybe a relatively trivial example

docs/source/developer_guide/tutorial.rst
harveydevereux: The idea is that an observable is supporting something that is computed over a set of atoms and then returns a fix dimensional value, e.g. stress, or something that also returns a value across all atoms as in velocity. Not sure how to re word it to be more clear.

docs/source/developer_guide/tutorial.rst
harveydevereux: this appears to be a bug I must have introduced when apply review comments

docs/source/developer_guide/tutorial.rst
harveydevereux: I can expand the stress and velocity sections to detail more about how it was done there

harveydevereux:
> Not really work for this PR as such, but how easy would it be to integrate/use these in post processing, rather than on-the-fly calculations?

All it needs is to be called on some sequence of atoms objects. But if you are post-processing why not just use post-process? The only advantage of these is they extract the data from Atoms in this case

ElliottKasoar:
> All it needs is to be called on some sequence of atoms objects. But if you are post-processing why not just use post-process? The only advantage of these is they extract the data from Atoms in this case

Most significantly, I'm thinking about how easy it is to extend postprocess to observables we've defined, but aren't supported yet. 

At the moment we can only do RDFs and VAFs, so if we didn't calculate any other observable on the fly, we might still want to be able to use an `Observable`.

It could also fit into unifying the interfaces e.g. for the VAF, depending on how their efficiencies compare, which would avoid questions about their consistency any time we make a change.

harveydevereux:
> Most significantly, I'm thinking about how easy it is to extend postprocess to observables we've defined, but aren't supported yet.
> 
> At the moment we can only do RDFs and VAFs, so if we didn't calculate any other observable on the fly, we might still want to be able to use an `Observable`.
> 
> It could also fit into unifying the interfaces e.g. for the VAF, depending on how their efficiencies compare, which would avoid questions about their consistency any time we make a change.

We could have a general ```correlation``` post-processing routine that could take in observable pairs and correlation arguments and compute the correlations over some given ```list[Atoms]``` or the like (trajectory).  The observables being convenient kernels to collect possibly complicated data out of ```Atoms```.

VAF is specialised to VAF of course, but really it just collects velocity data, sliced over some atoms, to then correlate. 

On-the-fly is the same of course but we don't start with all the data available, and can even bin it at soon as it is consumed to save i/o or memory. So conceptually it gets "attached" to a "generator" (md). That generator could just be anything that yields new ```Atoms``` though.

docs/source/developer_guide/tutorial.rst
harveydevereux: Fixed, and also the vaf test case now tests using ```slice```, ```range```, ```list[int]```, ```StartStopStep``` as atom slices

tests/test_utils.py
```diff
@@ -155,3 +161,41 @@ def test_none_to_dict(dicts_in):
...
+        ((range(10), 10), 10),
+        ((slice(0, None, 2), 10), 5),
+        (([-2, -1, 0], 9), 3),
+        (([-1], 10), 1),
+        (([0, -1, 2, 9], 10), 4),
+    ],
+)
+def test_selector_len(slc_len: tuple[Union[SliceLike, list], int], expected: int):
+    """Test converting SliceLike to StartStopStep."""
+    assert selector_len(*slc_len) == expected
```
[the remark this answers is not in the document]
harveydevereux: it does look a little cleaner

docs/source/conf.py
```diff
@@ -203,9 +203,13 @@
...
     ("py:class", "MaybeSequence"),
+    ("py:class", "SliceLike"),
     ("py:class", "PathLike"),
     ("py:class", "Atoms"),
     ("py:class", "Calculator"),
     ("py:class", "Context"),
     ("py:class", "Path"),
+    ("py:obj", "dimension"),
+    ("py:obj", "allowed_components"),
+    ("py:obj", "janus_core.processing.observables.Stress.value_count")
```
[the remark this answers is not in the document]
harveydevereux: just tried each, for the moment yes, except allowed_components which is no longer with us

janus_core/helpers/utils.py
```diff
@@ -407,3 +412,54 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
...
+
+    if isinstance(index, (slice, range)):
+        return (index.start, index.stop, index.step)
+
+    return index
```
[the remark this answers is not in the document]
harveydevereux: do you mean throwing an error if it is not actually a ```SliceLike```? 

i.e. these cases are no good but "run"

```python
>>> slicelike_to_startstopstep([1,2])
[1, 2]
>>> slicelike_to_startstopstep((None, None, None))
(None, None, None)
```

its probably a good move

docs/source/developer_guide/tutorial.rst
harveydevereux: along with the other doc comment above, I have written the docs walking through writing the ```Stress``` and ```Velocity``` methods and their meaning

janus_core/helpers/utils.py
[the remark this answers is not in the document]
harveydevereux: We could do

```python
def validate_slicelike(maybe_slicelike: SliceLike):
    # ...
    if isinstance(maybe_slicelike, (slice, range, int)):
        return
    if isinstance(maybe_slicelike, tuple) and len(maybe_slicelike) == 3:
        start, stop, step = maybe_slicelike
        if (
            isinstance(start, Optional[int])
            and isinstance(stop, Optional[int])
            and isinstance(step, int)
        ):
            return

    raise ValueError(f"{maybe_slicelike} is not a valid SliceLike")
```
?

janus_core/helpers/utils.py
```diff
@@ -407,3 +412,83 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
 
     with progress:
         yield from progress.track(sequence, description=description)
+
+
+def validate_slicelike(maybe_slicelike: SliceLike):
```
oerc0122, line 419:
```suggestion
def validate_slicelike(maybe_slicelike: SliceLike) -> None:
```

oerc0122 (commented):
Just so we know it's being run for side-effects.

janus_core/processing/observables.py
```diff
@@ -2,73 +2,297 @@
...
-            The stress component in GPa units.
+        list[float]
+            The stress components in GPa units.
         """
-        return (
-            atoms.get_stress(include_ideal_gas=self.include_ideal_gas, voigt=True)[
-                self._index
-            ]
+        sliced_atoms = atoms[self.atoms_slice]
+        sliced_atoms.calc = atoms.calc
```
ElliottKasoar, line 220: You may want to check `Atoms` isn't a list/traj, otherwise the slice works and you reach `AttributeError: 'list' object has no attribute 'calc'`, which may not be the clearest error.

docs/source/developer_guide/tutorial.rst
ElliottKasoar, line 1: I think a few things could still be more explicit.

I would aim to require as little further reading of the current code as possible, and at the moment there are quite a few assumptions.

- Can you change the `__call__` parameters, or would it always be an `Atoms` object (could it take a list?)
- What actually is ``value_count`, when/how is it used (briefly), and what is the default value (so why does it not need to be overloaded in the first case?
- What is `sliced_atoms`/when it is defined/is using atoms[self.atoms_slice]` pretty standard for any new observable, and do you always need to attach the calc (I assume yes)?
- What is `_indices`/where/when/how is is defined (briefly).

Essentially, I think breaking it down by line and explaining what the code is doing and why would be helpful! In particular, which parts are relatively general e.g.  the slicing, inherited attributes etc.

Some of the text is also relatively dense, particularly the start, so a few more spaces might be helpful.

harveydevereux: > * Can you change the `__call__` parameters, or would it always be an `Atoms` object (could it take a list?)
We could re-design it to do that, I guess to get values as a batch/average, which may be useful when using these things generally. As part of md we could have some quantity that takes in rolling windows of the trajectory, maybe averages them or whatever. It will require some thought on the user input side as to how to control that

> * What actually is `value_count`, when/how is it used (briefly), and what is the default value (so why does it not need to be overloaded in the first case?

This is a poor point of the design I don't like as well. I will try to re-design the correlations so that they are spawned dynamically, eliminating this. It comes from wanting to know how many correlators to spawn for an unknown observable. But this could be done in the first update call perhaps?


I shall add more explanation of generally covering these (the ```calc``` is because slicing the atoms seems to get rid of the calculator, if you know a way to do it without that happening, that would be great)
> * What is `sliced_atoms`/when it is defined/is using `atoms[self.atoms_slice]` pretty standard for any new observable, and do you always need to attach the `calc` (I assume yes)?

> * What is `_indices`/where/when/how is it defined (briefly).

janus_core/helpers/utils.py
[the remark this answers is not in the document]
ElliottKasoar: Yes, that sort of thing would be great

docs/source/developer_guide/tutorial.rst
ElliottKasoar: Ah so is `value_count` now removed completely? It reads a lot more clearly to me now, thanks!

harveydevereux: > Ah so is `value_count` now removed completely? It reads a lot more clearly to me now, thanks!

yes exactly, writing the docs got me thinking about it. It is the correlations that need to know this, and it can be done dynamically. Observables should not have to care, only their consumers

docs/source/developer_guide/tutorial.rst
ElliottKasoar: Thanks!

janus_core/helpers/utils.py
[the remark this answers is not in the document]
harveydevereux: Been added in https://github.com/stfc/janus-core/pull/285/files#diff-939a9a0d22343ac06615c18e3467b1ef73350a6a4fc1155337e2a0d780c0d751R419

and test https://github.com/stfc/janus-core/pull/285/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R221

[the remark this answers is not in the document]
ElliottKasoar: Ah cool thanks!

janus_core/processing/observables.py
```diff
@@ -2,73 +2,254 @@
...
+    """
+
+    def __init__(self, atoms_slice: list[int] | SliceLike | None = None):
+        """
+        Initialise an observable with a given dimensionality.
+
+        Parameters
+        ----------
+        atoms_slice : list[int] | SliceLike | None = None
+            A slice of atoms to observe.
```
ElliottKasoar, line 34:
```suggestion
        atoms_slice : list[int] | SliceLike | None
            A slice of atoms to observe. By default all atoms are included.
```
Or similar i.e. describe what happens if `None`

janus_core/processing/observables.py
```diff
@@ -2,73 +2,254 @@
...
+                "zz": 2,
+                "yz": 3,
+                "zy": 3,
+                "xz": 4,
+                "zx": 4,
+                "xy": 5,
+                "yx": 5,
+            },
+        )
+        self._set_components(components)
```
ElliottKasoar, line 165: Does it make sense to make this into a setter, rather than explicitly calling this here (and in `janus_core/processing/observables.py`)?

harveydevereux: As in instead of the to calls, one to __init__ and one to _set_components, just having one setter? If so we might as well roll it into one

harveydevereux: right I see what you mean

harveydevereux: @ElliottKasoar I've made ```self.components``` a property with a ```@components.setter``` so we have this behaviour/style now

docs/source/apidoc/janus_core.rst
```diff
@@ -278,6 +278,7 @@ janus\_core.processing.observables module
    :private-members:
    :undoc-members:
    :show-inheritance:
+   :inherited-members:
```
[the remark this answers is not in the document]
ElliottKasoar: It doesn't matter too much, but does it not still show `components` without this line? I tried removing it, and I still get

```
class janus_core.processing.observables.Velocity(*, components: list[str], atoms_slice: list[int] | SliceLike | None = None)
...
    Parameters:
        components: list[str]
            Symbols for velocity components, x, y, z.
...
    Attributes:
        components
            Get the symbolic components of the observable.
...
    property components: list[str]
        Get the symbolic components of the observable.
        Returns:
            list[str]
                The observables components.
```

docs/source/developer_guide/tutorial.rst
```diff
@@ -187,20 +187,87 @@ Alternatively, using ``tox``::
 Adding a new Observable
 =======================
 
-Additional built-in observable quantities may be added for use by the ``janus_core.processing.correlator.Correlation`` class. These should conform to the ``__call__`` signature of ``janus_core.helpers.janus_types.Observable``. For a user this can be accomplished by writing a function, or class also implementing a commensurate ``__call__``.
+A ``janus_core.processing.observables.Observable`` abstracts obtaining a quantity derived from ``Atoms``. They may be used as kernels for input into analysis such as a correlation.
```
oerc0122, line 190:
```suggestion
A :class:`janus_core.processing.observables.Observable` abstracts obtaining a quantity derived from ``Atoms``. They may be used as kernels for input into analysis such as a correlation.
```
Does this work?

docs/source/apidoc/janus_core.rst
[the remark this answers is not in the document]
harveydevereux: > It doesn't matter too much, but does it not still show `components` without this line? I tried removing it, and I still get
> 
> ```
> class janus_core.processing.observables.Velocity(*, components: list[str], atoms_slice: list[int] | SliceLike | None = None)
> ...
>     Parameters:
>         components: list[str]
>             Symbols for velocity components, x, y, z.
> ...
>     Attributes:
>         components
>             Get the symbolic components of the observable.
> ...
>     property components: list[str]
>         Get the symbolic components of the observable.
>         Returns:
>             list[str]
>                 The observables components.
> ```

It probably does, I did not check. But if we remove it then I was getting a warning and ci failure and this seemed better than ignoring it in the sphinx config

janus_core/processing/correlator.py
```diff
@@ -179,10 +179,12 @@ class Correlation:
 
     Parameters
     ----------
-    a : tuple[Observable, dict]
-        Getter for a and kwargs.
-    b : tuple[Observable, dict]
-        Getter for b and kwargs.
+    n_atoms : int
+        Number of possible atoms to track.
```
ElliottKasoar, line 183: Sorry, I think we've discussed this a few times already, but does this only refer to the number of atoms being tracked when you want to store a quantity for each of those atoms, as opposed to an observable that could depend on `n_atoms`, but we're only interested in a single value for it?

I assume they're quite different, but to me the description could refer to either?

janus_core/helpers/utils.py
```diff
@@ -432,3 +437,83 @@ def check_files_exist(config: dict, req_file_keys: Sequence[PathLike]) -> None:
...
+        Candidate to test.
+
+    Raises
+    ------
+    ValueError
+        If maybe_slicelike is not SliceLike.
+    """
+    if isinstance(maybe_slicelike, (slice, range, int)):
+        return
+    if isinstance(maybe_slicelike, tuple) and len(maybe_slicelike) == 3:
```
oerc0122, line 458: What if it's a `list` or some other sequence like I've done a `map`? Would it be better to use `itertools.islice` to take the front 3, check the iterator's empty and then check the properties?

**N.B.** being restrictive is fine, it's just whether we want to be.

docs/source/apidoc/janus_core.rst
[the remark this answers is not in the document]
ElliottKasoar: Fair enough, seems sensible!

janus_core/processing/observables.py
```diff
@@ -2,73 +2,267 @@
...
+        atoms_slice : list[int] | SliceLike | None
+            A slice of atoms to observe. By default all atoms are included.
+        """
+        if atoms_slice:
+            if isinstance(atoms_slice, list):
+                self.atoms_slice = atoms_slice
+            else:
+                self.atoms_slice = slice(*slicelike_to_startstopstep(atoms_slice))
+        else:
+            self.atoms_slice = slice(0, None, 1)
```
oerc0122, line 42:
```suggestion
        if not atoms_slice:
            self.atoms_slice = slice(0, None, 1)
            return
            
        if isinstance(atoms_slice, list):
            self.atoms_slice = atoms_slice
        else:
            self.atoms_slice = slice(*slicelike_to_startstopstep(atoms_slice))
```
Could remove a level of nesting with an early return making the logic easier to follow

janus_core/processing/correlator.py
harveydevereux: I don't think it is required actually, since now it is all inferred. Must have got left or re-added on a re-base I will check

janus_core/helpers/utils.py
harveydevereux: At the moment I have been following the definition hence the restriction. ```list``` could be included, but then will this effect post-processing and any other users of ```SliceLike```? Perhaps those functions do work with ```list``` but I have not tried. I don't have strong opinions either way.

```
StartStopStep = tuple[Optional[int], Optional[int], int]
SliceLike = Union[slice, range, int, StartStopStep]
```

oerc0122: We can always expand them later. It's not too hard.

janus_core/processing/correlator.py
```diff
@@ -278,7 +284,10 @@ def get(self) -> tuple[Iterable[float], Iterable[float]]:
         lags : Iterable[float]]
             The correlation lag times t'.
         """
-        return self._correlator.get()
+        if self._correlators:
+            _, lags = self._correlators[0].get()
+            return np.mean([cor.get()[0] for cor in self._correlators], axis=0), lags
```
ElliottKasoar, line 284: Does calling `get` an extra time for the lags not also recalculate non-trivial things that we don't need? Maybe we want a separate `get_lags` for this?

docs/source/developer_guide/tutorial.rst
harveydevereux: they do, added in, also for modules

janus_core/processing/correlator.py
harveydevereux: Also that the lags are recomputed each get, but we only need them once. They are split now

ElliottKasoar (approved):
Thanks for all the work, @harveydevereux!

I'll give it to the end of the day before merging, but I think it looks good from my perspective.

My only other comment, which I may make into a new issue (and assign to you, if that's ok?), is that some of the (existing) logic could probably do with some comments. E.g. `_propagate`, possibly partly as if uses attributes that aren't explicitly described themselves.

Beyond the scope of this PR though, anyway.

alinelena:
independent of this I will like a fully worked example in janus-tuorials.

harveydevereux:
> Thanks for all the work, @harveydevereux!
> 
> I'll give it to the end of the day before merging, but I think it looks good from my perspective.
> 
> My only other comment, which I may make into a new issue (and assign to you, if that's ok?), is that some of the (existing) logic could probably do with some comments. E.g. `_propagate`, possibly partly as if uses attributes that aren't explicitly described themselves.
> 
> Beyond the scope of this PR though, anyway.

Thanks @ElliottKasoar and yes go for it. I can also open an issue for janus-tutorials

PR #291, opened 2024-08-28
Fix set_calc default when reading structure file

### LINKED ISSUE #290: reusing and extxyz file wrote by janus crashes the workflow
do a geom opt of any system and save it as extxyz, or default

try to use it for example in another geom opt but with stricter params... 

you will get a stress not found or force not found...

ElliottKasoar:
Resolves #290

Fixes setting default for `set_calc` to be set to `True` if `struct_path` is specified.

This was supposed to be the case already, but was being overwritten by the structures read from file having `SinglePointCalculators `attached, due to their saved results, leading to errors.

Test added fails before fix implemented.

### DIFF
--- a/janus_core/helpers/utils.py
+++ b/janus_core/helpers/utils.py
@@ -369,10 +369,10 @@ def input_structs(
         if struct_path:
             set_calc = True
 
-        if isinstance(struct, Atoms):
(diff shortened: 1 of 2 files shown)


oerc0122 (approved):
If it does what's needed, looks good.

PR #292, opened 2024-08-28
Add auto restart

### LINKED ISSUE #169: Add auto restart for MD
> 3. autorestart shall be a different option and only for MD not for heating(this is to workaround supercomputers with easy access on preemptable queues.)

_Originally posted by @alinelena in https://github.com/stfc/janus-core/issues/108#issuecomment-2134808423_

ElliottKasoar:
Resolves #169

Adds auto restart option for MD, which generates `restart_stem` based on inputs (including `file_prefix` and `restart_stem`), and if restarting simulation, uses the most recent matching file for the input structure.

Also includes fixes for:

- `prefix_override`, which threw errors if given a `Path`s (e.g. anything via `--restart-stem`)
- How `restart_stem` was used, as currently for `restart_stem=stem`, files of the form `stem-T300.0-res-1.extxyz` are created, the same as the old usage of `struct_name`. With these changes, they should now be `stem-1.extxyz`

Note: the limit after restarting is the effectively the same as before - however far the previous simulation got + the full `steps` input (e.g. for `NaCl-nvt-T300.0-res-100.extxyz`, with `--steps 1000`, the last step would be 1100.)

One thing to consider is that this implementation can (and arguably is likely) to lead to duplicated steps for statistics and/or trajectory file.

E.g. if you were to print statistics every 100 steps, but restart every 1000 steps, then the statistics file may to be written for a while after the last restart file. When we restart, we then append duplicate steps.

Do we want to delete these duplicate steps? This is even harder for the trajectory, since we'd have to calculate which images would be duplicated.

### DIFF
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -196,11 +196,17 @@ For example:
 
 .. code-block:: bash
 
-    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart-every 100
-    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1200 --stats-every 10 --traj-every 100 --restart-every 1000
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1200 --stats-every 10 --traj-every 100 --restart
 
 will create, then restart from, ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``,
 running an additional 1000 steps, and appending the statistics and trajectory files created by the first command.
+
+
+.. note::
+    Depending on the frequency of statistics, trajectory, and restart file outputs, it is likely that outputs for some steps may be repeated.
+    These can be identified using the ``Step`` column in the statistics file, and the ``step`` info label for each trajectory image.
+    For example, in the above example, outputs between steps 1000 and 1200 would be repeated.
 
 
 For all options, run ``janus md --help``.
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -190,6 +190,18 @@ both before beginning dynamics and at steps 100 and 200 of the simulation.
 This performs an NVE molecular dynamics simulation at 300K for 200 steps (0.2 ps), saving the trajectory every 10 steps after the first 100, and the thermodynamical statistics every 10 steps,
 as well as changing the output file names for both.
 
+To restart a simulation, the restart file can be entered explicitly (e.g. ``--struct NaCl-npt-T300.0-p1.0-res-1000.extxyz``) in combination with the ``--restart`` option,
+but by default the ``--restart-auto`` option allows the same original structure to be used to infer the most recent restart file created.
+For example:
+
+.. code-block:: bash
+
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart-every 100
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart
+
+will create, then restart from, ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``,
+running an additional 1000 steps, and appending the statistics and trajectory files created by the first command.
+
 
 For all options, run ``janus md --help``.
 
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -4,6 +4,7 @@ import datetime
 from functools import partial
 from itertools import combinations_with_replacement
 from math import isclose
+from os.path import getmtime
 from pathlib import Path
 import random
 from typing import Any, Optional, Union
@@ -38,6 +39,7 @@ from janus_core.helpers.utils import (
     Architectures,
     ASEReadArgs,
     Devices,
+    input_structs,
     none_to_dict,
     output_structs,
     write_table,
@@ -107,6 +109,8 @@ class MolecularDynamics(BaseCalculation):
         and temperature.
     restart : bool
         Whether restarting dynamics. Default is False.
+    restart_auto : bool
+        Whether to infer restart file name if restarting dynamics. Default is True.
     restart_stem : str
         Stem for restart file name. Default inferred from `file_prefix`.
     restart_every : int
@@ -197,6 +201,7 @@ class MolecularDynamics(BaseCalculation):
         rescale_every: int = 10,
         file_prefix: Optional[PathLike] = None,
         restart: bool = False,
+        restart_auto: bool = True,
         restart_stem: Optional[PathLike] = None,
         restart_every: int = 1000,
         rotate_restart: bool = False,
@@ -276,6 +281,8 @@ class MolecularDynamics(BaseCalculation):
             and temperature.
         restart : bool
             Whether restarting dynamics. Default is False.
+        restart_auto : bool
+            Whether to infer restart file name if restarting dynamics. Default is True.
         restart_stem : str
             Stem for restart file name. Default inferred from `file_prefix`.
         restart_every : int
@@ -350,6 +357,7 @@ class MolecularDynamics(BaseCalculation):
         self.remove_rot = remove_rot
         self.rescale_every = rescale_every
         self.restart = restart
+        self.restart_auto = restart_auto
         self.restart_stem = restart_stem
         self.restart_every = restart_every
         self.rotate_restart = rotate_restart
@@ -483,11 +491,13 @@ class MolecularDynamics(BaseCalculation):
                 "filemode": "a",
             }
 
-        self.restart_files = []
         self.dyn: Union[Langevin, VelocityVerlet, ASE_NPT]
         self.n_atoms = len(self.struct)
 
         self.offset = 0
+        if self.restart:
+            self._prepare_restart()
+        self.restart_files = []
         self.created_final_file = False
 
         if "masses" not in self.struct.arrays:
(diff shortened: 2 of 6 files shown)


alinelena:
I would not delete the steps ideally one write restart at the same step as printing, I will add maybe a note to that effect is the two are different. let us not defend the users from themselves

ElliottKasoar:
Hopefully there would be a corresponding step, but realistically it's quite wasteful to generate a restart file as often as you'd want to output the statistics, so I'd expect this would happen for most use cases unavoidably.

ElliottKasoar:
Just to make sure we're on the same page, I'm talking about something like 

```
janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 10 --restart-every 100
janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 10 --restart-every 100  --restart
```

If the first command fails at step 140, then the `-stats` file will have steps 100, 110, 120, 130 and 140.

When we (auto) restart, it resumes from step 100, so we end up with stats for 110, 120, 130 and 140 again.

My issues is (1) I don't think the user should have to set `restart_every` to match `stats_every`, and even then this could be inconsistent with `traj_every`, and (2) there's not a nice way to avoid this otherwise, that I can think of...

ElliottKasoar:
From discussions with @oerc0122, perhaps for now it might be sufficient to add a label to each trajectory image (`dyn_step`?), which means the user can relatively straightforwardly determine from both the stats and traj files where any overlap is?

alinelena:
I will call it time_step or timestep otherwise is perfectly ok... maybe since you are at this add another time= time in fs...

docs/source/user_guide/command_line.rst
```diff
@@ -190,6 +190,18 @@ both before beginning dynamics and at steps 100 and 200 of the simulation.
...
+To restart a simulation, the restart file can be entered explicitly (e.g. ``--struct NaCl-npt-T300.0-p1.0-res-1000.extxyz``) in combination with the ``--restart`` option,
+but by default the ``--restart-auto`` option allows the same original structure to be used to infer the most recent restart file created.
+For example:
+
+.. code-block:: bash
+
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart-every 100
+    janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 1000 --stats-every 1 --restart
+
+will create, then restart from, ``NaCl-npt-T300.0-p1.0-res-1000.extxyz``,
```
alinelena, line 202: is NaCl-md-? now

janus_core/calculations/md.py
```diff
@@ -4,6 +4,7 @@
 from functools import partial
 from itertools import combinations_with_replacement
 from math import isclose
+from os.path import getmtime
```
alinelena, line 7: do we need os.path? cannot be done via Pathlib?

alinelena: indeed https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat this can do it

docs/source/user_guide/command_line.rst
ElliottKasoar: Not at the moment, no, I didn't realised that's what you wanted from your comment in #280.

At the moment, it's only added to the log and summary files, which I thought made sense since otherwise they're indistinguishable, whereas `-res`, `-stats`, `-traj` etc. files are all unique to each calculation already.

janus_core/calculations/md.py
ElliottKasoar: I did look into that, but don't think we can directly use it as a `sorted` `key` in the same way since you access it through the `st_mtime` attribute, rather than a function.

I think we could still use it through a lambda function, but it seemed unnecessarily messy.

Given that `Path.stat` returns an `os.stat_result` object anyway, I'm not sure we would gain much anyway?

ElliottKasoar:
> I will call it time_step or timestep otherwise is perfectly ok... maybe since you are at this add another time= time in fs...

It turned out we already had an `info["step"]` from when we set the statistics, but that meant it would only update at the stats frequency, so wasn't necessarily accurate at the time each trajectory step is written out.

I think the varying frequencies of updating different info keys is still a problem for others e.g. "density" and "real_time", but we can think about that in a separate issue.

janus_core/calculations/md.py
oerc0122: `operator.attrgetter` would do it but it wouldn't necessarily be cleaner.

janus_core/calculations/md.py
```diff
@@ -499,6 +509,80 @@ def __init__(
...
+        # Check offset can be read from steps
+        try:
+            with open(self.stats_file, encoding="utf8") as stats_file:
+                last_line = stats_file.readlines()[-1].split()
+            try:
+                self.offset = int(last_line[0])
+            except (IndexError, ValueError) as e:
+                raise ValueError("Unable to read restart file") from e
+        except FileNotFoundError as e:
+            raise FileNotFoundError("Unable to read restart file") from e
```
oerc0122, line 530: Do we need this nested? The outer one would probably handle all cases, particularly since the error messages are the same.

janus_core/calculations/md.py
```diff
@@ -499,6 +509,80 @@ def __init__(
...
+                    set_calc=True,
+                    logger=self.logger,
+                )
+
+                # Infer last dynamics step
+                last_stem = last_restart.stem
+                try:
+                    # Remove restart_stem from filename
+                    # Use restart_stem.name otherwise T300.0 etc. counts as extension
+                    self.offset = int("".join(last_stem.split(f"{restart_stem.name}-")))
```
oerc0122, line 559: The logic here is confusing. Would it be better to just split on `-` and take the appropriate index? Or is that not possible? Can you give me some examples of the filenames.

janus_core/helpers/utils.py
```diff
@@ -146,7 +146,7 @@ def _build_filename(
             built_filename = Path(filename)
         else:
             prefix = (
-                prefix_override
+                str(prefix_override)
                 if prefix_override is not None
                 else str(self.file_prefix)
             )
```
oerc0122, line 152:
```python
            prefix = str(
                prefix_override
                if prefix_override is not None
                else self.file_prefix
            )
```

tests/test_md.py
```diff
@@ -884,3 +885,176 @@ def test_logging(tmp_path):
...
+    """Test auto restarting simulation."""
+    # tmp_path for all files other than restart
+    # Include T300.0 to test Path.stem vs Path.name
+    final_path = tmp_path / "md-T300.0-final.extxyz"
+    traj_path = tmp_path / "md-T300.0-traj.extxyz"
+    stats_path = tmp_path / "md-T300.0-stats.dat"
+    log_file = tmp_path / "md.log"
+
+    # Predicted restart file, from defaults
+    restart_path = Path(".").absolute() / "NaCl-nvt-T300.0-res-4.extxyz"
```
oerc0122, line 900: Do we want restart in the cwd or with output?

janus_core/calculations/md.py
```diff
@@ -499,6 +509,80 @@ def __init__(
...
+        step = self.offset + self.dyn.nsteps
+        self.dyn.atoms.info["time_fs"] = time
+        self.dyn.atoms.info["step"] = step
+
+    def _prepare_restart(self) -> None:
+        """Prepare restart files, structure and offset."""
+        # Check offset can be read from steps
+        try:
+            with open(self.stats_file, encoding="utf8") as stats_file:
+                last_line = stats_file.readlines()[-1].split()
```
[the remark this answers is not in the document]
ElliottKasoar: Good point.

Since this is just maintaining the [current behaviour](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L907), I'll open this as a separate issue

janus_core/calculations/md.py
ElliottKasoar: Thanks. Unless someone definitely thinks it's a cleaner alternative I think I'll leave it as it is.

janus_core/calculations/md.py
ElliottKasoar: I'd probably prefer to distinguish between the two messages better, rather than compress it and potentially leave it more ambiguous what the cause of the error was.

oerc0122: The context will come from `e` just as it does here.

ElliottKasoar: Yes, but a better message like "Unable to infer step from last line of restart file" makes the problem more immediately clear than a generic message with the IndexError/ValueError context and needing to follow the trace to understand what caused them

janus_core/calculations/md.py
ElliottKasoar: Yeah it's not the nicest.

The default files are along the lines of `NaCl-nvt-T300.0-res-10.extxyz`. If you specify a file prefix: `prefix-res-10.extxyz`, or if you specify the file stem: `stem-10.extxyz`, so yes actually I think getting the last split with `-` should always work.

tests/test_md.py
ElliottKasoar: Both, I think?

I use `tmp_path` for everything else, but since I don't specify a `file_prefix` or `restart_stem`, this tests the default restart file path, which will be created in the cwd, so that's where we want to restart from.

In other cases, e.g. if we set `file_prefix` with `tmp_pth`, the restart file should be created, and then used, in a different directory.

janus_core/calculations/md.py
ElliottKasoar: E.g. at the moment, the output is essentially:

```
│    496 │   │                                                                                     │
│    497 │   │   self.offset = 0                                                                   │
│    498 │   │   if self.restart:                                                                  │
│ ❱  499 │   │   │   self._prepare_restart()                                                       │
│    500 │   │   self.restart_files = []                                                           │
│    501 │   │   self.created_final_file = False                                                   │
│    502                 

│    525 │   │   │   try:                                                                          │
│    526 │   │   │   │   self.offset = int(last_line[0])                                           │
│    527 │   │   │   except (IndexError, ValueError) as e:                                         │
│ ❱  528 │   │   │   │   raise ValueError("Unable to read restart file") from e                    │
│    529 │   │   except FileNotFoundError as e:                                                    │
│    530 │   │   │   raise FileNotFoundError("Unable to read restart file") from e                 │
│    531      

ValueError: Unable to read restart file
```

If we get rid of the differentiation, I get something more like:

```
│    496 │   │                                                                                                                                                           │
│    497 │   │   self.offset = 0                                                                                                                                         │
│    498 │   │   if self.restart:                                                                                                                                        │
│ ❱  499 │   │   │   self._prepare_restart()                                                                                                                             │
│    500 │   │   self.restart_files = []                                                                                                                                 │
│    501 │   │   self.created_final_file = False                                                                                                                         │
│    502         

│    526 │   │   │   self.offset = int(last_line[0])                                                                                                                     │
│    527 │   │                                                                                                                                                           │
│    528 │   │   except (FileNotFoundError, IndexError, ValueError) as e:                                                                                                │
│ ❱  529 │   │   │   raise FileNotFoundError("Unable to read restart file") from e                                                                                       │
│    530 │   │                                                                                                                                                           │
│    531 │   │   if self.restart_auto:                                                                                                                                   │
│    532 │   │   │   # Attempt to infer restart file

FileNotFoundError: Unable to read restart file
```

So actually neither is particularly enlightening at the moment, but at least the current version is a bit more specific

janus_core/calculations/md.py
alinelena: ok

janus_core/calculations/md.py
oerc0122: I suppose, to me that doesn't make that much of a difference, but I can see where you're coming from. Leave as-is.

oerc0122 (approved):

tests/test_md.py
oerc0122: Alright, just wanted to make sure. Sounds good.

PR #295, opened 2024-08-29
Save system name to info from file stem

### LINKED ISSUE #293: add system name label
at the moment when we write we do not add a system name label

that will be useful especially if one wants to filter structures.
something like system_name and infer from the file name will be useful.

ElliottKasoar:
Resolves #293

### DIFF
--- a/janus_core/helpers/utils.py
+++ b/janus_core/helpers/utils.py
@@ -453,7 +453,7 @@ def output_structs(
 
     # Add label for system
     for image in images:
-        if struct_path:
+        if struct_path and "system_name" not in image.info:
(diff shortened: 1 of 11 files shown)


janus_core/helpers/utils.py
```diff
@@ -447,6 +451,11 @@ def output_structs(
             if image.calc:
                 image.info["arch"] = image.calc.parameters["arch"]
 
+    # Add label for system
+    for image in images:
+        if struct_path:
+            image.info["system_name"] = Path(struct_path).stem
```
alinelena, line 457: I will add it only if the info does not have already a system_name

alinelena (approved):

PR #297, opened 2024-09-02
Tidy code

### LINKED ISSUE #259: Tidy docs
Some changes may be able to go into #255, but it would be useful to go through the docs again and check it all makes sense with all the recent changes.

One thought so far is that it might make sense to move the [configuration files](https://stfc.github.io/janus-core/user_guide/command_line.html#using-configuration-files) section to the top, as it's somewhat out of place and hidden away at the moment.

ElliottKasoar:
Resolves #259

- Reorder and rename phonon parameters
  - Groups relevant parameters together e.g. minimize with minimize kwargs, output options at end
  - `t_step` -> `temp_step` etc. to match CLI option names, similar to changes we made for MD
- Moves explanation of config files towards top of command line docs, as otherwise it's a bit hidden away

### DIFF
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -84,6 +84,40 @@ prints the following:
       --help               Show this message and exit.
 
 
+Using configuration files
+-------------------------
+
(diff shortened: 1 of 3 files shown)


oerc0122 (approved):
Looks good. Much cleaner.

PR #300, opened 2024-09-03
Speedup tests

### LINKED ISSUE #234: Investigate slow tests
A few tests, e.g. descriptors for a trajectory, are very slow, so we should check if they can be simplified, while retaining the same utility.

ElliottKasoar:
Resolves #234

A few quick changes to simplify some of the particularly slow tests:

- `train`
  - Replace (relatively) large training data with NaCl (from relabelling output of `singlepoint` on `NaCl-traj.xyz`
  - Removes unused computations
- `eos`
  - Reduce the number of structures that require minimising
- `phonons`
  - Remove unnecessary creation of hdf5 files
  - Reduce the `supercell` for a couple of tests
  - Moves testing "full" autoband output, to reduce need to write full file in two tests
  - (Hopefully) improve the efficiency of reading the `autoband` file, which is over 300,000 lines long
    - This is a slightly less rigorous test, as it doesn't check the yaml structure, but I can't find a nice way of partially loading a yaml file, so open alternatives


tests/data/mlip_fine_tune.yml
```diff
@@ -13,13 +13,13 @@ stress_weight: 100.0
 stress_key: dft_stress
 energy_key: dft_energy
 forces_key: dft_forces
-compute_stress: True
-compute_forces: True
+compute_stress: False
+compute_forces: False
```
alinelena, line 17: i woudl keep the forces, the rest I agree... also for testing will be useful to provide some valid file rather than a fraction

tests/data/mlip_fine_tune.yml
```diff
@@ -13,13 +13,13 @@ stress_weight: 100.0
...
 clip_grad: 100
```
alinelena, line 18:
```suggestion
clip_grad: 10
```

oerc0122 (dismissed):

tests/data/mlip_fine_tune.yml
ElliottKasoar, line 17:
```suggestion
compute_forces: True
```

PR #303, opened 2024-09-04
add orb support... tricky

alinelena:


janus_core/helpers/mlip_calculators.py
```diff
@@ -217,6 +217,36 @@ def choose_calculator(
...
+                    model = orb_ff.orb_d3_v1()
+                case "orb-d3-xs-v1":
+                    model = orb_ff.orb_d3_xs_v1()
+                case "orb-d3-sm-v1":
+                    model = orb_ff.orb_d3_sm_v1()
+                case _:
+                    raise ValueError(
+                        "Please specify `model_path`, as there is no "
+                        f"default model for {arch}"
+                    )
```
oerc0122, line 242: Should be split into multiple lines, but this is the gist
```suggestion
            model = getattr(orb_ff, model_path.sub("-", "_"), None)()
            if model is None:
                    raise ValueError(
                        "Please specify `model_path`, as there is no "
                        f"default model for {arch}"
                    )
```

oerc0122 (commented):
Could be simplified. Depends if you think it's more or less readable.

janus_core/helpers/mlip_calculators.py
alinelena: would be nice but orb-mptraj does not match the pattern.

janus_core/helpers/mlip_calculators.py
```diff
@@ -217,6 +217,36 @@ def choose_calculator(
...
         calculator = SevenNetCalculator(model=model, device=device, **kwargs)
 
+    elif arch == "orb":
+        __version__ = "0.3"
+        from orb_models.forcefield.calculator import ORBCalculator
+        from orb_models.forcefield.graph_regressor import GraphRegressor
+        import orb_models.forcefield.pretrained as orb_ff
+
+        if isinstance(model_path, str):
+            match model_path:
```
ElliottKasoar, line 227: I don't think we can use match case with 3.9?

ElliottKasoar:
Mostly for reference but just to note this may [pin torch to 2.2.0](https://github.com/orbital-materials/orb-models/blob/main/pyproject.toml#L23), which at the moment is fine, but could be a problem in the future.

(I need to double check but I think even uninstalled extras' dependencies are taken into account when installing)

janus_core/helpers/mlip_calculators.py
alinelena: yes it does not but it breaks my heart to write if elif elif elif like we are in fotran 77 world... is not pyhton3.9 eol this month?

janus_core/helpers/mlip_calculators.py
oerc0122: Don't we choose what "model_path" is? You could also easily special case that. Here, it's not clear that that *is* different at first glance.

alinelena: no really I user the canonical names from their website... which i suspect is what people will expect to call them

janus_core/helpers/mlip_calculators.py
alinelena: maybe we shall catch the guilty one and then deal with it separately?

ElliottKasoar: I think actually even 3.8 isn't eol until [next month](https://devguide.python.org/versions/#supported-versions), so a little way to go until 3.9.

alinelena: my good intentions destroyed...

alinelena: in fairness they have a dict already we can use that.

pyproject.toml
```diff
@@ -45,16 +45,18 @@ chgnet = {version = "0.3.8", optional = true}
...
 ruff = "^0.5.7"
 
 [tool.poetry.extras]
-all = ["alignn", "chgnet", "matgl", "dgl", "torchdata", "sevenn", "torch_geometric"]
+all = ["alignn", "chgnet", "matgl", "dgl", "torchdata", "sevenn", "torch_geometric", "orb"]
 alignn = ["alignn"]
 chgnet = ["chgnet"]
 m3gnet = ["matgl", "dgl", "torchdata"]
 sevennet = ["sevenn", "torch_geometric"]
+orb = ["orb_models"]
```
ElliottKasoar, line 59: Do we not need to add [Pynanoflann](https://github.com/orbital-materials/orb-models?tab=readme-ov-file#install) too?

ElliottKasoar:
I think the installation is all working now with `uv`, but since I don't think we're dropping 3.9 quite yet, the `match` statement still needs changing.

janus_core/helpers/mlip_calculators.py
ElliottKasoar: (We can keep assuming we merge #377)

janus_core/helpers/mlip_calculators.py
```diff
@@ -224,6 +224,36 @@ def choose_calculator(
...
+                case "orb-d3-xs-v2":
+                    model = orb_ff.orb_d3_xs_v2()
+                case "orb-d3-sm-v2":
+                    model = orb_ff.orb_d3_sm_v2()
+                case _:
+                    raise ValueError(
+                        "Please specify `model_path`, as there is no "
+                        f"default model for {arch}"
+                    )
+        elif isinstance(model_path, GraphRegressor):
```
ElliottKasoar, line 250: I think match case lets you do type checks like this too, doesn't it? Is there a nice way to combine it with the specific string checks?

ElliottKasoar (commented):
I think we want a test for the calculator init e.g. parameters are set, and (invalid) model path works as expected, but generally looks good!

janus_core/helpers/mlip_calculators.py
alinelena: this one is different buy maybe we can import it as the write pattern and then use your abstraction?

case "orb-mptraj-only-v1":
                    model = orb_ff.orb_v1_mptraj_only()

janus_core/helpers/mlip_calculators.py
alinelena: see comment above to jacob's abstraction

janus_core/helpers/mlip_calculators.py
```diff
@@ -224,6 +224,26 @@ def choose_calculator(
...
+        from orb_models.forcefield.calculator import ORBCalculator
+        from orb_models.forcefield.graph_regressor import GraphRegressor
+        import orb_models.forcefield.pretrained as orb_ff
+        from  orb_models.forcefield.pretrained import orb_v2_mptraj_only as orb_mptraj_only_v2
+
+        model = getattr(orb_ff, model_path.sub("-", "_"), None)()
+        if model is None:
+                    raise ValueError(
+                        "Please specify `model_path`, as there is no "
+                        f"default model for {arch}"
```
ElliottKasoar, line 238: Is there no obvious default that makes sense, like we have for say mace_mp?

ElliottKasoar: Ah  there is `model = orb_ff.orb_v2()` below, but when should that be reached? If you don't pass anything, just above it raises `AttributeError: 'NoneType' object has no attribute 'sub'`

ElliottKasoar:
Rebased, and updated docs. Setting the `model_path` needs fixing, and then a couple more tests to be added.

alinelena:
> Rebased, and updated docs. Setting the `model_path` needs fixing, and then a couple more tests to be added.
> 
> Then the MacOS `dm-tree` error needs fixing.

i wonder if we shall to a try catch... and then raise accordingly?

ElliottKasoar:
It's possible there are slightly neater versions for setting the model, but I think it now uses the default model correctly, as well as using the pretrained model labels, or a loaded model (which is tested with their XS checkpoint file).

Also installed cmake on our Mac, so hopefully those tests are fixed.

ElliottKasoar (approved):
I can't officially approve, but I think it looks good

harveydevereux (approved):
Also looks good to me

PR #304, opened 2024-09-04
add nequip support fixes #223

### LINKED ISSUE #223: Add support for nequip calculations

alinelena:


pyproject.toml
```diff
@@ -48,13 +48,15 @@ sevenn = { version = "0.9.3", optional = true }
 torchdata = {version = "0.7.1", optional = true} # Pin due to dgl issue
 torch_geometric = { version = "^2.5.3", optional = true }
 ruff = "^0.5.7"
+nequip = {version = "^0.6.1", optional = true }
```
ElliottKasoar, line 51:
```suggestion
nequip = {version = "0.6.1", optional = true }
```
Probably best to pin like the other MLIPs to avoid API/dependency breaking changes?

tests/test_mlip_calculators.py
ElliottKasoar, line 1: What happens if we give it a bad path?

janus_core/helpers/mlip_calculators.py
```diff
@@ -224,6 +224,21 @@ def choose_calculator(
         kwargs.setdefault("sevennet_config", None)
         calculator = SevenNetCalculator(model=model_path, device=device, **kwargs)
 
+    elif arch == "nequip":
+        from nequip import __version__
+        from nequip.ase import NequIPCalculator
+
+        if isinstance(model_path, Path):
+            model_path = str(model_path)
```
ElliottKasoar, line 232: Should we check it exists as a file? Or does nequip handle it nicely enough?

alinelena: we get 
```bash
ValueError: Please specify `model_path`, as there is no default model for nequip

```

so we are fine

tests/test_mlip_calculators.py
alinelena: raises value error... w

alinelena: added test for it I hope

tests/test_single_point.py
```diff
@@ -279,20 +280,30 @@ def test_mlips(arch, device, expected_energy):
...
+    ("sevennet", "cpu", -27.061979293823242, "NaCl.cif", {"model_path": SEVENNET_PATH}),
+    ("sevennet", "cpu", -27.061979293823242, "NaCl.cif", {}),
+    (
+        "sevennet",
+        "cpu",
+        -27.061979293823242,
+        "NaCl.cif",
+        {"model_path": "SevenNet-0_11July2024"},
+    ),
+    ("nequip", "cpu", -169815.1282456301, "toluene.xyz", {"model_path": NEQUIP_PATH}),
```
ElliottKasoar, line 295: Out of interest, is this a remotely sensible energy?

ElliottKasoar (approved):
Looks good, thanks!

ElliottKasoar (changes requested):
Ah sorry missed my checklist, can we update:

- [x] Update README (dependencies, features)

janus_core/helpers/mlip_calculators.py
```diff
@@ -224,6 +224,21 @@ def choose_calculator(
...
+        else:
+            raise ValueError(
```
ElliottKasoar, line 234: See: https://github.com/stfc/janus-core/pull/428#discussion_r1957949224

```
janus singlepoint --arch nequip --struct tests/data/NaCl.cif --model-path ./invalid/path

ValueError: Please specify `model_path`, as there is no default model for nequip
```

alinelena: your solution was very good since indeed are two different issues.

tests/test_single_point.py
alinelena: yes, they do not have pseudos so energy is not scaled... although I will not use that potential for anything serious.

ElliottKasoar (approved):

tests/test_single_point.py
alinelena: good thing is I have a training example for nequip... I wonder if we can use it for multi training?

PR #307, opened 2024-09-05
Add mesh CLI option and fix thermal mesh

### LINKED ISSUE #305: mesh is missing as an option for phonons
mesh, that is passes to run mesh shall be a cli option, also is missing for run_thermal

ElliottKasoar:
Starting point to resolves #305

### DIFF
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -60,6 +60,8 @@ class Phonons(BaseCalculation):
         Size of supercell for calculation. Default is 2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
+    mesh : MaybeSequence[float]
+        Mesh for sampling. Default is (10, 10, 10).
     symmetrize : bool
         Whether to symmetrize force constants after calculation.
         Default is False.
@@ -138,6 +140,7 @@ class Phonons(BaseCalculation):
         calcs: MaybeSequence[PhononCalcs] = (),
         supercell: MaybeList[int] = 2,
         displacement: float = 0.01,
+        mesh: MaybeSequence[float] = (10, 10, 10),
         symmetrize: bool = False,
         minimize: bool = False,
         minimize_kwargs: Optional[dict[str, Any]] = None,
@@ -186,6 +189,8 @@ class Phonons(BaseCalculation):
             Size of supercell for calculation. Default is 2.
         displacement : float
             Displacement for force constants calculation, in A. Default is 0.01.
+        mesh : MaybeSequence[float]
+            Mesh for sampling. Default is (10, 10, 10).
         symmetrize : bool
             Whether to symmetrize force constants after calculations.
             Default is False.
@@ -219,6 +224,7 @@ class Phonons(BaseCalculation):
 
         self.calcs = calcs
         self.displacement = displacement
+        self.mesh = mesh
         self.symmetrize = symmetrize
         self.minimize = minimize
         self.minimize_kwargs = minimize_kwargs
@@ -490,13 +496,18 @@ class Phonons(BaseCalculation):
             bplt.savefig(plot_file)
 
     def calc_thermal_props(
-        self, write_thermal: Optional[bool] = None, **kwargs
+        self,
+        mesh: Optional[MaybeSequence[float]] = None,
+        write_thermal: Optional[bool] = None,
+        **kwargs,
     ) -> None:
         """
         Calculate thermal properties and optionally write results.
 
         Parameters
         ----------
+        mesh : Optional[MaybeSequence[float]]
+            Mesh for sampling. Default is self.mesh.
         write_thermal : Optional[bool]
             Whether to write out thermal properties to file. Default is
             self.write_results.
@@ -505,6 +516,9 @@ class Phonons(BaseCalculation):
         """
         if write_thermal is None:
             write_thermal = self.write_results
+
+        if mesh is None:
+            mesh = self.mesh
 
         # Calculate phonons if not already in results
         if "phonon" not in self.results:
@@ -515,7 +529,7 @@ class Phonons(BaseCalculation):
             self.logger.info("Starting thermal properties calculation")
             self.tracker.start_task("Thermal calculation")
 
-        self.results["phonon"].run_mesh()
+        self.results["phonon"].run_mesh(mesh)
         self.results["phonon"].run_thermal_properties(
             t_step=self.temp_step, t_max=self.temp_max, t_min=self.temp_min
         )
@@ -563,7 +577,7 @@ class Phonons(BaseCalculation):
     def calc_dos(
         self,
         *,
-        mesh: MaybeList[float] = (10, 10, 10),
+        mesh: Optional[MaybeSequence[float]] = None,
         write_dos: Optional[bool] = None,
         **kwargs,
     ) -> None:
@@ -572,8 +586,8 @@ class Phonons(BaseCalculation):
 
         Parameters
         ----------
-        mesh : MaybeList[float]
-            Mesh for sampling. Default is (10, 10, 10).
+        mesh : Optional[MaybeSequence[float]]
+            Mesh for sampling. Default is self.mesh.
         write_dos : Optional[bool]
             Whether to write out results to file. Default is True.
         **kwargs
@@ -581,6 +595,9 @@ class Phonons(BaseCalculation):
         """
         if write_dos is None:
             write_dos = self.write_results
+
+        if mesh is None:
+            mesh = self.mesh
 
         # Calculate phonons if not already in results
         if "phonon" not in self.results:
@@ -665,7 +682,7 @@ class Phonons(BaseCalculation):
     def calc_pdos(
         self,
         *,
-        mesh: MaybeList[float] = (10, 10, 10),
+        mesh: Optional[MaybeSequence[float]] = None,
         write_pdos: Optional[bool] = None,
         **kwargs,
     ) -> None:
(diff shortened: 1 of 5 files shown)


janus_core/cli/phonons.py
```diff
@@ -138,6 +127,8 @@ def phonons(
         2x2x2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
+    mesh : tuple[float, float, float]
+        Mesh for sampling. Default is (10, 10, 10).
```
alinelena, line 131: why not the same as the supercell? for the specification

ElliottKasoar: When I implemented the supercell, I don't think I realised you can do tuples, so it initially looked like it looked like `-supercell x --supercell y --supercell z`, which is how you have to pass lists, and is obviously quite messy, and led to @oerc0122's suggestion of the current form.

With a tuple, we can do `--mesh x y z`, which seems a lot neater to me.

We do lose the flexibility of a single integer alternative, but I'd probsbly lean towards changing `supercell` to match this, rather than than the other way around.

oerc0122: Does the mesh want to be `float`s or `int`s?

alinelena: ints

alinelena: also is fine with me as long supercelll and mesh are consistent

janus_core/calculations/phonons.py
```diff
@@ -138,6 +140,7 @@ def __init__(
         calcs: MaybeSequence[PhononCalcs] = (),
         supercell: MaybeList[int] = 2,
         displacement: float = 0.01,
+        mesh: MaybeSequence[float] = (10, 10, 10),
```
alinelena, line 143:
```suggestion
        mesh: MaybeSequence[int] = (10, 10, 10),
```

janus_core/calculations/phonons.py
```diff
@@ -186,6 +189,8 @@ def __init__(
             Size of supercell for calculation. Default is 2.
         displacement : float
             Displacement for force constants calculation, in A. Default is 0.01.
+        mesh : MaybeSequence[float]
```
alinelena, line 192:
```suggestion
        mesh : MaybeSequence[int]
```

janus_core/calculations/phonons.py
```diff
@@ -490,13 +496,18 @@ def write_bands(
             bplt.savefig(plot_file)
 
     def calc_thermal_props(
-        self, write_thermal: Optional[bool] = None, **kwargs
+        self,
+        mesh: Optional[MaybeSequence[float]] = None,
```
alinelena, line 500:
```suggestion
        mesh: Optional[MaybeSequence[int]] = None,
```

janus_core/calculations/phonons.py
```diff
@@ -490,13 +496,18 @@ def write_bands(
...
+        write_thermal: Optional[bool] = None,
+        **kwargs,
     ) -> None:
         """
         Calculate thermal properties and optionally write results.
 
         Parameters
         ----------
+        mesh : Optional[MaybeSequence[float]]
```
alinelena, line 509:
```suggestion
        mesh : Optional[MaybeSequence[int]]
```

janus_core/calculations/phonons.py
```diff
@@ -563,7 +577,7 @@ def write_thermal_props(self, thermal_file: Optional[PathLike] = None) -> None:
     def calc_dos(
         self,
         *,
-        mesh: MaybeList[float] = (10, 10, 10),
+        mesh: Optional[MaybeSequence[float]] = None,
```
alinelena, line 580:
```suggestion
        mesh: Optional[MaybeSequence[int]] = None,
```

janus_core/calculations/phonons.py
```diff
@@ -572,8 +586,8 @@ def calc_dos(
 
         Parameters
         ----------
-        mesh : MaybeList[float]
-            Mesh for sampling. Default is (10, 10, 10).
+        mesh : Optional[MaybeSequence[float]]
```
alinelena, line 589:
```suggestion
        mesh : Optional[MaybeSequence[int]]
```

janus_core/calculations/phonons.py
```diff
@@ -665,7 +682,7 @@ def write_dos(
     def calc_pdos(
         self,
         *,
-        mesh: MaybeList[float] = (10, 10, 10),
+        mesh: Optional[MaybeSequence[float]] = None,
```
alinelena, line 685:
```suggestion
        mesh: Optional[MaybeSequence[int]] = None,
```

janus_core/calculations/phonons.py
```diff
@@ -674,8 +691,8 @@ def calc_pdos(
 
         Parameters
         ----------
-        mesh : MaybeList[float]
-            Mesh for sampling. Default is (10, 10, 10).
+        mesh : Optional[MaybeSequence[float]]
```
alinelena, line 694:
```suggestion
        mesh : Optional[MaybeSequence[int]]
```

janus_core/cli/phonons.py
```diff
@@ -39,30 +40,22 @@ def phonons(
...
-        Option(help="Supercell lattice vectors in the form '1x2x3'."),
+        str, Option(help="Supercell lattice vectors in the form '1x2x3'.")
     ] = "2x2x2",
     displacement: Annotated[
-        float,
-        Option(help="Displacement for force constants calculation, in A."),
+        float, Option(help="Displacement for force constants calculation, in A.")
     ] = 0.01,
+    mesh: Annotated[
+        tuple[float, float, float], Option(help="Mesh numbers along a, b, c axes.")
```
alinelena, line 49:
```suggestion
        tuple[int, int, int], Option(help="Mesh numbers along a, b, c axes.")
```

janus_core/cli/phonons.py
```diff
@@ -138,6 +127,8 @@ def phonons(
         2x2x2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
+    mesh : tuple[float, float, float]
```
oerc0122, line 130:
```suggestion
    mesh : tuple[int, int, int]
```

oerc0122 (changes requested):
Once those `float`s are `int`s, I'm happy for this to go in.

janus_core/calculations/phonons.py
```diff
@@ -60,6 +60,8 @@ class Phonons(BaseCalculation):
         Size of supercell for calculation. Default is 2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
+    mesh : MaybeSequence[float]
```
alinelena, line 63:
```suggestion
    mesh : MaybeSequence[int]
```

janus_core/calculations/phonons.py
```diff
@@ -60,6 +60,8 @@ class Phonons(BaseCalculation):
         Size of supercell for calculation. Default is 2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
+    mesh : MaybeSequence[int]
```
oerc0122, line 63: Is this actually a `MaybeSequence` (i.e. `int | Sequence[int]`). Is it not `int | tuple[int, int, int]`?

oerc0122: In fact, b definition I think it's just `tuple[int, int, int]`

ElliottKasoar: Why do you say that?

The type hinting for `set_mesh` is an `intc` array-like with shape `(3,)`: https://github.com/phonopy/phonopy/blob/develop/phonopy/api_phonopy.py#L1871, but the defaults for [run_mesh](https://github.com/phonopy/phonopy/blob/develop/phonopy/api_phonopy.py#L1833) and [init_mesh](https://github.com/phonopy/phonopy/blob/develop/phonopy/api_phonopy.py#L1723) (which are called by `set_mesh`), as well as other functions where it's set e.g. [auto_total_dos](https://github.com/phonopy/phonopy/blob/develop/phonopy/api_phonopy.py#L2323) etc. are generally a single floating point value (`100.0`), since

> When float value is given, uniform mesh is generated.

That's also why I had it as a float, rather than int, since floats are explicitly allowed.

alinelena: unfortunately phonopy api is not best example of how to design an interface. 
for janus i would say we shall restrict to an explicit mesh array.
float mesh is not only generating the grid but also sets the gamma centre to true... the only suggestion I have, to be done in this or separate is to have mesh-kwargs, so we can pass the rest of the arguments to mesh.

in general as part of this i went through it and i have a list of suggestions for phonons in general. @oerc0122 feedback encouraged.
[ ] symmetrize also structure if force constants are symmetrized
[ ] dos-kwargs, allow passing freq_min, max and pitch or explicit params? implications are as dos-kwargs, we concentrate only on dos, as explicit we shall apply min,max to bands plotting too.
[ ] add path passing
[ ] enable compression for bands
[ ] remove autoband, and replace it with band, natural if we add paths
[ ] write thermal props using phonopy api. so we can reuse output for other tools.

some proof of concept in here https://gitlab.com/drFaustroll/lavello/-/blob/main/phonons.py?ref_type=heads

alinelena (approved):

ElliottKasoar:
(Draft until decided on approach in sync with #320)

alinelena:
I will not draft it.. but merge it and 320 rebased  based on it.

ElliottKasoar:
I agree 320 needs to be rebased, I mostly just want to be sure we want to change the type here.

Feels silly to change the type, then immediately change it again, but yeah I think a tuple either just for a diagonal supercell, or a custom Union of tuples, is looking like the way to go, so this should be a sensible starting point.

PR #317, opened 2024-10-02
Add final energy to geomopt log

### LINKED ISSUE #315: Save the final energy in geometry optimisation to log/summary
Hi!

In my workflow using Janus CLI, I would like to access the final energy of the geometry optimisation for further processing. However, I wasn't able to find this value in either the `geomopt-summary.yml` file or the `geomopt-log.yml` file (I only found the final force). Would it be possible to save the final energy at the end of the geometry optimisation somewhere?

Thank you!

ElliottKasoar:
May resolve #315

### DIFF
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -301,6 +301,7 @@ class GeomOpt(BaseCalculation):
         if self.logger:
             self.logger.info("After optimization spacegroup: %s", s_grp)
             self.logger.info("Max force: %.6f", max_force)
+            self.logger.info("Final energy: %.6f", self.struct.get_potential_energy())
 
         if not converged:
             warnings.warn(
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -300,8 +300,8 @@ class GeomOpt(BaseCalculation):
 
         if self.logger:
             self.logger.info("After optimization spacegroup: %s", s_grp)
-            self.logger.info("Max force: %.6f", max_force)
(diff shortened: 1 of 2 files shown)


janus_core/calculations/geom_opt.py
```diff
@@ -301,6 +301,7 @@ def run(self) -> None:
         if self.logger:
             self.logger.info("After optimization spacegroup: %s", s_grp)
             self.logger.info("Max force: %.6f", max_force)
+            self.logger.info("Final energy: %.6f", self.struct.get_potential_energy())
```
oerc0122, line 304: Just to double check this is "guaranteed" to not start a new calculation.

alinelena: also be sure you do not use a format one may want the precise energy

ElliottKasoar: > Just to double check this is "guaranteed" to not start a new calculation.

As long as energy is in the results dictionary (which it should be at this point), I believe so, from this function: https://gitlab.com/ase/ase/-/blob/master/ase/calculators/calculator.py?ref_type=heads#L515



> also be sure you do not use a format one may want the precise energy

Are you saying we don't want `.6f`?

alinelena: yes no format in summary

ElliottKasoar: This is in the log rather than the summary, so I think it needs some kind of formatting. We could do `%s` which I think basically keeps the full precision?

Would we want the same for the max force?

alinelena: if you are at it yes do the max force... yap whatever you like.

alinelena (approved):

PR #318, opened 2024-10-02
Add lags to vafs

### LINKED ISSUE #253: post-process issues
- [x] Mismatched argument names - filename, filenames
- [ ] Data not returned on vafs
- [x] `str` is `Sequence` so passing string as filename does not work.

oerc0122:
Fixes #253

### DIFF
--- a/janus_core/processing/post_process.py
+++ b/janus_core/processing/post_process.py
@@ -191,7 +191,7 @@ def compute_vaf(
     index: SliceLike = (0, None, 1),
     filter_atoms: MaybeSequence[MaybeSequence[Optional[int]]] = ((None),),
     time_step: float = 1.0,
-) -> NDArray[float64]:
+) -> tuple[NDArray[float64], list[NDArray[float64]]]:
     """
     Compute the velocity autocorrelation function (VAF) of `data`.
 
@@ -219,8 +219,30 @@ def compute_vaf(
 
     Returns
     -------
-    MaybeSequence[NDArray[float64]]
+    lags : numpy.ndarray
+        Lags at which the VAFs have been computed.
+    vafs : list[numpy.ndarray]
         Computed VAF(s).
+
+    Notes
+    -----
+    `filter_atoms` is given as a series of sequences of atoms, where
+    each element in the series denotes a VAF subset to calculate and
+    each sequence determines the atoms (by index) to be included in that VAF.
+
+    E.g.
+
+    .. code-block: Python
+
+        # Species indices in cell
+        na = (1, 3, 5, 7)
+        cl = (2, 4, 6, 8)
+
+        compute_vaf(..., filter_atoms=(na, cl))
+
+    Would compute separate VAFs for each species.
+
+    By default, one VAF will be computed for all atoms in the structure.
     """
     # Ensure if passed scalars they are turned into correct dimensionality
     if not isinstance(filter_atoms, Sequence):
--- a/janus_core/processing/post_process.py
+++ b/janus_core/processing/post_process.py
@@ -1,5 +1,5 @@
 """Module for post-processing trajectories."""
-1;4000;35c
+
 from collections.abc import Sequence
 from itertools import combinations_with_replacement
 from typing import Optional, Union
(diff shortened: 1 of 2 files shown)


janus_core/helpers/post_process.py
```diff
@@ -270,17 +274,24 @@ def compute_vaf(
...
+    lags = np.arange(n_steps) * time_step
+
     if fft:
         vafs = np.fft.fft(vafs, axis=0)
-
-    vafs = [
-        np.average([vafs[used_atoms[i]] for i in atoms], axis=0)
-        for atoms in filter_atoms
-    ]
+        lags = 1. / lags
```
oerc0122, line 281: Does this need to be 
```suggestion
        lags = 1. / reversed(lags)
```

janus_core/helpers/post_process.py
```diff
@@ -1,5 +1,5 @@
 """Module for post-processing trajectories."""
-
+1;4000;35c
```
ElliottKasoar, line 2:
```suggestion
```
I assume this is unintended?

ElliottKasoar:
Are these lags consistent with the on-the-fly correlations (@harveydevereux)?

harveydevereux:
> Are these lags consistent with the on-the-fly correlations (@harveydevereux)?

Yes these will be, given the right on-the-fly options. But it is possible to have correlations returned with lags that don't have a uniform step due to the window averaging, sample frequency options etc.
```python
lags = np.arange(n_steps) * time_step
```

Currently don't have fft as an option though, perhaps we should

janus_core/helpers/post_process.py
```diff
@@ -190,6 +190,7 @@ def compute_vaf(
     fft: bool = False,
     index: SliceLike = (0, None, 1),
     filter_atoms: MaybeSequence[MaybeSequence[Optional[int]]] = ((None),),
+    time_step: float = 1.0,
 ) -> NDArray[float64]:
```
ElliottKasoar, line 194: Not entirely new, but this is inconsistent with the corresponding docstring

```
    Returns
    -------
    MaybeSequence[NDArray[float64]]
        Computed VAF(s).
```

I think both need updating now that `vafs` is a tuple.

janus_core/helpers/post_process.py
```diff
@@ -212,6 +213,9 @@ def compute_vaf(
     filter_atoms : MaybeSequence[MaybeSequence[Optional[int]]]
         Compute the VAF averaged over subsets of the system.
         Default is all atoms.
```
ElliottKasoar, line 215: This isn't newly introduced, but to me this isn't entirely clear how this is used (why could be be a sequence of a sequence?)

I also wonder if 

```
    filter_atoms = [
        atom if atom[0] is not None else range(n_atoms) for atom in filter_atoms
    ]
```

etc. is as clear as it could be that `atom` is generally referring to an index, rather than `Atom` objects as if often the case for similar code elsewhere.

janus_core/helpers/post_process.py
ElliottKasoar: Presumably the vafs would also need to be reversed, so it's probably fine without?

ElliottKasoar (changes requested):
I think `test_post_process.py::test_vaf` needs updating to get the vaf from the returned tuple, and ideally probably should check the lag while it's at it too.

janus_core/helpers/post_process.py
ElliottKasoar: Presumably the vafs would also need to be reversed, so it's probably fine without?

janus_core/helpers/post_process.py
harveydevereux: On ```MaybeSequence[MaybeSequence[Optional[int]]]``` I think the idea is to be able to do 

```python
   vaf_post = post_process.compute_vaf(
        traj, use_velocities=True, filter_atoms=(na, cl)
    )
```

where ```(na, cl) === ( (0, 2, 4, ...), (1, 3, 5, ...) )```, which is a sequence of sequence. That way you get the partial VAF's for Na and Cl at once.

The one liner could use some thought, I wasn't sure what to do last pr.

Incidentally I have tested rebasing #285 onto this and all is fine (just will need to account for the new lags in the return type).

ElliottKasoar (approved):
Looks good. Thanks for the notes on `filter_atoms`!

PR #316, opened 2024-10-02
Allow multiple emissions trackers

ElliottKasoar:
Resolves #314

PR #319, opened 2024-10-02
Add progress bar to phonon calculations

### LINKED ISSUE #268: Visual progress information during phonon calculations
Phonon calculations can take a while to run and currently Janus provides no visual feedback so it is unclear if the calculation is progressing well or if the user should consider killing it and choosing e.g. a smaller supercell.

A `tqdm` progress bar would be lovely (and provides time estimates) but without that there could be a line like "Calculating displacement 3/48..." which can be updated using the "\r" escape character.

RastislavTuranyi:
Closes #268 

Adds a progress bar to phonon calculations using `tqdm`, which is already a dependency. A new argument was added to `Phonons.__init__` to control the presence of this feature, disabled by default to minimise impact on code in general. However, the progress bar is always enabled for CLI.

One issue that I and @ajjackson have noted is that the bar only shows the progress over displacements, meaning that when bands, DOS, or PDOS are chosen to be calculated, the 100% bar just continues sitting there despite the prolonged calculations. This could be confusing to users. One solution could be writing to stdout before/after the progress bar appears, saying what the bar is for, but we weren't sure what the best path forward is. What do you think?

Additionally, I should probably mention that I ran into #316 while working on this despite only running janus only in serial.

### DIFF
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -216,6 +216,8 @@ class Phonons(BaseCalculation):
         file_prefix : Optional[PathLike]
             Prefix for output filenames. Default is inferred from structure name, or
             chemical formula of the structure.
+        enable_progress_bar : bool
+            Whether to show a progress bar during phonon calculations. Default is False.
         """
         (read_kwargs, minimize_kwargs) = none_to_dict((read_kwargs, minimize_kwargs))
 
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -867,12 +867,23 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
     Iterable
         An iterable of the values in the sequence.
     """
+    from rich.style import Style
+
     text_column = TextColumn("{task.description}")
-    bar_column = BarColumn(bar_width=None)
+    bar_column = BarColumn(
+        bar_width=None,
+        complete_style=Style(color="#FBBB10"),
+        finished_style=Style(color="#E38408"),
+    )
     completion_column = MofNCompleteColumn()
     time_column = TimeRemainingColumn()
     progress = Progress(
-        text_column, bar_column, completion_column, time_column, expand=True
+        text_column,
+        bar_column,
+        completion_column,
+        time_column,
+        expand=True,
+        auto_refresh=False,
     )
 
     with progress:
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -1,14 +1,20 @@
 """Phonon calculations."""
 
-from collections.abc import Sequence
-from typing import Any, Optional, get_args
+from collections.abc import Iterable, Sequence
+from typing import Any, Optional, Union, get_args
 
 from ase import Atoms
 from numpy import ndarray
 import phonopy
 from phonopy.file_IO import write_force_constants_to_hdf5
 from phonopy.structure.atoms import PhonopyAtoms
-from tqdm import tqdm
+from rich.progress import (
+    BarColumn,
+    MofNCompleteColumn,
+    Progress,
+    TextColumn,
+    TimeRemainingColumn,
+)
 
 from janus_core.calculations.base import BaseCalculation
 from janus_core.calculations.geom_opt import GeomOpt
@@ -89,6 +95,8 @@ class Phonons(BaseCalculation):
     file_prefix : Optional[PathLike]
         Prefix for output filenames. Default is inferred from chemical formula of the
         structure.
+    enable_progress_bar : bool
+        Whether to show a progress bar during phonon calculations. Default is False.
 
     Attributes
     ----------
@@ -150,7 +158,7 @@ class Phonons(BaseCalculation):
         write_results: bool = True,
         write_full: bool = True,
         file_prefix: Optional[PathLike] = None,
-        enable_progress_bar: bool = False
+        enable_progress_bar: bool = False,
     ) -> None:
         """
         Initialise Phonons class.
@@ -363,7 +371,9 @@ class Phonons(BaseCalculation):
         disp_supercells = phonon.supercells_with_displacements
 
         if self.enable_progress_bar:
-            disp_supercells = tqdm(disp_supercells)
+            disp_supercells = track_progress(
+                disp_supercells, "Computing displacements..."
+            )
 
         phonon.forces = [
             self._calc_forces(supercell)
@@ -835,3 +845,35 @@ class Phonons(BaseCalculation):
             self.calc_dos(plot_bands="bands" in self.calcs)
         if "pdos" in self.calcs:
             self.calc_pdos()
+
+
+def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Iterable:
+    """
+    Track the progress of iterating over a sequence.
+
+    This is done by displaying a progress bar in the console using the rich library.
+    The function is an iterator over the sequence, updating the progress bar each
+    iteration.
+
+    Parameters
+    ----------
+    sequence : Iterable
+        The sequence to iterate over. Must support "len".
+    description : str
+        The text to display to the left of the progress bar.
+
+    Returns
+    -------
+    Iterable
+        An iterable of the values in the sequence.
+    """
+    text_column = TextColumn("{task.description}")
+    bar_column = BarColumn(bar_width=None)
+    completion_column = MofNCompleteColumn()
+    time_column = TimeRemainingColumn()
+    progress = Progress(
+        text_column, bar_column, completion_column, time_column, expand=True
+    )
+
+    with progress:
+        yield from progress.track(sequence, description=description)
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -8,6 +8,7 @@ from numpy import ndarray
 import phonopy
 from phonopy.file_IO import write_force_constants_to_hdf5
 from phonopy.structure.atoms import PhonopyAtoms
+from tqdm import tqdm
 
 from janus_core.calculations.base import BaseCalculation
 from janus_core.calculations.geom_opt import GeomOpt
@@ -149,6 +150,7 @@ class Phonons(BaseCalculation):
         write_results: bool = True,
         write_full: bool = True,
         file_prefix: Optional[PathLike] = None,
+        enable_progress_bar: bool = False
     ) -> None:
         """
         Initialise Phonons class.
@@ -229,6 +231,7 @@ class Phonons(BaseCalculation):
         self.plot_to_file = plot_to_file
         self.write_results = write_results
         self.write_full = write_full
+        self.enable_progress_bar = enable_progress_bar
 
         # Ensure supercell is a valid list
         self.supercell = [supercell] * 3 if isinstance(supercell, int) else supercell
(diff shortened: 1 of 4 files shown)


janus_core/calculations/phonons.py
```diff
@@ -8,6 +8,7 @@
 import phonopy
 from phonopy.file_IO import write_force_constants_to_hdf5
 from phonopy.structure.atoms import PhonopyAtoms
+from tqdm import tqdm
```
oerc0122, line 11: `tqdm` may be a dependency of a dependency and that's not a good thing to rely on. You need to add this explicitly to dependencies if you want to use it.

oerc0122: Also, `rich` is also a dependency of a dependency, that *may* be a better choice here since Jupyter notebooks and IPython are encouraged ways of using Janus

ElliottKasoar: Do you know how rich compares in terms of performance/overheads?

oerc0122: I have no idea, but I should think minimal in relation to a phonon calculation.

ElliottKasoar:
Thanks for this!

> One issue that I and @ajjackson have noted is that the bar only shows the progress over displacements, meaning that when bands, DOS, or PDOS are chosen to be calculated, the 100% bar just continues sitting there despite the prolonged calculations. This could be confusing to users. One solution could be writing to stdout before/after the progress bar appears, saying what the bar is for, but we weren't sure what the best path forward is. What do you think?

I agree it's a bit confusing, so I'd say yes we probably do want some kind of output labelling the progress bar.
 
> Additionally, I should probably mention that I ran into #316 while working on this despite only running janus only in serial.

Thanks for the report! If you still encounter this after updating with the latest fixes, please open a new issue.

janus_core/cli/phonons.py
```diff
@@ -258,6 +258,7 @@ def phonons(
         "write_results": True,
         "write_full": write_full,
         "file_prefix": file_prefix,
+        "enable_progress_bar": True
```
ElliottKasoar, line 261:
```suggestion
        "enable_progress_bar": True,
```
To make ruff happy (best merged with other suggestion)

janus_core/calculations/phonons.py
```diff
@@ -149,6 +150,7 @@ def __init__(
         write_results: bool = True,
         write_full: bool = True,
         file_prefix: Optional[PathLike] = None,
+        enable_progress_bar: bool = False
```
ElliottKasoar, line 153:
```suggestion
        enable_progress_bar: bool = False,
```
To make ruff happy (best merged with other suggestion)

janus_core/calculations/phonons.py
RastislavTuranyi: [Supposedly](https://github.com/Textualize/rich/issues/176#issuecomment-663995614), they should be pretty comparable nowadays. Regardless, as mentioned, the milliseconds are going to be negligible.

oerc0122:
> I agree it's a bit confusing, so I'd say yes we probably do want some kind of output labelling the progress bar.

`tqdm` allows you to pass a label.
> desc: str, optional
> Prefix for the progressbar.

-- [tqdm docs](https://tqdm.github.io/docs/tqdm/)

RastislavTuranyi:
I am looking into `rich` at the moment, as per suggestion, and it can also have a label. The progress bar can also be made to disappear at the end, so I think a text could be displayed in its place to indicate that further calculations may be ongoing. Unfortunately, since the bands/DOS/etc. calculations are handled entirely by Phonopy, I don't think there is a way to display a progress bar there, though we could maybe do like a spinner or something of that sort, if desired.

janus_core/calculations/phonons.py
ElliottKasoar: I think in the context of issues linked in #272, some of the slowness of typer seemed to be due to rich imports on timescales noticeable by users, but I'm happy with which ever seems nicer, as long as performance seems ok

RastislavTuranyi: Oof, I don't know about imports, to be honest. `tqdm` is supposed to be leightweight and has no dependencies, so it might be better in that regard, but then again this implementation is in `calculations`, so as long as the import of the `calculations.phonons` happens after the CLI is loaded (which is what that PR proposes, if I'm not wrong), then it shouldn't make a difference, right?

oerc0122: If `rich` is already imported due to other imports, it will already be in the loaded modules and thus free to import.

ElliottKasoar: > Oof, I don't know about imports, to be honest. `tqdm` is supposed to be leightweight and has no dependencies, so it might be better in that regard, but then again this implementation is in `calculations`, so as long as the import of the `calculations.phonons` happens after the CLI is loaded (which is what that PR proposes, if I'm not wrong), then it shouldn't make a difference, right?

Sorry, I meant just in terms of it potentially having a bigger impact than you might expect when running calculations.

> If `rich` is already imported due to other imports, it will already be in the loaded modules and thus free to import.

True, I think it will imported if you're using the CLI, but not necessarily if you're using the Python interface.

I think it should be fine, if it is nicer.

RastislavTuranyi: Personally, I prefer the `tqdm` default progress bar, though `rich` appears to be more customisable

oerc0122: That's all fine if everyone's happy. The main thing was that it was an option which is why (although not necessarily obvious in the GitHub font) "may" was italicised.

oerc0122: > On the other hand, it does appear to me that `tqdm` integration with jupyter notebook could be non-trivial

Don't worry the integration is that it prints to the output perfectly happily.

RastislavTuranyi: So, to confirm, we are sticking with `tqdm`, is that right?

ElliottKasoar: I obviously haven't seen how this specific one would look in comparison, but in general `rich` maybe looks slightly nicer to me, and maybe fits stylistically with the typer interface better.

So, all things being equal (e.g. no immediately noticeable performance issues, which it sounds like there shouldn't be), I'd maybe tend towards `rich`.

I am completely fine with `tqdm` if it's more straightforward to finish this PR up though, as long as it addresses the issue.

RastislavTuranyi: This is how the different implementations look on my linux VM:

`tqdm`:
in progress:
image

complete:
image

code:
```
        if self.enable_progress_bar:
            disp_supercells = tqdm(disp_supercells, desc="Computing displacements", unit="displacement")

        phonon.forces = [
            self._calc_forces(supercell)
            for supercell in disp_supercells
            if supercell is not None
        ]

        if self.enable_progress_bar:
            disp_supercells.write("Displacements complete; computing requested properties...")
```

`rich` basic useage:
in progress:
image
complete:
image
code:
```
        if self.enable_progress_bar:
            disp_supercells = track(disp_supercells, description="Computing displacements...")

        phonon.forces = [
            self._calc_forces(supercell)
            for supercell in disp_supercells
            if supercell is not None
        ]
```

`rich` with customisations:
in progress:
image

complete:
image

code:
```
        if self.enable_progress_bar:
            text_column = TextColumn("{task.description}")
            bar_column = BarColumn(bar_width=None)
            completion_column = MofNCompleteColumn()
            time_column = TimeRemainingColumn()
            progress = Progress(text_column, bar_column, completion_column, time_column, expand=True)
            with progress:
                disp_supercells = progress.track(disp_supercells, description="Running displacements...")
                phonon.forces = [
                        self._calc_forces(supercell)
                        for supercell in disp_supercells
                        if supercell is not None
                ]
        else:
            phonon.forces = [
                self._calc_forces(supercell)
                for supercell in disp_supercells
                if supercell is not None
            ]
```

ajjackson: They both look nice to me! Comes down to what Janus is likely to use more widely.

(One other program that could use a progress bar is `singlepoint`, when applied to a file containing many structures.)

ElliottKasoar: Thanks, @RastislavTuranyi!

On balance, I think I'd prefer rich. I don't mind too much about customisations, as long as the options are sensible.

> (One other program that could use a progress bar is `singlepoint`, when applied to a file containing many structures.)

I suppose it depends how often people are using them for large numbers of structures, since each individual calculation is generally pretty quick. (Descriptors can also be calculated for a trajectory, so could also benefit)

ajjackson: > > (One other program that could use a progress bar is `singlepoint`, when applied to a file containing many structures.)
> 
> I suppose it depends how often people are using them for large numbers of structures, since each individual calculation is generally pretty quick. (Descriptors can also be calculated for a trajectory, so could also benefit)

It's quite handy as part of a workflow where configurations are sampled. Getting the results in a single file is also useful in such cases!

RastislavTuranyi:
I have pushed the `rich` version, though we chose the one with the customisations as we thought that having the number counter for the progress through displacements is much nicer (and more helpful!) than the %. For niceness, all the customisation bit was placed into a separate function that works just like `rich.progress.track`, so that in the main code it's just a one-liner. I wasn't sure where to place this function, so for now I put it to `helpers.utils` with the `rich` imports inside the function in order to avoid polluting the entire file with the (apparently slow) rich imports (since I believe `rich` might not necessarily be imported when using Python API).

Additionally, as can be seen in the pictures I posted earlier, my terminal was interpreting the colour of the finished progress bar as grey (even though it should be a dark green), the same as the empty progress bar. Since I might not be the only one to run into this, we decided to change the colours of this progress bar to the janus (logo) colours of light (`#FBBB10`) and dark orange (`#E38408`).

I have also run ruff, so that should hopefully be ok, and I have also added rich as a dependency.

Please let me know if you have any comments/suggestions/etc.!

janus_core/calculations/phonons.py
ElliottKasoar: As long as your input is a single file with many configurations, the output of `singlepoint` should also be a single file with the results as part of the info for each configuration, or do you mean in other cases? 

(I'm going to open an issue for potentially using the progress bar elsewhere, and resolve this since the original comment has been addressed)

janus_core/helpers/utils.py
```diff
@@ -674,3 +674,53 @@ def _dump_csv(
...
+
+    Parameters
+    ----------
+    sequence : Iterable
+        The sequence to iterate over. Must support "len".
+    description : str
+        The text to display to the left of the progress bar.
+
+    Returns
+    -------
```
ElliottKasoar, line 695:
```suggestion
    Yields
    ------
```

ElliottKasoar (changes requested):
This looks great, thanks!

I've added one suggestion that I think should fix the docs.

I think in terms of imports, I'd maybe move them to the top still, to be consistent with our current approach elsewhere.

We definitely want to address imports for the CLI (#272), since it's a much more noticeable issue for autocompletion, but otherwise it's not clear that it's a problem (I mentioned it just as something to keep in mind), and I think maintaining consistency will make it easier if we do decide to change our approach later.

RastislavTuranyi:
Sure! I can definitely move the imports among the other imports; should I leave it in the `helpers.utils` or put it elsewhere? I wasn't really sure

ElliottKasoar:
> Sure! I can definitely move the imports among the other imports; should I leave it in the `helpers.utils` or put it elsewhere? I wasn't really sure

Thanks! `helpers.utils` is perfect

RastislavTuranyi:
Should be fixed now

ElliottKasoar (approved):
Looks good to me, thanks for this!

janus_core/calculations/phonons.py
ajjackson: Yes, I think this is the existing (and desired) behaviour

Issue #310, opened 2024-09-14
refactor phonons


alinelena:
unfortunately phonopy api is not best example of how to design an interface. 
for janus i would say we shall restrict to an explicit mesh array.
float mesh is not only generating the grid but also sets the gamma centre to true... the only suggestion I have, to be done in this or separate is to have mesh-kwargs, so we can pass the rest of the arguments to mesh.

in general as part of this i went through it and i have a list of suggestions for phonons in general. @oerc0122 feedback encouraged.
- [x] symmetrize also structure if force constants are symmetrized
- [ ] dos-kwargs, allow passing freq_min, max and pitch or explicit params? implications are as dos-kwargs, we concentrate only on dos, as explicit we shall apply min,max to bands plotting too.
- [ ] add Q-point/high-symmetry path passing
- [ ] enable compression for bands
- [ ] remove autoband, and replace it with band, natural if we add paths
- [ ] write thermal props using phonopy api. so we can reuse output for other tools.
- [x] support non-diagonal supercells -- @ajjackson below

some proof of concept in here https://gitlab.com/drFaustroll/lavello/-/blob/main/phonons.py?ref_type=heads

_Originally posted by @alinelena in https://github.com/stfc/janus-core/pull/307#discussion_r1759690132_

ajjackson:
I would like to add
- [ ] support non-diagonal supercells

ElliottKasoar:
> [ ] symmetrize also structure if force constants are symmetrized

Does it make sense to add this an option to geometry optimization (either as the only way it's called for phonons, or additionally)?

alinelena:
yes it may make sense, but let us speak in 10 about it

ajjackson:
As well as sum-rules (the current symmetrisation?) Phonopy supports space group symmetrisation https://github.com/phonopy/phonopy/blob/0f881824261903e78bcc45e2b8db44adfdda8ada/phonopy/api_phonopy.py#L1217

ElliottKasoar:
> As well as sum-rules (the current symmetrisation?)

I think we discussed this, but for reference, we use `symmetrize_force_constants(level=1)`, which is described as "This applies translational and permutation symmetries successfully, but not simultaneously.", if that's what you mean?

I'm not against a mutually exclusive `--symmetrize-by-space-group` option, if it would be useful.

ajjackson:
I don't think they are mutually exclusive? It would generally be more useful to do spacegroup symmetrisation followed by sum rules.

[...]

PR #320, opened 2024-10-03
Add support for non-diagonal supercells

RastislavTuranyi:
As discussed, and mentioned in #310, I had a try at adding support for non-diagonal supercells to phonon calculations. The Python API change is small: simply adding support for lists of length 9, which are then turned into a 3x3 matrix. The CLI API is much trickier due to the restrictions that typer/click impose ([certain hacks](https://stackoverflow.com/questions/48391777/nargs-equivalent-for-options-in-click) notwithstanding); for now, I and @ajjackson decided to require the supercell input to be a whitespace-separated string, e.g. `"1 2 3"`), which would have to be passed in as a string. The other option we considered was having two different options for the two cases, e.g. `--supercell-diagonal` and `--supercell-nondiagonal`. However, we aren't sure what would be the best approach; what do you think?

Also, I have tried adding some tests, but let me know if more/different tests are required.

### DIFF
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -194,7 +194,13 @@ class Phonons(BaseCalculation):
         calcs : Optional[MaybeSequence[PhononCalcs]]
             Phonon calculations to run. Default calculates force constants only.
         supercell : MaybeList[int]
-            Size of supercell for calculation. Default is 2.
+            The size of a supercell for calculation, or the supercell itself.
+            If a single number is provided, it is interpreted as the size, so a
+            symmetrical supercell of that size in all dimensions is constructed.
+            If three values are provided, they are interpreted as the diagonal
+            values of a symmetrical supercell. If nine values are provided, they
+            are assumed to be the full supercell matrix in the style of Phonopy.
+            Default is 2.
         displacement : float
             Displacement for force constants calculation, in A. Default is 0.01.
         mesh : tuple[int, int, int]
@@ -249,8 +255,8 @@ class Phonons(BaseCalculation):
 
         # Ensure supercell is a valid list
         self.supercell = [supercell] * 3 if isinstance(supercell, int) else supercell
-        if len(self.supercell) != 3:
-            raise ValueError("`supercell` must be an integer, or list of length 3")
+        if len(self.supercell) not in [3, 9]:
+            raise ValueError("`supercell` must be an integer, or list of length 3 or 9")
 
         # Read last image by default
         read_kwargs.setdefault("index", -1)
@@ -370,11 +376,19 @@ class Phonons(BaseCalculation):
 
         cell = self._ASE_to_PhonopyAtoms(self.struct)
 
-        supercell_matrix = (
-            (self.supercell[0], 0, 0),
-            (0, self.supercell[1], 0),
-            (0, 0, self.supercell[2]),
-        )
+        if len(self.supercell) == 3:
+            supercell_matrix = (
+                (self.supercell[0], 0, 0),
+                (0, self.supercell[1], 0),
+                (0, 0, self.supercell[2]),
+            )
+        else:
+            supercell_matrix = (
+                tuple(self.supercell[:3]),
+                tuple(self.supercell[3:6]),
+                tuple(self.supercell[6:]),
+            )
+
         phonon = phonopy.Phonopy(cell, supercell_matrix)
         phonon.generate_displacements(distance=self.displacement)
         disp_supercells = phonon.supercells_with_displacements
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -257,7 +257,13 @@ class Phonons(BaseCalculation):
         # Ensure supercell is a valid list
         self.supercell = [supercell] * 3 if isinstance(supercell, int) else supercell
         if len(self.supercell) not in [3, 9]:
-            raise ValueError("`supercell` must be an integer, or list of length 3 or 9")
+            raise ValueError(
+                "`supercell` must be an integer, or list of length 3 or 9. A list of "
+                "length 3 must specify the values of a diagonal supercell, and a list "
+                "of length 9 must specify all values of a full supercell matrix, where "
+                "the first three values are the first row, the second three are the "
+                "second row, etc."
+            )
 
         # Read last image by default
         read_kwargs.setdefault("index", -1)
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -196,11 +196,12 @@ class Phonons(BaseCalculation):
         supercell : MaybeList[int]
             The size of a supercell for calculation, or the supercell itself.
             If a single number is provided, it is interpreted as the size, so a
-            symmetrical supercell of that size in all dimensions is constructed.
+            diagonal supercell of that size in all dimensions is constructed.
             If three values are provided, they are interpreted as the diagonal
-            values of a symmetrical supercell. If nine values are provided, they
-            are assumed to be the full supercell matrix in the style of Phonopy.
-            Default is 2.
+            values of a diagonal supercell. If nine values are provided, they
+            are assumed to be the full supercell matrix in the style of Phonopy,
+            so the first three values will be used as the first row, the second
+            three as the second row, etc. Default is 2.
         displacement : float
             Displacement for force constants calculation, in A. Default is 0.01.
         mesh : tuple[int, int, int]
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -59,7 +59,14 @@ class Phonons(BaseCalculation):
     calcs : Optional[MaybeSequence[PhononCalcs]]
         Phonon calculations to run. Default calculates force constants only.
     supercell : MaybeList[int]
-        Size of supercell for calculation. Default is 2.
+        The size of a supercell for calculation, or the supercell itself.
+        If a single number is provided, it is interpreted as the size, so a
+        diagonal supercell of that size in all dimensions is constructed.
+        If three values are provided, they are interpreted as the diagonal
+        values of a diagonal supercell. If nine values are provided, they
+        are assumed to be the full supercell matrix in the style of Phonopy,
+        so the first three values will be used as the first row, the second
+        three as the second row, etc. Default is 2.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
     mesh : tuple[int, int, int]
--- a/janus_core/cli/phonons.py
+++ b/janus_core/cli/phonons.py
@@ -3,7 +3,7 @@
 from pathlib import Path
 from typing import Annotated, Optional
 
-from typer import BadParameter, Context, Option, Typer
+from typer import Context, Option, Typer
 from typer_config import use_config
 
 from janus_core.cli.types import (
@@ -29,26 +29,19 @@ def phonons(
     ctx: Context,
     struct: StructPath,
     supercell: Annotated[
-        tuple[int, int, int],
+        str,
         Option(
-            help="Supercell lattice vectors in the form '1 2 3'. "
-            "Mutually exclusive with --supercell-matrix. "
-            "Default is '2 2 2', which is used when neither "
-            "--supercell nor --supercell-matrix are provided."
+            help="Supercell specification as a string, in the Phonopy style. "
+            "There are three different valid inputs: single value (i.e. '2') "
+            "which specifies what all of the diagonal values of the supercell "
+            "matrix whould be, three values (i.e. '1 2 3') which specify the "
+            "individual diagonal elements of the supercell matrix, or nine "
+            "values (i.e. '1 2 3 4 5 6 7 8 9') which specify all the elements "
+            "of the full supercell matrix. In the last case, the first three "
+            "values form the first row of the matrix, the second three the "
+            "second row, etc."
         ),
-    ] = None,
-    supercell_matrix: Annotated[
-        tuple[int, int, int, int, int, int, int, int, int],
-        Option(
-            help="Full supercell matrix in the form '1 2 3 4 5 6 7 8 9'"
-            " The values are interpreted as in phonopy, i.e. "
-            "the first three values specify the first row of "
-            "the supercell matrix, the second three the second "
-            "row, etc. Mutually exclusive with --supercell. If "
-            "neither --supercell nor --supercell-matrix are "
-            "provided, '--supercell 2 2 2' is used as default."
-        ),
-    ] = None,
+    ] = "2 2 2",
     displacement: Annotated[
         float, Option(help="Displacement for force constants calculation, in A.")
     ] = 0.01,
@@ -132,17 +125,16 @@ def phonons(
         Typer (Click) Context. Automatically set.
     struct : Path
         Path of structure to simulate.
-    supercell : tuple[int, int, int]
-        Supercell lattice vectors in the form '1 2 3'. Mutually exclusive with
-        `supercell_matrix`. Default is `(2, 2, 2)`, which is used when neither
-        the `supercell` nor `supercell_matrix` parameters are provided.
-    supercell_matrix : tuple[int, int, int, int, int, int, int, int, int]
-        Full supercell matrix in the form '1 2 3 4 5 6 7 8 9' The values are
-        interpreted as in phonopy, i.e. the first three values specify the first
-        row of the supercell matrix, the second three the second row, etc.
-        Mutually exclusive with `supercell`. If neither `supercell` nor
-        `supercell_matrix` are provided, this parameter is ignored and the default
-        value for the `supercell` parameter is used.
+    supercell : str
+        Supercell specification as a string, in the Phonopy style.
+        There are three different valid inputs: single value (i.e. '2')
+        which specifies what all of the diagonal values of the supercell
+        matrix whould be, three values (i.e. '1 2 3') which specify the
+        individual diagonal elements of the supercell matrix, or nine
+        values (i.e. '1 2 3 4 5 6 7 8 9') which specify all the elements
+        of the full supercell matrix. In the last case, the first three
+        values form the first row of the matrix, the second three the
+        second row, etc.
     displacement : float
         Displacement for force constants calculation, in A. Default is 0.01.
     mesh : tuple[int, int, int]
@@ -217,18 +209,6 @@ def phonons(
         set_read_kwargs_index,
         start_summary,
     )
-    if supercell is not None:
-        if supercell_matrix is not None:
-            raise BadParameter(
-                "The --supercell and --supercell-matrix parameters are mutually "
-                "exclusive. Please use only one one of the options."
-            )
-        supercell_value = list(supercell)
-    else:
-        if supercell_matrix is not None:
-            supercell_value = list(supercell_matrix)
-        else:
-            supercell_value = [2, 2, 2]
 
     # Check options from configuration file are all valid
     check_config(ctx)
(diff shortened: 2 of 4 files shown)


janus_core/calculations/phonons.py
```diff
@@ -232,8 +238,8 @@ def __init__(
 
         # Ensure supercell is a valid list
         self.supercell = [supercell] * 3 if isinstance(supercell, int) else supercell
-        if len(self.supercell) != 3:
-            raise ValueError("`supercell` must be an integer, or list of length 3")
+        if len(self.supercell) not in [3, 9]:
+            raise ValueError("`supercell` must be an integer, or list of length 3 or 9")
```
alinelena, line 242: you need to be more informative, 9 in what order for matrix.

RastislavTuranyi: The goal was to reproduce the [phonopy behaviour](https://phonopy.github.io/phonopy/setting-tags.html#dimension-tag), so first three values are the first row, etc. Should I write that explicitly into the documentation, or would you prefer it elsewhere?

ElliottKasoar:
We'll need to figure out how this fits with #307. If that's merged, supercell would be be a tuple (`--supercell x y z`), but I don't think this works if we want a input variable length, so I may need to revert that change.

ElliottKasoar:
> > We'll need to figure out how this fits with #307. If that's merged, supercell would be be a tuple (`--supercell x y z`), but I don't think this works if we want a input variable length, so I may need to revert that change.
> 
> In that case, we could go for the other option of having two different options, `--supercell-diagonal 1 2 3` and `--supercell-nondiagoal 1 2 3 4 5 6 7 8 9`, if that would be preferable? We weren't sure which approach would work better with the API you are aiming for

I'm not set on anything too specific, but yes given that the approach consistent with the current main would be `"1x2x3x4x5x6x7x8x9"`, which is a bit unwieldy, separating the options might make most sense.

I'm hesitant to change `--supercell` though. Do you think it would still make sense to have `--supercell` and `--supercell-nondiag`?

If you can find alternatives that let us neatly unify them (maybe with a custom type - I've seen that as a suggestion for [click Unions](https://github.com/pallets/click/issues/1729), which typer is based on) that would also be welcome.

janus_core/calculations/phonons.py
ajjackson: In some ways "following Phonopy convention" seems _more_ informative than listing array elements.

ajjackson:
> I'm not set on anything too specific, but yes given that the approach consistent with the current main would be `"1x2x3x4x5x6x7x8x9"`, which is a bit unwieldy, separating the options might make most sense.

It isn't only unwieldy; it's nonsensical! In the diagonal case we can "multiply" to describe the overall cell volume, but it doesn't work for matrix elements.

RastislavTuranyi:
Another option I thought of would be to have only `--supercell`, but make it be able to be called multiple times, so `--supercell 1 2 3` would be diagonal, and `--supercell 1 2 3 --supercell 4 5 6 --supercell 7 8 9` would be the full matrix. That said, though, I am not sure I like it, and from some quick text I have a feeling it would not necessarily be easy to implement with typer. Overall, it seems to me that the `--supercell` and `--supercell-matrix` might be the best option.

> If you can find alternatives that let us neatly unify them (maybe with a custom type - I've seen that as a suggestion for https://github.com/pallets/click/issues/1729, which typer is based on) that would also be welcome.

I had a try experimenting with this, but I can only see how that would allow us to parse the two strings, i.e. `--supercell "1 2 3"` vs `--supercell "1 2 3 4 5 6 7 8 9"`; I can't figure out how that could be done for the tuple case, but then I am not experienced with typer/click so there might be a way. What I have tried is to use the `click_type=Union()` parameter, but that leads to only the first value being passed in to the custom `Union` type, i.e. `--supercell 1 2 3` -> the `Union.convert()` method only receives the `1`.

I have also tried the [hack](https://stackoverflow.com/questions/48391777/nargs-equivalent-for-options-in-click) I mentioned earlier, but I'm unclear on how to make it work with typer since it's unclear where/how typer actually uses `click.Option`, and typer doesn't have _any_ API documentation 

There is probably a way to use both of the above workarounds, but I reckon it will probably need much deeper understanding of how typer works.

Overall, it seems to me that the `--supercell` and `--supercell-matrix` might be the best option.

oerc0122:
> Another option I thought of would be to have only `--supercell`, but make it be able to be called multiple times, so `--supercell 1 2 3` would be diagonal, and `--supercell 1 2 3 --supercell 4 5 6 --supercell 7 8 9` would be the full matrix. That said, though, I am not sure I like it, and from some quick text I have a feeling it would not necessarily be easy to implement with typer. Overall, it seems to me that the `--supercell` and `--supercell-matrix` might be the best option.

Single responsibility principle in me is screaming at this suggestion. They're two different jobs which have two different outcomes and two different meanings. They shouldn't share a name even if they ultimately affect the same var.

ajjackson:
I think this might be exactly how Typer wants to do it, but not how humans want to do it 😅

oerc0122:
> > > Another option I thought of would be to have only `--supercell`, but make it be able to be called multiple times, so `--supercell 1 2 3` would be diagonal, and `--supercell 1 2 3 --supercell 4 5 6 --supercell 7 8 9` would be the full matrix. That said, though, I am not sure I like it, and from some quick text I have a feeling it would not necessarily be easy to implement with typer. Overall, it seems to me that the `--supercell` and `--supercell-matrix` might be the best option.
> > 
> > 
> > Single responsibility principle in me is screaming at this suggestion. They're two different jobs which have two different outcomes and two different meanings. They shouldn't share a name even if they ultimately affect the same var.
> 
> Is your objection to using `--supercell` in principle for both diagonal terms (if three values are specified) and the full matrix (if nine are specified), which doesn't seem unreasonable in theory, or that the meaning of the first `--supercell x y z` changes, which to me is more problematic?

That the meaning of `supercell` is changing depending on how many times it's called.

ElliottKasoar:
> > > > Another option I thought of would be to have only `--supercell`, but make it be able to be called multiple times, so `--supercell 1 2 3` would be diagonal, and `--supercell 1 2 3 --supercell 4 5 6 --supercell 7 8 9` would be the full matrix. That said, though, I am not sure I like it, and from some quick text I have a feeling it would not necessarily be easy to implement with typer. Overall, it seems to me that the `--supercell` and `--supercell-matrix` might be the best option.
> > > 
> > > 
> > > Single responsibility principle in me is screaming at this suggestion. They're two different jobs which have two different outcomes and two different meanings. They shouldn't share a name even if they ultimately affect the same var.
> > 
> > 
> > Is your objection to using `--supercell` in principle for both diagonal terms (if three values are specified) and the full matrix (if nine are specified), which doesn't seem unreasonable in theory, or that the meaning of the first `--supercell x y z` changes, which to me is more problematic?
> 
> That the meaning of `supercell` is changing depending on how many times it's called. Notably with "the same arguments"

I wouldn't quite describe it as being called, it's the typer way of creating lists, so [(x, y, z)] being different to [(a, b, c), (d, e, f), (g, h, i)] isn't quite so unreasonable, but yes it probably is still more confusing than alternatives, giving that you might start with `--supercell x y z` and want to add off diagonal terms.

ajjackson:
I agree that both
- within Typer's way of thinking it makes sense that it would form an array
- to a reasonable person, it looks more like you are declaring multiple job steps

RastislavTuranyi:
Just found this [typer PR](https://github.com/fastapi/typer/pull/800) which does almost what we want, jut requiring a specific separator.

Also found [this hack](https://github.com/fastapi/typer/issues/554#issuecomment-2323176039) which actually pre-processes the CLI and therefore seems to be the only thing that has a chance of working with typer.

ElliottKasoar:
One option which I'm pretty sure works if duplicating the custom `TyperDict` class and parsers we already have, which then allows us to parse arbitrary tuples e.g. `--supercell "(2, 2, 2)"`.

I think it's marginally nicer as a tuple, but the same thing could be done to parse comma separated text (`--supercell "1,1,1")`.

We'd just need to validate the length etc.

janus_core/calculations/phonons.py
alinelena: do not assume that anyone will know what is the phonopy convention... just list it.

RastislavTuranyi:
Since there was no additional discussion, I have implemented the `--supercell 1 2 3` and `--supercell-matrix 1 2 3 4 5 6 7 8 9` version. Please let me know if you have any comments/suggestions/etc.!

janus_core/calculations/phonons.py
RastislavTuranyi: I have changed the documentation and error messages; is the current version OK, or do I need to be more detailed/write it in another way?

alinelena:
@RastislavTuranyi this will need to be rebased, can you ask @ajjackson to help you if you never done it before?

ElliottKasoar:
> Since there was no additional discussion, I have implemented the `--supercell 1 2 3` and `--supercell-matrix 1 2 3 4 5 6 7 8 9` version. Please let me know if you have any comments/suggestions/etc.!

Thanks for this!

Did you rule out the approach I suggested with building tuples similarly to how dictionaries are currently implemented?

To me, it feels nicer to have the same option, `--supercell "(1, 2, 3)"` or `--supercell "(1, 2, 3, 4, 5, 6, 7, 8, 9)"`, and it could also allow us to take a single integer for the supercell through the CLI, which would be consistent with the Python interface.

Whether it's two similarly-named options, or one option that handles different lengths slightly differently, I think some ambiguity before reading the docs is unavoidable, so I don't think there's a perfect option.

I don't think the customs tuples have quite the same issues as we discussed with `--supercell 1 2 3 ` vs  `--supercell 1 2 3 --supercell 4 5 6 --supercell 7 8 9`, but it is admittedly less nice to manually parse the tuples than use the built-in tuples.

I'm ok with  `--supercell` / `--supercell-matrix`, if other people who use phonopy etc more think it makes more sense, just want to check the options have been considered.

alinelena:
> I'm ok with `--supercell` / `--supercell-matrix`, if other people who use phonopy etc more think it makes more sense, just want to check the options have been considered.

technically there are three meaningful ways to specify a supercell, depending what you are after, single scalar n, for same repetition in all directions, most common choice, n1 n2 n3, maybe different numbers in all 3 directions, and the full matrix. given the other ones are shortcuts to the generalised version personally I will go with one --supercell, and based on the input, decide if we deal with generalised or one of the two shortcuts. I agree with @oerc0122 in using different names for things that do different things but in this case we do the same thing specify a supercell with different shortcuts.

oerc0122:
> I agree with @oerc0122 in using different names for things that do different things but in this case we do the same thing specify a supercell with different shortcuts.

I would argue that because the arguments are different types (3-tuple, 9-tuple, nothing in-between) It's pretty reasonable to specify them through different cmdline args so you have to be explicit and there's less confusion in scripts. For scalar -> `supercell_matrix` / scalar -> `supercell` I think that's a fairly obvious abbreviation with the same meaning, but the fact that 3-tuple passes to the diagonal only, it's a different situation. 

@ElliottKasoar 's suggestion on the tuples is fine, but would a user expect to enter "(1, 2, 3, 4, 5, 6, 7, 8, 9)" or "((1, 2, 3), (4, 5, 6), (7, 8, 9))" would they be different? How about "(1, 2, 3)" or "(3,)" or "(3)"? What about if they put in "((1,), (2, 3), (4, 5, 6, 7, 8, 9))". It's a lot more effort to define in some senses.

ElliottKasoar:
> @ElliottKasoar 's suggestion on the tuples is fine, but would a user expect to enter "(1, 2, 3, 4, 5, 6, 7, 8, 9)" or "((1, 2, 3), (4, 5, 6), (7, 8, 9))" would they be different? How about "(1, 2, 3)" or "(3,)" or "(3)"? What about if they put in "((1,), (2, 3), (4, 5, 6, 7, 8, 9))". It's a lot more effort to define in some senses.

I'm open to alternatives, but I assumed we'd define the input as either an integer (`"1"`) or 3- or 9-tuples (`"(1, 2, 3)"` or `"(1, 2, 3, 4, 5, 6, 7, 8, 9)"`, which seems relatively simple to explain.

I think something like `"(1)"`is generally treated as an integer, (`isinstance((1), int) is True`), but if it didn't work I don't think too many people would complain.

We could also permit `"(1,)"`, since for parsing we may want it in that form anyway, but again I wouldn't object if it were rejected.

ajjackson:
I'm not a fan of parentheses in the string, unless we have line breaks as well it does very little to improve legibility.

Phonopy allows the user to provide 3 or 9 arguments to the same parameter, so the most "phonopy native" syntax is not really viable. String quoting with spaces in between numbers is an approximation to this. I would be fine with giving 9 values all the time but this is likely unpopular with other users...

I think having separate `--supercell` and `--supercell-matrix` arguments gives the least terrible type hints, given the limitations of Typer.

ElliottKasoar:
> I'm not a fan of parentheses in the string, unless we have line breaks as well it does very little to improve legibility.
> 
> Phonopy allows the user to provide 3 or 9 arguments to the same parameter, so the most "phonopy native" syntax is not really viable. String quoting with spaces in between numbers is an approximation to this. I would be fine with giving 9 values all the time but this is likely unpopular with other users...
> 
> I think having separate `--supercell` and `--supercell-matrix` arguments gives the least terrible type hints, given the limitations of Typer.

I wouldn't necessarily say it's about legibility, more just familiarity with how you would define it in the Python, and convenience for parsing it into an actual tuple. Removing the parenthesis and having space/comma separated options in a string would be a similar, reasonable solution, but would maybe feels a bit more arbitrary.

Again, I'm happy to go with the consensus, but when we already have a lot of options, and potentially more on the way with the refactor, splitting it into two feels a bit unnecessary. Also, while the type hint for each option's purpose may be slightly simpler, we'd probably also need to describe in at least one of them, maybe both, how they interact, which adds a bit more complication.

ajjackson:
> I wouldn't necessarily say it's about legibility, more just familiarity with how you would define it in the Python, and convenience for parsing it into an actual tuple.

Typically command-line arguments that would be mapped to a python tuple accept NARGS values; that's the familiar interface for such things. Typing Python code as a command-line argument feels a bit odd and uncomfortable; maybe because it triggers a suspicion that this string is about to get `eval()`ed!

ElliottKasoar:
> Typically command-line arguments that would be mapped to a python list or tuple accept NARGS values directly; that's the familiar interface for such things. Typing Python code as a command-line argument feels a bit odd and uncomfortable; maybe because it alerts a primal fear that this string is about to get `eval()`ed!

A legitimate fear! I agree it's very far from perfect, and unintuitive for CLI users, but I think we're stuck finding the least bad option.

My point about other alternatives being more arbitrary was that if it can't be done in a way that's familiar in terms of CLIs, mirroring the Python at least adds some level familiarity, and consistency with kwarg dict inputs, so it's probably easier to remember than a more arbitrary construct, even if it does look strange.

Splitting it into two options does allow nicer inputs, so the question is whether that's worth the added complication of two similar,  competing options, no integer input option, and a less obvious mapping to phonopy.

I think we probably do need to settle on something though, at least between one or two command line options!

ElliottKasoar:
In the interest of finishing this up, and after discussing it a little more with @alinelena, I think we should go for `--supercell` as the only input option, and then we should allow one, three, or nine integers, dealing with them as required.

I don't mind too much between "x y z" and "(x, y, z)", if it works. (Maybe the former is a little closer how we'd like the tuple input to be, so makes slightly more sense? I've only tried the latter though, so implementation may be tricker.)

Would that be ok, @RastislavTuranyi? If how to deal with the parsing etc. is unclear, I'm happy to help more.

ajjackson:
> I've only tried the latter though, so implementation may be tricker

Here are a couple of ideas

```python
tuple(map(int, supercell.split())) # "1 2 3"

json.loads(supercell.replace("(", "[").replace(")", "]")) # "(1, 2, 3)"
```

Whitespace separation looks cleaner to me!

oerc0122:
> > I've only tried the latter though, so implementation may be tricker
> 
> Here are a couple of ideas
> 
> ```python
> tuple(map(int, supercell.split())) # "1 2 3"
> 
> json.loads(supercell.replace("(", "[").replace(")", "]")) # "(1, 2, 3)"
> ```
> 
> Whitespace separation looks cleaner to me!

For the latter 
```python
tuple(map(int, supercell.strip("()").split(","))
```

ajjackson:
If the parser is going to strip the parens off, why ask for them?

oerc0122:
> If the parser isn't going to look at the parens, why ask for them?

Agreed, I personally wouldn't expect to add parens to a CLI.

alinelena:
rebase

RastislavTuranyi:
> rebase

Done! I have also reverted to the string implementation, since that appeared to be the consensus?

janus_core/calculations/phonons.py
ElliottKasoar: @alinelena do we actually want the `ValueError` to be quite so detailed?

I agree when we do specify the form of the inputs, we should be clear in our code and not just reference external codes e.g. Phonopy, but to me, something like the order the matrix elements makes more sense sticking to the docstring, otherwise we end up with a lot of duplication.

Maybe it makes sense though, given that it's probably the most likely thing people might trip up on

janus_core/cli/phonons.py
```diff
@@ -206,6 +225,24 @@ def phonons(
         raise ValueError("'fmax' must be passed through the --fmax option")
     minimize_kwargs["fmax"] = fmax
 
+    try:
+        supercell = [int(x) for x in supercell.split()]
+    except ValueError as exc:
+        raise ValueError(
+            "Please pass lattice vectors as integers in the form '1 2 3'"
```
ElliottKasoar, line 232: This doesn't necessarily need the full set of options again, but could maybe be a little more descriptive. Perhaps

```suggestion
            "Please pass lattice vectors as space-separated integers in quotes. "
            "For example, '1 2 3'"
```

ElliottKasoar (changes requested):
Looking good!

Could you please update the docs? In particular, the command line interface page of the user guide, but worth checking `supercell` doesn't appear anywhere else.

My only other thought is if we can make the docstrings slightly more concise, but if in doubt I'm happy to err on the side of clarity.

janus_core/cli/phonons.py
RastislavTuranyi: Done!

RastislavTuranyi:
> Could you please update the docs? In particular, the command line interface page of the user guide, but worth checking `supercell` doesn't appear anywhere else.

I have updated the code examples to use quotes for the integers, but would you like me to add a paragraph explaining the various possible inputs for `--supercell` as well? Or maybe just reuse the existing code examples by passing in the supercell using a different format in each?

I haven't found any mentions of phonons or supercell in the rest of the .rst documents; they either don't mention it or are auto-generated. Maybe the jupyter notebook examples should be updated?

janus_core/cli/phonons.py
```diff
@@ -29,8 +29,19 @@ def phonons(
...
+        Option(
+            help="Supercell specification as a string, in the Phonopy style. "
+            "There are three different valid inputs: single value (i.e. '2') "
+            "which specifies what all of the diagonal values of the supercell "
+            "matrix whould be, three values (i.e. '1 2 3') which specify the "
+            "individual diagonal elements of the supercell matrix, or nine "
+            "values (i.e. '1 2 3 4 5 6 7 8 9') which specify all the elements "
+            "of the full supercell matrix. In the last case, the first three "
+            "values form the first row of the matrix, the second three the "
+            "second row, etc."
```
ElliottKasoar, line 42:
```suggestion
            help="Supercell matrix, in the Phonopy style. Must be passed as a string "
            "in one of three forms: single integer ('2'), which specifies all "
            "diagonal elements; three integers ('1 2 3'), which specifies each "
            "individual diagonal element; or nine values ('1 2 3 4 5 6 7 8 9'), which "
            "specifies all elements, filling the matrix row-wise."
```

Maybe, to be slightly more concise for the CLI help in particular? If not, then just note there's a typo: `whould` here and in the `supercell : str` description

ElliottKasoar (changes requested):
I can't comment on the code, but the `supercell : MaybeList[int]` description for the `Phonons` class parameters needs updating to match the `__init__` definition.

Would be good if anyone else who uses Phonopy can confirm they're happy, but other than the minor suggestions (/typo corrections) I think this is good to go from my perspective

janus_core/cli/phonons.py
RastislavTuranyi: Adam was ok with these rewrites, so I have added them as is, thanks!

RastislavTuranyi:
Everything should be updated now, though please do let me know if the CLI documentation is too verbose or not detailed enough.

ElliottKasoar (approved):

ElliottKasoar:
> Everything should be updated now, though please do let me know if the CLI documentation is too verbose or not detailed enough.

I think the section you added is a good amount, thanks! Generally I lean towards more concise wording for something like the output from `--help`, so it doesn't fill up the terminal, but slightly more verbose for the full .rst docs, since space is less of a problem, but either way clarity and unambiguity should come first.

Thank you everyone for the work that's gone into this!

PR #322, opened 2024-10-03
Fix unimplemented descriptors

### LINKED ISSUE #309: Catch unsupported descriptor calculations
If you try to calculate descriptors for non-MACE architectures, we should handle the error better to make it clear these are not currently supported.

```
janus descriptors --struct tests/data/NaCl.cif --arch chgnet

...
...
AttributeError: 'CHGNetCalculator' object has no attribute 'get_descriptors'
```

We should also consider how we want to deal with `SumCalculator`s which include a valid MACE calculator:

```
janus descriptors --struct NaCl-results.extxyz --calc-kwargs "{'dispersion': True}"
...
...
AttributeError: 'SumCalculator' object has no attribute 'get_descriptors'
```

ElliottKasoar:
Resolves #309

The solution for the `SumCalculator` case isn't the nicest, but hopefully the checks are specific enough that it's reasonable.

The two added tests both fail prior to the subsequent changes :

```
FAILED tests/test_descriptors.py::test_dispersion - AttributeError: 'SumCalculator' object has no attribute 'get_descriptors'
FAILED tests/test_descriptors.py::test_not_implemented_error - Failed: DID NOT RAISE <class 'NotImplementedError'>
```

### DIFF
--- a/janus_core/calculations/descriptors.py
+++ b/janus_core/calculations/descriptors.py
@@ -7,7 +7,6 @@ from ase import Atoms
 from ase.calculators.calculator import Calculator
 from ase.calculators.mixing import SumCalculator
 import numpy as np
-from torch_dftd.torch_dftd3_calculator import TorchDFTD3Calculator
 
 from janus_core.calculations.base import BaseCalculation
 from janus_core.helpers.janus_types import (
@@ -192,7 +191,7 @@ class Descriptors(BaseCalculation):
         if isinstance(calc, SumCalculator):
             if (
                 len(calc.mixer.calcs) == 2
-                and isinstance(calc.mixer.calcs[1], TorchDFTD3Calculator)
+                and calc.mixer.calcs[1].name == "TorchDFTD3Calculator"
                 and hasattr(calc.mixer.calcs[0], "get_descriptors")
             ):
                 calc.get_descriptors = calc.mixer.calcs[0].get_descriptors
--- a/janus_core/calculations/descriptors.py
+++ b/janus_core/calculations/descriptors.py
@@ -4,7 +4,10 @@ from collections.abc import Sequence
 from typing import Any, Optional
 
 from ase import Atoms
+from ase.calculators.calculator import Calculator
+from ase.calculators.mixing import SumCalculator
 import numpy as np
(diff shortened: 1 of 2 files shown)


janus_core/calculations/descriptors.py
```diff
@@ -4,7 +4,10 @@
 from typing import Any, Optional
 
 from ase import Atoms
+from ase.calculators.calculator import Calculator
+from ase.calculators.mixing import SumCalculator
 import numpy as np
+from torch_dftd.torch_dftd3_calculator import TorchDFTD3Calculator
```
alinelena, line 10: why do we need torch? descriptors should not need it.

ElliottKasoar: We don't, and the descriptors don't use it, which is why if the calculator is a `SumCalculator` including `TorchDFTD3Calculator`, `get_descriptors` uses the MLIP calculator only.

If a user already has, for example, a `SinglePoint` object using mace-mp + dispersion for other reasons, but wants to also get the descriptors, it would fail without this.

alinelena: I am not very keen to import an entire package just because of the user... I wonder if makes more sense to check if features that involve sum calculator are involved and error on that...

oerc0122: Can we be hacky and get away with a name check? It's not as reliable, but avoids the import. Not sure there're any other ways to avoid it and that one also makes my skin crawl to some extent.

ElliottKasoar: Yeah that seems reasonable

janus_core/calculations/descriptors.py
```diff
@@ -163,12 +165,43 @@ def __init__(
         ):
             raise ValueError("Please attach a calculator to `struct`.")
 
+        if isinstance(self.struct, Atoms):
+            self._check_calculator(self.struct.calc)
+        if isinstance(self.struct, Sequence):
```
oerc0122, line 170: Does this need to be `elif`? Isn't `Atoms` a sequence?

oerc0122 (changes requested):

oerc0122 (approved):

PR #326, opened 2024-10-08
Add symmetry refining

ElliottKasoar:
Address part of #310

- Adds `snap_symmetry` function to utils, currently used explicitly in geom_opt, and indirectly for phonons
  - Currently this only applies after optimization. As discussed with @oerc0122 and others, it may also be useful to apply this before optimization, but I'm not sure if that should be built in to this, or simply an option via the Python utils, which this allows
- Change defaults to save optimized file automatically
- ~Add displacement_kwargs, passed to `generate_displacements`~

These could arguably be three separate PRs, but the latter two will be useful for testing `snap_symmetry` (e.g. setting the `seed`), so I've kept them together for now.

To do:

- [x] Add tests

### DIFF
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -307,10 +307,10 @@ class GeomOpt(BaseCalculation):
 
             # Update max force
             old_max_force = max_force
-            if self.filter_func is not None:
-                max_force = linalg.norm(self.filtered_struct.get_forces(), axis=1).max()
-            else:
-                max_force = linalg.norm(self.struct.get_forces(), axis=1).max()
+            struct = (
+                self.filtered_struct if self.filter_func is not None else self.struct
+            )
+            max_force = linalg.norm(struct.get_forces(), axis=1).max()
 
             if max_force >= old_max_force:
                 warnings.warn(
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -18,6 +18,7 @@ from janus_core.helpers.utils import (
     Devices,
     none_to_dict,
     output_structs,
+    snap_symmetry,
     spacegroup,
 )
 
@@ -56,6 +57,8 @@ class GeomOpt(BaseCalculation):
         Set force convergence criteria for optimizer in units eV/Å. Default is 0.1.
     steps : int
         Set maximum number of optimization steps to run. Default is 1000.
+    symmetrize : bool
+        Whether to refine symmetry after geometry optimization. Default is False.
     symmetry_tolerance : float
         Atom displacement tolerance for spglib symmetry determination, in Å.
         Default is 0.001.
@@ -104,6 +107,7 @@ class GeomOpt(BaseCalculation):
         tracker_kwargs: Optional[dict[str, Any]] = None,
         fmax: float = 0.1,
         steps: int = 1000,
+        symmetrize: bool = False,
         symmetry_tolerance: float = 0.001,
         angle_tolerance: float = -1.0,
         filter_func: Optional[Union[Callable, str]] = FrechetCellFilter,
@@ -148,6 +152,8 @@ class GeomOpt(BaseCalculation):
             Set force convergence criteria for optimizer in units eV/Å. Default is 0.1.
         steps : int
             Set maximum number of optimization steps to run. Default is 1000.
+        symmetrize : bool
+            Whether to refine symmetry after geometry optimization. Default is False.
         symmetry_tolerance : float
             Atom displacement tolerance for spglib symmetry determination, in Å.
             Default is 0.001.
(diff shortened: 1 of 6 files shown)


janus_core/calculations/geom_opt.py
```diff
@@ -56,6 +57,8 @@ class GeomOpt(BaseCalculation):
         Set force convergence criteria for optimizer in units eV/Å. Default is 0.1.
     steps : int
         Set maximum number of optimization steps to run. Default is 1000.
+    symmetrize : bool
```
oerc0122, line 60: Just to double check US/UK ise/ize consistency in API.

janus_core/calculations/geom_opt.py
```diff
@@ -289,15 +296,30 @@ def run(self) -> None:
...
 
+        if self.symmetrize:
+            snap_symmetry(self.struct, self.symmetry_tolerance)
+
+            # Update max force
+            old_max_force = max_force
+            if self.filter_func is not None:
+                max_force = linalg.norm(self.filtered_struct.get_forces(), axis=1).max()
+            else:
+                max_force = linalg.norm(self.struct.get_forces(), axis=1).max()
```
oerc0122, line 313:
```suggestion
            struct = self.filtered_struct if self.filter_func is not None else self.struct
            max_force = linalg.norm(struct.get_forces(), axis=1).max()
```

janus_core/helpers/utils.py
```diff
@@ -188,6 +189,21 @@ def spacegroup(
     )
 
 
+def snap_symmetry(struct: Atoms, sym_tolerance: float = 0.001) -> None:
+    """
+    Refine symmetry of structure.
```
oerc0122, line 194: "Refine" is an odd phrasing for forcing the symmetry

janus_core/calculations/geom_opt.py
```diff
@@ -289,15 +296,30 @@ def run(self) -> None:
...
-        self.struct.info["final_spacegroup"] = s_grp
-
         # Calculate current maximum force
         if self.filter_func is not None:
             max_force = linalg.norm(self.filtered_struct.get_forces(), axis=1).max()
         else:
             max_force = linalg.norm(self.struct.get_forces(), axis=1).max()
 
+        if self.symmetrize:
+            snap_symmetry(self.struct, self.symmetry_tolerance)
```
alinelena, line 306: probably you want also the angle passed in here. also should it be snap_to_symmetry?

janus_core/calculations/geom_opt.py
ElliottKasoar: I somewhat regret the choice but we generally use ize e.g. optimize/minimize, so this should be consistent.

(From a very quick look I think the only thing I've been inconsistent with is "Initialise" in docstrings, but variables are all ok I think)

janus_core/helpers/utils.py
ElliottKasoar: I'm borrowing from the ASE, but we can change it if you think force or similar is more explicit

```
def refine_symmetry(atoms, symprec=0.01, verbose=False):
    """
    Refine symmetry of an Atoms object
```

janus_core/calculations/geom_opt.py
ElliottKasoar: `refine_symmetry` only takes one parameter for identifying the symmetry, so how would the angle be used? https://wiki.fysik.dtu.dk/ase/_modules/ase/spacegroup/symmetrize.html#refine_symmetry

ElliottKasoar:
>   * Currently this only applies after optimization. As discussed with @oerc0122 and others, it may also be useful to apply this before optimization, but I'm not sure if that should be built in to this, or simply an option via the Python utils, which this allows

Actually, the `FixSymmetry` constraint already calls `refine_symmetry` as part of its init (https://gitlab.com/ase/ase/-/blob/3.23.0/ase/constraints.py?ref_type=tags#L2301), so does that cover the main use case already?

janus_core/calculations/geom_opt.py
alinelena: yes the issue is upstream, they use it inconsitent
they do not pass angle_tolerance even if spglib needs it, I would suggest correcting this upstream. there are 2 or 3 places where needs to be corrected, practically api of refine_symmetry and maybe 2 functions.

ElliottKasoar: In spglib itself, it does say 

```
    angle_tolerance : float
        Symmetry search tolerance in the unit of angle deg.
        Normally it is not recommended to use this argument.
        See a bit more detail at :ref:`variables_angle_tolerance`.
        If the value is negative, an internally optimized routine is used to
        judge symmetry.
```

Either way, even if we contributed to the upstream now, we wouldn't be able to consistently use it until the next tagged release.

alinelena: yes will not be now but will be good to be done. I know the warning issue is that we are inconsitent

oerc0122 (approved):

PR #329, opened 2024-10-10
Refactor phonons

### LINKED ISSUE #310: refactor phonons

ElliottKasoar:
Resolves #310

I think this addresses the remaining items, other than non-diagonal supercells, which will be covered by #320.

To do:

- [x] Add tests for displacement kwargs
- [x] Add tests for paths (@alinelena do you have a suitable, very basic, example input file?)
- [x] Update docs for paths in particular
- [x] Add option to define `npoints`, rather than hard coded value?


janus_core/calculations/phonons.py
```diff
@@ -496,17 +524,44 @@ def write_bands(
...
+            ), "Number of labels is different to number of q-points"
+
+            q_points, connections = get_band_qpoints_and_path_connections(
+                band_paths=loaded_paths["paths"], npoints=loaded_paths["npoints"]
+            )
+
+        else:
+            bands_file = self._build_filename("auto_bands.yml.xz", filename=bands_file)
+            q_points, labels, connections = get_band_qpoints_by_seekpath(
+                self.results["phonon"].primitive, 51
```
alinelena, line 547: this shall be read from some param with a default value.

alinelena:
I have attached two paths.yml we can still discuss the best way to have the info in.

1. is full what comes from seekpath 
NaCl-auto_bands
[paths.yaml.txt](https://github.com/user-attachments/files/17437426/paths.yaml.txt)

2. this is a random path (just second branch from above) 
NaCl-small-mace_mp-bands
[path_a.yml.txt](https://github.com/user-attachments/files/17437468/path_a.yml.txt)

janus_core/cli/phonons.py
```diff
@@ -10,6 +10,8 @@
     Architecture,
     CalcKwargs,
     Device,
+    DisplacementKwargs,
+    DoSKwargs,
```
alinelena, line 14: also pdos takes extra arg, they are same as dos but few extra on top

alinelena:
this will need a rebase otherwise is fine with me

alinelena (approved):

ElliottKasoar:
(Rebased)

janus_core/calculations/phonons.py
```diff
@@ -223,15 +246,24 @@ def __init__(
             Default is False.
         minimize_kwargs : Optional[dict[str, Any]]
             Keyword arguments to pass to geometry optimizer. Default is {}.
+        n_qpoints : int
+            Number of q-points to sample along generated path, including end points.
+            Unused if `paths` is specified. Default is 51.
+        paths : PathLike | None
```
oerc0122, line 252: Maybe `qpoint_path_file` or similar? `paths` has a few too many other meanings.

janus_core/calculations/phonons.py
```diff
@@ -534,17 +584,44 @@ def write_bands(
...
-            write_yaml=True,
-            filename=bands_file,
+        if self.paths:
+            bands_file = self._build_filename("bands.yml.xz", filename=bands_file)
+
+            with open(self.paths, encoding="utf8") as paths_file:
+                paths_info = safe_load(paths_file)
+
+            labels = paths_info["labels"]
+            num_q_points = sum([len(q) for q in paths_info["paths"]])
```
oerc0122, line 594:
```suggestion
            num_q_points = sum(len(q) for q in paths_info["paths"])
```

janus_core/calculations/phonons.py
```diff
@@ -534,17 +584,44 @@ def write_bands(
...
+            num_labels = len(labels)
+            assert (
```
oerc0122, line 596: Small thing of note is that `assert` is disabled if run with `python -O`, while not very common, it's liable to become more common as python has recently implemented bytecode optimisation. Do you want this to be disabled?

tests/test_phonons_cli.py
```diff
@@ -59,7 +60,7 @@ def test_phonons():
 
         has_eigenvectors = False
         has_velocity = False
-        with open(bands_path, encoding="utf8") as file:
```
oerc0122, line 62: If this is a `yaml` file, do we want to just parse it and check the properties?

tests/test_phonons_cli.py
```diff
@@ -179,13 +183,19 @@ def test_dos(tmp_path):
...
+            "{'freq_min': -1, 'freq_max': 0}",
             "--no-hdf5",
             "--file-prefix",
             file_prefix,
         ],
     )
     assert result.exit_code == 0
     assert dos_results.exists()
+    with open(dos_results, encoding="utf8") as file:
+        lines = file.readlines()
```
oerc0122, line 196:
```suggestion
    lines = dos_results.read_text().splitlines()
```

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
...
+            "--file-prefix",
+            file_prefix_2,
+        ],
+    )
+    assert result.exit_code == 0
+
+    # Check parameters
+    with open(displacement_file_1, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_1 = len(params["displacements"])
```
oerc0122, line 504:
```suggestion
        n_displacements_1 = len(params["displacements"])
```

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
...
+
+    with open(displacement_file_2, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_2 = len(params["displacements"])
+
+    assert n_displacments_2 == 2
+
+
+def test_paths(tmp_path):
+    """Test displacment_kwargs can be set."""
```
oerc0122, line 516:
```suggestion
    """Test displacement_kwargs can be set."""
```

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
...
+    # Check parameters
+    with open(displacement_file_1, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_1 = len(params["displacements"])
+
+    assert n_displacments_1 == 4
+
+    with open(displacement_file_2, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_2 = len(params["displacements"])
```
oerc0122, line 510:
```suggestion
        n_displacements_2 = len(params["displacements"])
```

tests/test_phonons_cli.py
ElliottKasoar: I wanted to avoid parsing the full file, since its a huge file (335100 lines), and `eigenvector`/`group_velocity` appear within the first 60 lines.

Originally we did check the yaml, but it slows the tests down quite noticeably, and I don't think I found a way to partially parse yaml particularly nicely.

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
...
+        ],
+    )
+    assert result.exit_code == 0
+
+    # Check parameters
+    with open(displacement_file_1, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_1 = len(params["displacements"])
+
+    assert n_displacments_1 == 4
```
ElliottKasoar, line 506:
```suggestion
    assert n_displacements_1 == 4
```

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
...
+
+    with open(displacement_file_2, encoding="utf8") as file:
+        params = yaml.safe_load(file)
+        n_displacments_2 = len(params["displacements"])
+
+    assert n_displacments_2 == 2
```
ElliottKasoar, line 512:
```suggestion
    assert n_displacements_2 == 2
```

tests/test_phonons_cli.py
```diff
@@ -443,3 +459,83 @@ def test_no_carbon(tmp_path):
     with open(summary_path, encoding="utf8") as file:
         phonon_summary = yaml.safe_load(file)
         assert "emissions" not in phonon_summary
+
+
+def test_displacement_kwargs(tmp_path):
+    """Test displacment_kwargs can be set."""
```
ElliottKasoar, line 465:
```suggestion
    """Test displacement_kwargs can be set."""
```

janus_core/calculations/phonons.py
ElliottKasoar: I think it would just end up throwing a less helpful error, so I think that's fine if someone wants it to be optimised

janus_core/calculations/phonons.py
ElliottKasoar: Good point, that's probably sensible

ElliottKasoar: Do you think `qpoint_file` (`--qpoint-file`) is clear enough? Mostly to avoid long CLI options

oerc0122 (approved):

alinelena (approved):

PR #330, opened 2024-10-14
Add lazy imports for CLI and calculators

### LINKED ISSUE #267: Tab completion slow and raises warnings
Apologies if this was already spotted/fixed, I am using 0.6.2.

The typer command-line interface allows shell completions to be installed for a nicer interactive experience. However, when I try to use these, or run the `--help` command:

a) the response is quite slow, which negates the workflow improvement
b) sometimes minor deprecations warnings are raised by dependencies such as e3nn/pytorch.

This could probably be avoided by delaying the import of these modules so that it is possible for typer to do whatever it wants to do without loading a lot of dependencies.

ElliottKasoar:
Resolves #267

A lot of this has turned out to be similar to #272. I hadn't appreciated that our current structure with individual modules for each command (`janus_core.cli.singlepoint` etc.) meant that we actually already shared a lot of the structure with the solutions I linked, but to get their benefits we still need to move imports into the `@app.command()` functions.

I think moving `from janus_core.calculations.single_point import SinglePoint` etc. into these functions is all we really need for a much more responsive CLI, but I also moved the utils imports to keep the required CLI command modules as clean as possible.

I've then added `TYPE_CHECKING` for the the CLI types and utils, to minimize imports needed for defining the CLI options, without need to worry too much about other parts of `janus_core`.

Finally, I also added `TYPE_CHECKING` for `torch` and `Calculator` in `mlip_calculators.py`. In general, I didn't want to worry about, and I'm not sure there would be huge benefits, to lazy imports throughout the rest of the code base, particularly in `calculations` modules, but in this case we already have imports specific to each MLIP, and `torch` can be particularly slow, so I think it's reasonable.

Note: sphinx doesn't seem to play very well with `TYPE_CHECKING`, so the `nitpick_ignore` has grown quite a lot, but I couldn't find any ways around it for most cases. I think for `Path`, using `pathlib.Path` may be a solution, but I didn't want to define it differently in a single file to everywhere else.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -203,7 +203,9 @@ nitpick_ignore = [
     ("py:class", "Architectures"),
     ("py:class", "Devices"),
     ("py:class", "MaybeSequence"),
+    ("py:class", "PathLike"),
     ("py:class", "Atoms"),
+    ("py:class", "Calculator"),
     ("py:class", "Context"),
     ("py:class", "Path"),
 ]
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -200,4 +200,10 @@ nitpick_ignore = [
     ("py:class", "ase.optimize.optimize.Optimizer"),
     ("py:class", "codecarbon.emissions_tracker.OfflineEmissionsTracker"),
     ("py:class", "_io.StringIO"),
+    ("py:class", "Architectures"),
+    ("py:class", "Devices"),
+    ("py:class", "MaybeSequence"),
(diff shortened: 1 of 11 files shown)


ajjackson:
Indeed this looks like a tidier and more systematic take on my #272 hacking. The latency on tab-completion feels qualitatively the same as that branch: i.e. usable but still slightly annoying. It seems worth having.

oerc0122 (approved):
Looks sensible and if there's improvement (sound like there is) and the tests are passing, looks good to go.

PR #334, opened 2024-10-15
Test mac on self-hosted runner

ElliottKasoar:


.github/workflows/mac.yml
```diff
@@ -0,0 +1,53 @@
+name: mac-ci
+
+on: [push, pull_request]
+
+jobs:
+
+  tests:
+    runs-on: self-hosted
```
ElliottKasoar, line 8:
```suggestion
    runs-on: [self-hosted, macOS]
```

ElliottKasoar:
We should be able to replace the python installation once #358 is merged, which should solve the current errors.

I'll probably stick this in the dev guide at some point, but for future reference, we haven't managed to find a nice way to set up the self-hosted runner as a [service](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/configuring-the-self-hosted-runner-application-as-a-service?platform=mac) (see issues e.g. https://github.com/actions/runner/issues/349, https://github.com/actions/runner/issues/1959, https://github.com/actions/runner/issues/1476), which may partly be because we are unable to set auto login, and as we wanted to avoided running things with `sudo` access.

The solution, which seems to be working for now, is to use the standard `run.sh` script in the background, and disowning the task, so we can log out:

```
nohup ./run.sh &!
```

Sleep options etc. may need revisiting, but the runner has remained online for over an hour.

alinelena (approved):

Issue #240, opened 2024-08-02
support calculator lists


alinelena:
will be nice to be able to support for a structure, structures calculator lists.

eg, I want to have for a system, energies, forces stresses with mace and sevennet
we will need to think of best approach

ElliottKasoar:
We may be able to extend/subclass one of the [mixing calculator](https://wiki.fysik.dtu.dk/ase/ase/calculators/mixing.html) classes, or at least implement something similar to the Mixer base class, without the need for weights.

PR #336, opened 2024-10-22
Save model info

### LINKED ISSUE #327: add model info to extxyz
at the moment we add arch to various which is great will be useful and informative... and necessary to add also the info about the model. maybe add in info['{arch}_{model}']=model (use an absolute path where a path was given)

ElliottKasoar:
Resolves #327

For `model_path` (or equivalent inputs e,g, `model`), this adds `mlip_model` as an info label when structures are output.

For strings and paths, this is fine. For already-loaded torch models, it's less clear what the best label is, so I'm open to alternatives.

This will also interact with any implementation of #240, but we can deal with that when we come to it.

@alinelena, was the suggested `info['{arch}_{model}'] = model` replacing `info["arch"]` or additional? To me it doesn't make sense to have `"{arch}_model"` as well as `"arch"`, which is how I originally interpreted it, but perhaps it makes sense to combine them into a single info label, which I just realised may be what you meant.

### DIFF
--- a/docs/source/user_guide/python.rst
+++ b/docs/source/user_guide/python.rst
@@ -14,7 +14,7 @@ Jupyter Notebook tutorials illustrating the use of currently available calculati
 Calculation outputs
 ===================
 
-By default, calculations performed will modify the underlying `ase.Atoms <https://wiki.fysik.dtu.dk/ase/ase/atoms.html>`_ object
+By default, calculations performed will modify the underlying :class:`ase.Atoms` object
 to store information in the ``Atoms.info`` and ``Atoms.arrays`` dictionaries about the MLIP used.
 
 Additional dictionary keys include ``arch``, corresponding to the MLIP architecture used,
--- a/docs/source/user_guide/python.rst
+++ b/docs/source/user_guide/python.rst
@@ -9,3 +9,51 @@ Jupyter Notebook tutorials illustrating the use of currently available calculati
 - `Molecular Dynamics <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/md.ipynb>`_
 - `Equation of State <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/eos.ipynb>`_
 - `Phonons <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/phonons.ipynb>`_
+
+
+Calculation outputs
+===================
+
+By default, calculations performed will modify the underlying `ase.Atoms <https://wiki.fysik.dtu.dk/ase/ase/atoms.html>`_ object
+to store information in the ``Atoms.info`` and ``Atoms.arrays`` dictionaries about the MLIP used.
+
+Additional dictionary keys include ``arch``, corresponding to the MLIP architecture used,
+and ``model_path``, corresponding to the model path, name or label.
+
+Results from the MLIP calculator, which are typically stored in ``Atoms.calc.results``, will also,
+by default, be copied to these dictionaries, prefixed by the MLIP ``arch``.
+
+For example:
+
+.. code-block:: python
+
+    from janus_core.calculations.single_point import SinglePoint
+
+    single_point = SinglePoint(
+        struct_path="tests/data/NaCl.cif",
+        arch="mace_mp",
+        model_path="tests/models/mace_mp_small.model",
+    )
+
+    single_point.run()
+    print(single_point.struct.info)
+
+will return
+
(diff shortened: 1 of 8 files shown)


janus_core/helpers/utils.py
```diff
@@ -257,6 +257,8 @@ def results_to_info(
     if struct.calc and "arch" in struct.calc.parameters:
         arch = struct.calc.parameters["arch"]
         struct.info["arch"] = arch
+    if struct.calc and "mlip_model" in struct.calc.parameters:
```
ElliottKasoar, line 260:
```suggestion
    if struct.calc and "model" in struct.calc.parameters:
```

janus_core/helpers/utils.py
```diff
@@ -474,6 +476,8 @@ def output_structs(
         for image in images:
             if image.calc and "arch" in image.calc.parameters:
                 image.info["arch"] = image.calc.parameters["arch"]
+            if image.calc and "mlip_model" in image.calc.parameters:
```
ElliottKasoar, line 479:
```suggestion
            if image.calc and "model" in image.calc.parameters:
```

alinelena:
we need to think about this one, what happens if we have two calculations, both arch=mace but model=small and model=large, how will the extxyz look and similarly if we have some models passed by paths(do we get only the basename for mlip?). I think will be nice to document all this structure of what we want and then review the code accordingly

ElliottKasoar:
> we need to think about this one, what happens if we have two calculations, both arch=mace but model=small and model=large, how will the extxyz look and similarly if we have some models passed by paths(do we get only the basename for mlip?). I think will be nice to document all this structure of what we want and then review the code accordingly

Good question. At the moment I think having the same `arch` is relatively clean, if not perfect, since `info["arch]"`, `info["mlip_model"]`, and labelled results should all just be replaced. This means we can't save both sets of results, but at least I don't think it should be a confusing mix of both.

E.g. `arch=mace, mlip_model=small, mace_energy=x` -> `arch=mace, mlip_model=large, mace_energy=y`

When we change architecture, this is a bit more complicated, since we'd change `info["arch]"`, `info["mlip_model"]`, but keep both sets of results.

E.g. `arch=mace_off, mlip_model=small, mace_off_energy=x` -> `arch=mace_mp, mlip_model=large, mace_off_energy=x, mace_mp_energy=y`.

I implemented it this way for now since it means `mlip_model` essentially works the same as `arch` already does - we can only keep the last one, even if we add new results.

We could change both to potentially be lists (if more than one value), either as two individual lists or a combined entry `{arch}_model={model}`.

This does make things more complicated though, since we'd then want to do something like convert repeated `{arch}_{property}` entries e.g. `mace_energy` into a list too, but only if the model has changed...

> similarly if we have some models passed by paths(do we get only the basename for mlip?)

Regarding paths, at the moment model paths are saved as `str(Path(model_path).expanduser())` e.g. `~` expands to `mlip_model=/home/ek/.cache/mace/46jrkm3v`, but you might get `mlip_model=46jrkm3v` for a relative model path. Maybe we should change this to absolute.

alinelena:
ok I think we need to document very well the behaviour of this with examples since there is no obvious solution.

docs/source/user_guide/python.rst
```diff
@@ -9,3 +9,51 @@ Jupyter Notebook tutorials illustrating the use of currently available calculati
 - `Molecular Dynamics <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/md.ipynb>`_
 - `Equation of State <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/eos.ipynb>`_
 - `Phonons <https://colab.research.google.com/github/stfc/janus-tutorials/blob/main/phonons.ipynb>`_
+
+
+Calculation outputs
+===================
+
+By default, calculations performed will modify the underlying `ase.Atoms <https://wiki.fysik.dtu.dk/ase/ase/atoms.html>`_ object
```
oerc0122, line 17: Might this be better as:
```suggestion
By default, calculations performed will modify the underlying :class:`ase.Atoms` object
```
?

tests/test_singlepoint_cli.py
```diff
@@ -54,16 +54,22 @@ def test_singlepoint():
 
     finally:
         # Check atoms can read read, then delete file
```
oerc0122, line 56: "read read"?

alinelena (approved):

PR #341, opened 2024-10-29
Restructure helpers directory

### LINKED ISSUE #144: Split up helpers directory?

ElliottKasoar:
Resolves #144

- Breaks up `helpers` directory into `helpers`, `processing`, and `training`
  - `helpers` is for general utility functions, which almost always only make sense to be called to simplify other functions
  - `processing` is for physics-based utilities e.g. pre/post processing, which generally could be used in isolation
  - `training` is for functions used for ML training, currently only `train.py`, but in future any active learning etc. too

- Breaks up `utils.py`, since it was getting a bit unwieldy
  - The symmetry functions have been moved to their own module in `processing`
  - Functions used for inputting and outputting structures have been moved into their own `helpers` module, `struct_io.py`
  - Functions that only look at the calculator have been moved to `mlip_calculators.py`

I'm very open to further/alternative suggestions.

Note: this may cause conflicts with #285 and #318, and potentially most other PRs.


alinelena (dismissed):

alinelena:
will need a rebase it seems

ElliottKasoar:
Odd, it didn't show any conflicts for me when I rebased locally, but seems ok...

janus_core/helpers/utils.py
```diff
@@ -166,54 +149,6 @@ def _build_filename(
...
-    sym_tolerance : float
-        Atom displacement tolerance for spglib symmetry determination, in Å.
-        Default is 0.001.
-    """
-    refine_symmetry(struct, symprec=sym_tolerance, verbose=False)
-
-
 def none_to_dict(dictionaries: Sequence[Optional[dict]]) -> Generator[dict, None, None]:
     """
     Ensure dictionaries that may be None are dictionaires.
```
harveydevereux, line 154: minor typo

```suggestion
    Ensure dictionaries that may be None are dictionaries.
```

and this below but I don't seem to be able to make a suggestion on it

```python
dictionaries : Sequence[dict]
        Sequence of dictionaries that be be None. # could be None?
```

harveydevereux (approved):
I think this is a good move generally, just spotted some typos while looking. But it looks fine regarding my stuff.

I can't think of how to split utils further without making individual files for those left in there which is not helpful. Maybe in future we'll want a ```file_io``` with perhaps write_table added in there.

janus_core/helpers/utils.py
ElliottKasoar: Thanks!

janus_core/helpers/mlip_calculators.py
```diff
@@ -231,3 +233,33 @@ def choose_calculator(
...
+    attribute : str
+        Attribute to check calculator for.
+    """
+    # If dispersion added to MLIP calculator, use only MLIP calculator for calculation
+    if isinstance(calc, SumCalculator):
+        if (
+            len(calc.mixer.calcs) == 2
+            and calc.mixer.calcs[1].name == "TorchDFTD3Calculator"
+            and hasattr(calc.mixer.calcs[0], attribute)
+        ):
```
oerc0122, line 259: Should these conditions be merged?

janus_core/helpers/struct_io.py
```diff
@@ -0,0 +1,291 @@
...
+        False.
+    logger : Optional[logging.Logger]
+        Logger if log file has been specified. Default is None.
+
+    Returns
+    -------
+    MaybeSequence[Atoms]
+        Structure(s) with attached MLIP calculators.
+    """
+    read_kwargs = read_kwargs if read_kwargs else {}
```
oerc0122, line 152: Was there a function for this?

janus_core/helpers/struct_io.py
```diff
@@ -0,0 +1,291 @@
...
+    # If set_calc not specified, set to True if reading file or calculators not attached
+    if set_calc is None:
+        if struct_path:
+            set_calc = True
+
+        elif isinstance(struct, Atoms):
+            set_calc = struct.calc is None
+
+        elif isinstance(struct, Sequence):
+            set_calc = any(image.calc is None for image in struct)
```
oerc0122, line 198: Optionally?
```python
set_calc = (struct_path or
            (isinstance(struct, Atoms) and struct.calc is None) or
            (isinstance(struct, Sequence) and any(image.calc is None for image in struct)))
```

oerc0122 (approved):
Looks pretty sensible. Just a couple of things we may want to sort while we're here, none of which are important

janus_core/helpers/struct_io.py
ElliottKasoar: We have `none_to_dict`, which iterates over a sequence of dicts, but I'm not sure calling it for one dict would cleaner than this (we'd also need to make it work for non-sequences)

janus_core/helpers/struct_io.py
ElliottKasoar: I think this is probably a bit less readable

janus_core/helpers/mlip_calculators.py
ElliottKasoar: Good point, thanks

harveydevereux (approved):

PR #344, opened 2024-10-31
Add preprocessing training data

### LINKED ISSUE #37: add data reduction to hdf5

ElliottKasoar:
Resolves #37

This probably should should probably be moved to `training` after #341 is merged.

This also currently relies on us passing each config option. I'm hoping we should be able to update this (see [MACE PR](https://github.com/ACEsuit/mace/pull/664)) to allow the config to be passed straight through, as we do for training.

### DIFF
--- a/janus_core/cli/preprocess.py
+++ b/janus_core/cli/preprocess.py
@@ -1,0 +1,74 @@
+# noqa: I002, FA102
+"""Set up MLIP preprocessing commandline interface."""
+
+# Issues with future annotations and typer
+# c.f. https://github.com/maxb2/typer-config/issues/295
+# from __future__ import annotations
+
+from pathlib import Path
+from typing import Annotated
+
+from typer import Option, Typer
+
+app = Typer()
+
+
+@app.command()
+def preprocess(
+    mlip_config: Annotated[
+        Path, Option(help="Configuration file to pass to MLIP CLI.")
+    ],
+    log: Annotated[Path, Option(help="Path to save logs to.")] = Path(
+        "preprocess-log.yml"
+    ),
+    tracker: Annotated[
+        bool, Option(help="Whether to save carbon emissions of calculation")
+    ] = True,
+    summary: Annotated[
+        Path,
+        Option(
+            help=(
+                "Path to save summary of inputs, start/end time, and carbon emissions."
+            )
+        ),
+    ] = Path("preprocess-summary.yml"),
+):
+    """
+    Convert training data to hdf5 by passing a configuration file to the MLIP's CLI.
+
+    Parameters
+    ----------
+    mlip_config : Path
+        Configuration file to pass to MLIP CLI.
+    log : Optional[Path]
+        Path to write logs to. Default is Path("preprocess-log.yml").
+    tracker : bool
+        Whether to save carbon emissions of calculation in log file and summary.
+        Default is True.
+    summary : Optional[Path]
+        Path to save summary of inputs, start/end time, and carbon emissions. Default
+        is Path("preprocess-summary.yml").
+    """
+    from janus_core.cli.utils import carbon_summary, end_summary, start_summary
+    from janus_core.training.preprocess import preprocess as run_preprocess
+
+    inputs = {"mlip_config": str(mlip_config)}
+
+    # Save summary information before preprocessing begins
+    start_summary(command="preprocess", summary=summary, inputs=inputs)
+
+    log_kwargs = {"filemode": "w"}
+    if log:
+        log_kwargs["filename"] = log
+
+    # Run preprocessing
+    run_preprocess(
+        mlip_config, attach_logger=True, log_kwargs=log_kwargs, track_carbon=tracker
+    )
+
+    # Save carbon summary
+    if tracker:
+        carbon_summary(summary=summary, log=log)
+
+    # Save time after preprocessing has finished
+    end_summary(summary)
--- a/janus_core/helpers/utils.py
+++ b/janus_core/helpers/utils.py
@@ -428,7 +428,7 @@ def check_files_exist(config: dict, req_file_keys: Sequence[PathLike]) -> None:
         If a key from `req_file_keys` is in the configuration file, but the
         file corresponding to the configuration value do not exist.
     """
-    for file_key in req_file_keys:
+    for file_key in config.keys() & req_file_keys:
         # Only check if file key is in the configuration file
-        if file_key in config and not Path(config[file_key]).exists():
+        if not Path(config[file_key]).exists():
             raise FileNotFoundError(f"{config[file_key]} does not exist")
--- a/janus_core/helpers/utils.py
+++ b/janus_core/helpers/utils.py
@@ -409,3 +409,26 @@ def track_progress(sequence: Sequence | Iterable, description: str) -> Iterable:
 
     with progress:
         yield from progress.track(sequence, description=description)
+
+
+def check_files_exist(config: dict, req_file_keys: Sequence[PathLike]) -> None:
+    """
+    Check files specified in a dictionary read from a configuration file exist.
+
+    Parameters
+    ----------
+    config : dict
+        Dictionary read from configuration file.
+    req_file_keys : Sequence[Pathlike]
+        Files that must exist if defined in the configuration file.
+
+    Raises
+    ------
+    FileNotFoundError
+        If a key from `req_file_keys` is in the configuration file, but the
+        file corresponding to the configuration value do not exist.
+    """
+    for file_key in req_file_keys:
+        # Only check if file key is in the configuration file
+        if file_key in config and not Path(config[file_key]).exists():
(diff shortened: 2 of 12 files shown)


janus_core/cli/preprocess.py
```diff
@@ -0,0 +1,69 @@
...
+        Path to save summary of inputs, start/end time, and carbon emissions. Default
+        is Path("preprocess-summary.yml").
+    """
+    from janus_core.cli.utils import carbon_summary, end_summary, start_summary
+    from janus_core.training.preprocess import preprocess as run_preprocess
+
+    inputs = {"mlip_config": str(mlip_config)}
+
+    # Save summary information before preprocessing begins
+    start_summary(command="preprocess", summary=summary, inputs=inputs)
```
oerc0122, line 53: Because it has an entry/exit like structure, not sure whether we might want to restructure this as a context manager if we haven't tried already.

Not for here, though.

janus_core/helpers/utils.py
```diff
@@ -407,3 +407,26 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
 
     with progress:
         yield from progress.track(sequence, description=description)
+
+
+def check_files_exist(config: dict, req_file_keys: Sequence[PathLike]) -> None:
```
oerc0122, line 412: Is this a more generally useful function? In which case we may want to consider the following:

- Do we want this to `raise` or would it be better to return a bool/set of missing and allow functions to determine whether to raise?
- Missing `file_key`s may also be a sign of problems. Does that want a bool to handle them?

janus_core/helpers/utils.py
```diff
@@ -407,3 +407,26 @@ def track_progress(sequence: Union[Sequence, Iterable], description: str) -> Ite
...
+
+    Raises
+    ------
+    FileNotFoundError
+        If a key from `req_file_keys` is in the configuration file, but the
+        file corresponding to the configuration value do not exist.
+    """
+    for file_key in req_file_keys:
+        # Only check if file key is in the configuration file
+        if file_key in config and not Path(config[file_key]).exists():
```
oerc0122, line 431:
```suggestion
    for file_key in config.keys() & req_file_keys:
        # Only check if file key is in the configuration file
        if not Path(config[file_key]).exists():
```
Would also work, not sure if it's actually cleaner though. Magic sometimes isn't.

janus_core/training/preprocess.py
```diff
@@ -0,0 +1,88 @@
...
+    attach_logger : bool
+        Whether to attach a logger. Default is False.
+    log_kwargs : Optional[dict[str, Any]]
+        Keyword arguments to pass to `config_logger`. Default is {}.
+    track_carbon : bool
+        Whether to track carbon emissions of calculation. Default is True.
+    tracker_kwargs : Optional[dict[str, Any]]
+        Keyword arguments to pass to `config_tracker`. Default is {}.
+    """
+    (log_kwargs, tracker_kwargs) = none_to_dict((log_kwargs, tracker_kwargs))
```
oerc0122, line 45:
```suggestion
    log_kwargs, tracker_kwargs = none_to_dict((log_kwargs, tracker_kwargs))
```
Don't strictly need these for unpacking

janus_core/training/preprocess.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if req_file_keys is None:
+        req_file_keys = ("train_file", "test_file", "valid_file")
```
oerc0122, line 48: Why not just have this as default argument?

oerc0122 (commented):
Looks good. 

Suggested a couple of possible extensions/simplifications which may not be fore this PR.

janus_core/training/preprocess.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if logger and "foundation_model" in options:
+        logger.info("Fine tuning model: %s", options["foundation_model"])
+
+    # Parse options from config, as MACE cannot read config file yet
+    args = []
+    for key, value in options.items():
+        if isinstance(value, bool):
+            if value is True:
+                args.append(f"--{key}")
```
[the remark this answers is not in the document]
ElliottKasoar: Doesn't this then append `--{key} False`, which we don't want?

janus_core/training/preprocess.py
```diff
@@ -0,0 +1,88 @@
...
+        log_kwargs.setdefault("filename", "preprocess-log.yml")
+    log_kwargs.setdefault("name", __name__)
+    logger = config_logger(**log_kwargs)
+    tracker = config_tracker(logger, track_carbon, **tracker_kwargs)
+
+    if logger and "foundation_model" in options:
+        logger.info("Fine tuning model: %s", options["foundation_model"])
+
+    # Parse options from config, as MACE cannot read config file yet
+    args = []
```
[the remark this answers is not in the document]
ElliottKasoar: I haven't used it, but I'm not sure defining a syntax just for this, which is hopefully only temporary itself, is worth it?

janus_core/training/preprocess.py
ElliottKasoar: Good point. I think this stems from the training equivalent being a list originally (for no clear reason, which is why I changed that too)

janus_core/training/preprocess.py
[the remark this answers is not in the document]
oerc0122: It shouldn't because if `value` is `False` it won't pass the `if`

janus_core/training/preprocess.py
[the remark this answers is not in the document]
oerc0122: If you build the string, e.g
`mace build --proc --arg a b --flag`
`shlex.split` should turn that into the appropriate tokens.

janus_core/helpers/utils.py
ElliottKasoar: I can imagine it could be, but at the moment I don't think we do anything quite like this elsewhere at the moment.

Here, and in `train`, we're only really passing the config files on to MACE, and so this is just an extra protection to handle errors nicely. From what I've seen, missing keys are handled clearly enough by their CLI, so I don't think it's needed yet.

In most other cases, the config is completely optional, and we (should) already have relevant error handling dealt with, either in the Python calculation (e.g. missing restart files for MD), or by typer (e.g. `--struct` being required).

I think for now it's probably fine, until we find uses for it elsewhere, but it's definitely useful to consider.

janus_core/training/preprocess.py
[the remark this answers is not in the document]
ElliottKasoar: Yes, I think I get `janus: error: unrecognized arguments: False`, if I add this.

janus_core/training/preprocess.py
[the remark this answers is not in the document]
ElliottKasoar: Oh, but that doesn't help with building the string from yaml config, does it?

For context, eventually we can hopefully do the same thing we do for training, which is just pass the config straight through

```python
mlip_args = mace_parser().parse_args(["--config", str(mlip_config)])
```

but for now, building the string from the parsed config is still needed.

What's the benefit in using `shlex` over their `parse_args` (which has some additional handling for unknown args etc.) to then tokenise this?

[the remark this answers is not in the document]
oerc0122: It's mostly for when you need to pass a string directly to a call to `subprocess.run` or similar.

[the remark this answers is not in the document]
ElliottKasoar: Ah ok, we don't do that ourselves, we just pass the args back to MACE's run preprocess/training, so we probably wouldn't benefit?

oerc0122 (approved):

ElliottKasoar:
(Rebased and updated use of `none_to_dict` in preprocess as well)

tests/test_preprocess_cli.py
```diff
@@ -0,0 +1,207 @@
...
+def test_help():
+    """Test calling `janus preprocess --help`."""
+    result = runner.invoke(app, ["preprocess", "--help"])
+    assert result.exit_code == 0
+    assert "Usage: janus preprocess [OPTIONS]" in strip_ansi_codes(result.stdout)
+
+
+def test_preprocess(tmp_path):
+    """Test MLIP preprocessing."""
+    train_path = Path("train")
```
alinelena, line 61: these are hdf5 files, maybe stuck an .h5 as extension not to confuse the users?

ElliottKasoar: This is referring to the directory they're stored in e.g.

```
ls processed_data/
statistics.json  test  train  val

ls processed_data/test/

ls processed_data/train/
train_0.h5  train_1.h5  train_2.h5  train_3.h5

ls processed_data/val/
val_0.h5  val_1.h5  val_2.h5  val_3.h5
```

alinelena: my bad misread it..

PR #349, opened 2024-11-08
Slight change to none_to_dict to avoid nesting brackets.

oerc0122:
Also use for single instances of `dict` `None`-ness check.


ElliottKasoar (approved):
Looks good, thanks!

PR #352, opened 2024-11-15
Update MACE

### LINKED ISSUE #335: Model loading
Placeholder for issue reported by @alinelena 

Issue relates to passing a string e.g. `"medium"` for `"mace_off"`, which janus apparently tries to load as a path, rather than using the label.

ElliottKasoar:
Resolves #335

We should be able to update MACE, which has quite a few new features since 0.3.6.

In terms of dependencies, everything seems ok, ~but there seems to be an issue with loading the model, which I haven't looked into, and the stress for molecules has changed slightly (see [issue](https://github.com/ACEsuit/mace/issues/700))~

- The model loading error was because we never actually had MACE-OFF saved, but as MACE was previously ignoring the path, this wasn't visible
- It might be easiest to remove the stress check for molecules, but I think it's useful to have some indication of changes to MACE's outputs, even if the output itself ought to be discarded in general, so I've added a relatively small allowed absolute error

Still to do:
- [x] Fix training bug

This hopefully also fixes an issue reported by @alinelena, along the lines of `TypeError: object of type 'Loss' has no len()` during training, due to MACE outputting the [loss object explicitly](https://github.com/ACEsuit/mace/blob/main/mace/cli/run_train.py#L550).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 *.png
 *.svg
 *.xz
+*.txt
 ~*
 *~
 .project
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Tools for machine learnt interatomic potentials
 
 - Python >= 3.9
 - ASE >= 3.23
-- mace-torch = 0.3.6
+- mace-torch = 0.3.8
 - chgnet = 0.3.8 (optional)
 - matgl = 1.1.3 (optional)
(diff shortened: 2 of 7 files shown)


alinelena:
for non periodic is better to ignores the stress or to tests it is less than single precision noise.

alinelena (dismissed):

alinelena (approved):

PR #358, opened 2024-11-27
Change poetry to uv

ElliottKasoar:
Initial tests at migrating from poetry to uv.

- Updates pyproject.toml
- Updates CI and docs workflows (both seem to work from tests on my fork)

To do:

- [x]  Update publish workflow
- [x] Update/remove tox file (uv can do something similar, I think, but can also work with tox)
- [x] Update dev guide
- [x] Test conflicting dependencies
  - I don't think it should be part of this PR, assuming it's merged, but we should at least see how practical it is to update torch, for example, and leave matgl as a conflict


ElliottKasoar:
I think in terms of conflicts we'll have to think quite carefully about how/when we declare things as conflicting. I think it will be useful, but as this part of uv is so new, there remain a few bugs e.g. https://github.com/astral-sh/uv/issues/9622 and (until earlier today) https://github.com/astral-sh/uv/issues/9289.

For example, I'm not sure we can have things that are "conflicting" in `all`, and we may need to move `mace` from a full dependency to optional, in order to allow it to conflict.

The main other issue I've encountered is that I think `tox` non-pytest sections currently use your current Python version, rather than 3.11 it used previously, but the tests do seem to respect the matrix, which is the more important thing.

alinelena (approved):

Issue #357, opened 2024-11-25
Add description to observables code


ElliottKasoar:
Some parts of the observables/correlation code, e.g. `Correlator._propagate`, have become relatively complex, with a lot of undefined attributes, making understanding the full details of what's happening slightly opaque.

Adding a few comments would be useful.

It might also help to document the attributes, although most of them aren't public so we wouldn't normally. It's probably worth seeing if comments are sufficient first.

PR #369, opened 2024-12-12
Comment correlation algorithm, remove flattening of Velocity observable

harveydevereux:
- Addresses #357 
- Also removes the flattening in ```Velocity``` to preserve dimension for general use cases.


ElliottKasoar (commented):
Minor comment, but otherwise I think this looks good!

We probably want to change the title slightly to reflect the code changes, since it's not just changes to the docs.

alinelena (dismissed):

ElliottKasoar:
Looks good, thanks @harveydevereux!

PR #371, opened 2024-12-13
Improve logging and carbon tracking consistency

### LINKED ISSUE #368: Check if tracker initialised correctly
Similar to the discussions in #316 and #343, there can be issues in setting up the carbon tracker (e.g. NVML library mismatches), which mean the tracker is not set up correctly, but this does not raise an error until attempting to start a task.

We should check the tracker is set up, and if not, raise an error informing users of this, so they can either resolve the issue, or disable tracking.

ElliottKasoar:
Resolves #368

Also resolves various issues relating to logging/tracking, including:

- Setting `attach_logger` default based on whether filename is in `log_kwargs`, which we effectively already did, but did not say we were doing, which is confusing
- Require logging for carbon tracking, which (again) we essentially already did
- Allow kwargs to be passed to carbon tracking, which allows default power etc. to be passed (this may soon provide another alternative to bypass MacOS sudo requests, although not yet)
- Add troubleshooting section for carbon tracking issues
- Raise an error after carbon tracker initialisation if errors occurred, as otherwise it doesn't raise until the calculation is run

### DIFF
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -47,11 +47,12 @@ class BaseCalculation(FileNameMixin):
         Keyword arguments to pass to the selected calculator. Default is {}.
     set_calc : bool | None
         Whether to set (new) calculators for structures. Default is None.
-    attach_logger : bool
-        Whether to attach a logger. Default is False.
+    attach_logger : bool | None
+        Whether to attach a logger. Default is True if "filename" is passed in
+        log_kwargs, else False.
     log_kwargs : dict[str, Any] | None
             Keyword arguments to pass to `config_logger`. Default is {}.
-    track_carbon : bool
+    track_carbon : bool | None
         Whether to track carbon emissions of calculation. Requires attach_logger.
         Default is True if attach_logger is True, else False.
     tracker_kwargs : dict[str, Any] | None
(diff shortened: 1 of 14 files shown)


janus_core/calculations/base.py
```diff
@@ -146,12 +150,23 @@ def __init__(
...
+        else:
+            attach_logger = attach_logger if attach_logger else False
+
+        if not attach_logger:
+            if track_carbon:
+                raise ValueError("Carbon tracking requires logging to be enabled")
+            self.track_carbon = False
+        else:
+            self.track_carbon = track_carbon if track_carbon is not None else True
+
```
harveydevereux, line 169: Generally looks good to me.

Is there a utility we could write to handle setting/checking ```attach_logger``` and ```self.track_carbon``` (or ```track_carbon```) in ```training/preprocess```, ```training/train```, and ```BaseCalculation```? Or is that more trouble than its worth.

ElliottKasoar: Good point, the training ones are slightly different since they don't operate on `self.logger`, but I can probably pass them through to make them consistent, which would probably be slightly neater.

alinelena (approved):

oerc0122 (approved):
Looks good. 

Only possible comment would be whether we want to move `codecarbon` to an optional dependency rather than a mandatory one and disable it if it's not present now that it can be disabled (with suitable error), but this is probably another issue/PR.

Issue #374, opened 2025-01-03
Drop Python 3.9 support?


ElliottKasoar:
This has been discussed at various stages (e.g. #303, #312, #328, #372), but at some point we will drop support for Python 3.9.

My general feeling was to roughly stick to the [Python release cycle](https://devguide.python.org/versions/), or at least wait until we can properly support 3.13, but some packages (roughly) follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which has already dropped 3.9.

In particular, this seems to be the case for Materials Project packages e.g. [pymatgen](https://github.com/materialsproject/pymatgen/pull/4009), and projects which interact with it e.g. [CHGNet](https://github.com/CederGroupHub/chgnet/pull/204).

It's worth noting that PyTorch [decided against dropping 3.9 yet](https://github.com/pytorch/pytorch/issues/129805), but perhaps more practically, the question is how much we lose from dropping support now? My impression is that most HPC users are required to to install a new version of Python anyway, and I would imagine locally using a different version is less of an issue.

The issue discussed in #347 may also be fixable.

PR #376, opened 2025-01-06
Implement temporary fix for CI

### LINKED ISSUE #372: DeprecationWarning raised in CI
Currently (re-)running the CI throws an error, due to a deprecation warning when loading CHGNet (see also: https://github.com/materialsproject/pymatgen/issues/4243):

```
==================================== ERRORS ====================================
_______________ ERROR collecting tests/test_mlip_calculators.py ________________
tests/test_mlip_calculators.py:8: in <module>
    from chgnet.model.model import CHGNet
.venv/lib/python3.10/site-packages/chgnet/model/__init__.py:3: in <module>
    from chgnet.model.dynamics import CHGNetCalculator, MolecularDynamics, StructOptimizer
.venv/lib/python3.10/site-packages/chgnet/model/dynamics.py:24: in <module>
    from pymatgen.analysis.eos import BirchMurnaghan
.venv/lib/python3.10/site-packages/pymatgen/analysis/eos.py:23: in <module>
    from pymatgen.core.units import FloatWithUnit
.venv/lib/python3.10/site-packages/pymatgen/core/__init__.py:12: in <module>
    from pymatgen.core.composition import Composition
.venv/lib/python3.10/site-packages/pymatgen/core/composition.py:22: in <module>
    from pymatgen.core.periodic_table import DummySpecies, Element, ElementType, Species, get_el_sp
.venv/lib/python3.10/site-packages/pymatgen/core/periodic_table.py:63: in <module>
    class ElementBase(Enum):
.venv/lib/python3.10/site-packages/pymatgen/core/periodic_table.py:755: in ElementBase
    @deprecated(is_rare_earth, message="is_rare_earth is corrected to include Y and Sc.", deadline=(2025, 1, 1))
.venv/lib/python3.10/site-packages/monty/dev.py:157: in deprecated
    raise_deadline_warning()
.venv/lib/python3.10/site-packages/monty/dev.py:86: in raise_deadline_warning
    raise DeprecationWarning(
E   DeprecationWarning: This function should have been removed on 2025-01-01.

---------- coverage: platform linux, python 3.10.16-final-0 ----------
Coverage XML written to file coverage.xml

=========================== short test summary info ============================
ERROR tests/test_mlip_calculators.py - DeprecationWarning: This function should have been removed on 2025-01-01.
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 4.56s ===============================
```

Since the warning is generated by `raise`, it seems that `--disable-warnings` and other pytest filters don't suppress this, so I'm not sure if there's a simple workaround?

One option could be to alter he environment variables that are checked before the warning is raised  e.g. GITHUB_REPOSITORY/CI - see [CI/repository checks](https://github.com/materialsvirtuallab/monty/blob/master/src/monty/dev.py#L51), but that may have unexpected side effects, so I don't think it's a good solution.

Hopefully this can be removed from `pymatgen` relatively quickly, but unfortunately they only support Python 3.10+, so we may need to reconsider 3.9 support to make use of any fixes.

ElliottKasoar:
Resolves #372

See also, discussion in `pymatgen`: https://github.com/materialsproject/pymatgen/issues/4243

I think unsetting `CI` is the only solution, as I think otherwise `os.getenv("CI") is not None` evaluates to `True`, even if `CI` is set to `""`, `False`, etc.

This change should be reverted once #374 is implemented, as we should be able to update monty to include the fix: https://github.com/materialsvirtuallab/monty/pull/735.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,9 @@ jobs:
         env:
           # show timings of tests
           PYTEST_ADDOPTS: "--durations=0"
-        run: uv run pytest --run-extra-mlips --cov janus_core --cov-append .
+        run: |
+          unset CI
+          uv run pytest --run-extra-mlips --cov janus_core --cov-append .
 
       - name: Report coverage to Coveralls
         uses: coverallsapp/github-action@v2


DanielYang59:
> I think unsetting CI is the only solution, as I think otherwise os.getenv("CI") is not None evaluates to True, even if CI is set to "", False, etc.

AFAIK, to unset an env var (or any other value in yaml in general), you would need to set it to `null` instead of some other falsy string like `""`, `"false"` and so on. Let me know if this doesn't work

ElliottKasoar:
> > I think unsetting CI is the only solution, as I think otherwise os.getenv("CI") is not None evaluates to True, even if CI is set to "", False, etc.
> 
> AFAIK, to unset an env var (or any other value in yaml in general), you would need to set it to `null` instead of some other falsy string like `""`, `"false"` and so on (because our code doesn't check the "semantic meaning" of the string, but it's presence). Let me know if this doesn't work

I think setting it to `null` would work it it were directly being read into the Python, but as this is setting the environment variable itself, I think this still ends up roughly equivalent to `""` - see [failed test](https://github.com/ElliottKasoar/janus-core/actions/runs/12636360460/workflow) and [debugging test](https://github.com/ElliottKasoar/janus-core/actions/runs/12636887293/job/35209973448), where I echo `$CI` and print `os.getenv("CI")`/`os.getenv("CI") is not None`.

(I went through a similar realisation about what `monty` was doing, initially assuming setting CI to `False` would would work, before realising it needed to evaluate to `None`.)

oerc0122:
The real problem is that the code is actually checking `if "CI" in os.environ` rather than checking `if os.getenv("CI").lower() in ("true", "1")` which would be more useful.

DanielYang59:
> The real problem is that the code is actually checking if "CI" in os.environ rather than checking if os.getenv("CI").lower() in ("true", "1") which would be more useful.

Yep that was the original implementation (and now we decided to drop this warning/exception completely). This was intended as we anticipate either one of the two situations:
- Either running in Github Actions, [`CI` would always be set to `true`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)
- If not, `CI` would be absent

We cannot think of too many cases where the user would need to change the value of `CI`, therefore we only checked if it's present

alinelena (approved):

PR #377, opened 2025-01-07
Drop Python 3.9 support

### LINKED ISSUE #374: Drop Python 3.9 support?

ElliottKasoar:
Resolves #374

As I mention in the issue, we should think carefully about when we want to merge this, but I wanted to check it's also a viable solution to our CI problems, and these changes shouldn't clash too much with ongoing work that may be merged before this.

- Drops Python 3.9 support, remove Python 3.9 from testing matrices, and apply fixes to satisfy `ruff`
- Adds in missing `from __future__ import annotations`, which relied on a fix only available for Python 3.10+

Note: if merged after #376, those changes can be reverted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.9','3.10','3.11','3.12']
+        python-version: ["3.10","3.11","3.12"]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,7 @@ jobs:
         env:
           # show timings of tests
           PYTEST_ADDOPTS: "--durations=0"
-        run: |
-          unset CI
-          uv run pytest --run-extra-mlips --cov janus_core --cov-append .
+        run: uv run pytest --run-extra-mlips --cov janus_core --cov-append .
 
       - name: Report coverage to Coveralls
         uses: coverallsapp/github-action@v2
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -10,7 +10,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python-version: ['3.9','3.10','3.11','3.12']
+        python-version: ["3.10","3.11","3.12"]
 
     steps:
       - uses: actions/checkout@v4
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Tools for machine learnt interatomic potentials
 
 `janus-core` dependencies currently include:
 
-- Python >= 3.9
+- Python >= 3.10
 - ASE >= 3.23
 - mace-torch = 0.3.8
 - chgnet = 0.3.8 (optional)
--- a/docs/source/getting_started/getting_started.rst
+++ b/docs/source/getting_started/getting_started.rst
@@ -7,7 +7,7 @@ Dependencies
 
 ``janus-core`` dependencies currently include:
 
-- Python >= 3.9
+- Python >= 3.10
 - ASE >= 3.23
 - mace-torch = 0.3.8
 - chgnet = 0.3.8 (optional)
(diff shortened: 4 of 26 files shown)


alinelena:
personally I think we drop 3.9 we add 3.13...

ElliottKasoar:
> personally I think we drop 3.9 we add 3.13...

It would be nice (and was my original intent) to add 3.13 when we drop 3.9, but it depends in what sense we support 3.13. PyTorch won't fully support it for MacOS/Windows until at least [PyTorch 2.6](https://github.com/pytorch/pytorch/issues/130249), so we wouldn't be able to fully support it yet in that sense.

I also need to finally decide on a solution for matgl (M3GNet) to be able to move on from PyTorch 2.2, otherwise we definitely can't.

oerc0122:
> > personally I think we drop 3.9 we add 3.13...
> 
> It would be nice (and was my original intent) to add 3.13 when we drop 3.9, but it depends in what sense we support 3.13. PyTorch won't fully support it for MacOS/Windows until at least [PyTorch 2.6](https://github.com/pytorch/pytorch/issues/130249), so we wouldn't be able to fully support it yet in that sense.
> 
> I also need to finally decide on a solution for matgl (M3GNet) to be able to move on from PyTorch 2.2, otherwise we definitely can't.

Anyone running 3.13 must also run PyTorch 2.6+ so I'm not sure that's a problem as that's with their environment and should be effectively handled by pip and PyTorch's `minimum_python`. Provided PyTorch 2.6 is compatible with other things, it's not a problem.

ElliottKasoar:
> > > personally I think we drop 3.9 we add 3.13...
> > 
> > 
> > It would be nice (and was my original intent) to add 3.13 when we drop 3.9, but it depends in what sense we support 3.13. PyTorch won't fully support it for MacOS/Windows until at least [PyTorch 2.6](https://github.com/pytorch/pytorch/issues/130249), so we wouldn't be able to fully support it yet in that sense.
> > I also need to finally decide on a solution for matgl (M3GNet) to be able to move on from PyTorch 2.2, otherwise we definitely can't.
> 
> Anyone running 3.13 must also run PyTorch 2.6+ so I'm not sure that's a problem as that's with their environment and should be effectively handled by pip and PyTorch's `minimum_python`. Provided PyTorch 2.6 is compatible with other things, it's not a problem.

I think about this slightly different to most dependency resolution issues, in that I'd like our tests to run 3.13 if we support it, whereas for other things that pip/uv resolves, that isn't something we need to worry about.

Since testing on MacOS isn't merged yet (#334), and that has its own Python matrix anyway (I imagine Windows will be similar), this isn't necessarily a problem, but `"torch<=2.2,>=2.1"` in our pyproject.toml for `matgl` is blocking.

ElliottKasoar:
It seems like the latest version of MACE is also no longer compatible with 3.9: https://github.com/ACEsuit/mace/issues/699, in a way that `pip`/`uv` are unaware of, so I think we probably do need to drop 3.9 now, and hopefully we can add 3.13 as soon as possible.

oerc0122 (dismissed):
All looks good!

ElliottKasoar:
> Ready to merge?

I'm hoping to tidy up and merge #369 and #378 first, and do a final release supporting 3.9 first, so hopefully this week, but probably not today.

How does it impact you, if you don't mind me asking?

ElliottKasoar:
Rebased, and applied (minimal) ruff fixes to newly added code.

PR #378, opened 2025-01-07
Add units

### LINKED ISSUE #131: units
at the moment we use default ase units. will be a good idea not to depend on whatever is ase default codata but to fix them to a certain set. example in ase documentation.

ElliottKasoar:
Resolves #131 and #277

- Creates dictionary of units shared by all calculations, to save relevant items when writing out files
- Renames `time_fs` to `time`, as this is labelled now by the units.
- Pins default [ASE units](https://wiki.fysik.dtu.dk/ase/ase/units.html) to 2014, matching current default

One slightly messy aspect is this effectively adds a second, slightly different version of the [MD units dictionary](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L789). These should probably be combined.

~As with all current PRs, this depends on #376 or #377 for tests to pass~

### DIFF
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -216,7 +216,7 @@ class BaseCalculation(FileNameMixin):
             self.logger, self.track_carbon, **self.tracker_kwargs
         )
 
-    def _set_units(self, keys: Sequence = ("energy", "forces", "stress")) -> None:
+    def _set_units(self, keys: Sequence[str] = ("energy", "forces", "stress")) -> None:
         """
         Save units to structure info.
 
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -22,6 +22,7 @@ UNITS = {
     "energy": "eV",
     "forces": "ev/Ang",
     "stress": "ev/Ang^3",
+    "hessian": "ev/Ang^2",
     "time": "fs",
     "real_time": "s",
     "temperature": "K",
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -27,7 +27,8 @@ UNITS = {
     "real_time": "s",
     "temperature": "K",
     "pressure": "GPa",
-    "momenta": "(eV / u)^0.5",
+    "momenta": "(eV/u)^0.5",
+    "density": "g/cm^3",
 }
 
 
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -27,7 +27,7 @@ UNITS = {
     "real_time": "s",
     "temperature": "K",
     "pressure": "GPa",
-    "momenta": "(eV/u)^0.5",
+    "momenta": "(eV*u)^0.5",
     "density": "g/cm^3",
 }
 
--- a/janus_core/calculations/base.py
+++ b/janus_core/calculations/base.py
@@ -29,6 +29,7 @@ UNITS = {
     "pressure": "GPa",
     "momenta": "(eV*u)^0.5",
     "density": "g/cm^3",
+    "volume": "Ang^3",
 }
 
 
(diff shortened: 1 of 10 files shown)


tests/test_md_cli.py
```diff
@@ -108,6 +108,23 @@ def test_md(ensemble):
...
 
+        expected_units = {
+            "time": "fs",
+            "real_time": "s",
+            "energy": "eV",
+            "forces": "ev/Ang",
+            "stress": "ev/Ang^3",
+            "temperature": "K",
+            "density": "g/cm^3",
+            "momenta": "(eV/u)^0.5",
```
ElliottKasoar, line 119:
```suggestion
            "momenta": "(eV*u)^0.5",
```

janus_core/calculations/base.py
```diff
@@ -18,6 +18,19 @@
...
+UNITS = {
+    "energy": "eV",
+    "forces": "ev/Ang",
+    "stress": "ev/Ang^3",
+    "hessian": "ev/Ang^2",
+    "time": "fs",
+    "real_time": "s",
+    "temperature": "K",
+    "pressure": "GPa",
+    "momenta": "(eV/u)^0.5",
```
ElliottKasoar, line 30:
```suggestion
    "momenta": "(eV*u)^0.5",
```

oerc0122 (dismissed):
As always, I approve of adding proper units. One consideration is whether we want to make an issue and add support for [`pint`](https://pint.readthedocs.io/en/stable/) to handle the units for us and allow users to dump them however they want. 

This might be a major piece of work though and again quite pervasive throughout the code. Definitely another PR, but this is a positive change!

ElliottKasoar:
> One slightly messy aspect is this effectively adds a second, slightly different version of the [MD units dictionary](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L789). These should probably be combined.

The stats units now use the units dictionary in `Base`, which means they should all be consistent now (e.g. Angstrom was referred to as `A` in stats, but `Ang` otherwise).

janus_core/calculations/md.py
```diff
@@ -43,6 +44,7 @@
 from janus_core.processing.correlator import Correlation
 from janus_core.processing.post_process import compute_rdf, compute_vaf
 
+units = create_units("2014")
```
oerc0122, line 47: While "consistent", it's slightly confusing to have `UNITS` and `units` simultaneously. Also, not sure what the difference between `create_units` (from ASE) and `_set_units` is at a glance.

oerc0122 (dismissed):
Looks good.

Only other consideration I can think of (having written a unit-handler again recently), do you think Janus will ever need to offer alternative unit schemes or is that someone else's problem?

janus_core/calculations/md.py
oerc0122: What does `set_info_units` mean? `get_unit_names`? `UNITS` -> `JANUS_UNITS` and/or `units` -> `ase_units` works, but should be done through qualified import (i.e. `import X as ...`)

ElliottKasoar: It's not just getting units, it's setting something like `struct.info["units"]`, so it is literally setting the units, to the info dictionary.

As in, `import UNITS as JANUS_UNITS`? Why not just rename `UNITS` in `calculations/base`, where it's defined?

oerc0122: Ok, that answers the question. I'm happy with `_set_info_units`.

The reason I said that is mostly because `UNITS` is not a generally a bad name for import, it's only because we have a conflicting name from ASE that it's a problem.

oerc0122:
> > Looks good.
> > Only other consideration I can think of (having written a unit-handler again recently), do you think Janus will ever need to offer alternative unit schemes or is that someone else's problem?
> 
> I can imagine it as an eventuality, but I don't see it happening particularly soon. Perhaps some sort of conversion tool for input/output files would be a first step anyway.

That boils down to "someone else's problem" for now. Happy with that.

oerc0122 (approved):

PR #379, opened 2025-01-13
Add installation troubleshooting

### LINKED ISSUE #373: Document phonopy cmake error troubleshooting
When installing, I (and others) have encountered errors along the lines of:

```
CMake Error at /tmp/pip-build-env-j2jx9pp5/overlay/lib/python3.11/site-packages/nanobind/cmake/nanobind-config.cmake:243 (target_compile_features):
    target_compile_features The compiler feature "cxx_std_17" is not known to
    CXX compiler

    "GNU"

    version 4.8.5.
  Call Stack (most recent call first):
    /tmp/pip-build-env-j2jx9pp5/overlay/lib/python3.11/site-packages/nanobind/cmake/nanobind-config.cmake:358 (nanobind_build_library)
    CMakeLists.txt:108 (nanobind_add_module)


  -- Configuring incomplete, errors occurred!

  *** CMake configuration failed
  [end of output]

note: This error originates from a subprocess, and is likely not a problem with pip.
ERROR: Failed building wheel for phonopy
Building wheel for python-hostlist (setup.py) ... done
Created wheel for python-hostlist: filename=python_hostlist-2.2.1-py3-none-any.whl size=39604 sha256=44f9f27a42895e61a521cf9129a6a3ad03e633b201390da5ef76d5f59db3b94f
Stored in directory: ...
Successfully built janus-core python-hostlist
Failed to build phonopy
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (phonopy)
```

This isn't an issue with janus as such, but since users may be less familiar with cmake/compilers, a small troubleshooting section, e.g. suggesting ensuring compilers are identifiable and updated, may be useful.

ElliottKasoar:
Resolves #373

### DIFF
--- a/docs/source/user_guide/troubleshooting.rst
+++ b/docs/source/user_guide/troubleshooting.rst
@@ -5,40 +5,9 @@ Troubleshooting
 Installation
 ------------
 
-When installing ``janus-core``, CMake errors can occur when building ``phonopy``, such as:
+When installing ``janus-core``, CMake errors can occur when building ``phonopy``, often related to the CXX compiler.
 
-.. collapse:: Example CMake error
-
-    .. code-block:: bash
-
-        CMake Error at /tmp/pip-build-env-j2jx9pp5/overlay/lib/python3.11/site-packages/nanobind/cmake/nanobind-config.cmake:243 (target_compile_features):
-            target_compile_features The compiler feature "cxx_std_17" is not known to
-            CXX compiler
-
-            "GNU"
-
-            version 4.8.5.
-        Call Stack (most recent call first):
-            /tmp/pip-build-env-j2jx9pp5/overlay/lib/python3.11/site-packages/nanobind/cmake/nanobind-config.cmake:358 (nanobind_build_library)
-            CMakeLists.txt:108 (nanobind_add_module)
-
-
-        -- Configuring incomplete, errors occurred!
-
-        *** CMake configuration failed
-        [end of output]
-
-        note: This error originates from a subprocess, and is likely not a problem with pip.
-        ERROR: Failed building wheel for phonopy
-        Building wheel for python-hostlist (setup.py) ... done
-        Created wheel for python-hostlist: filename=python_hostlist-2.2.1-py3-none-any.whl size=39604 sha256=44f9f27a42895e61a521cf9129a6a3ad03e633b201390da5ef76d5f59db3b94f
-        Stored in directory: ...
-        Successfully built janus-core python-hostlist
-        Failed to build phonopy
-        ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (phonopy)
-
-
-This can typically be resolved by ensuring your C++ compiler is updated, and that the `CXX <https://cmake.org/cmake/help/latest/envvar/CXX.html>`_ environment variable is set.
+Please refer to their `installation documentation <http://phonopy.github.io/phonopy/install.html#missing-or-unknown-cxx-compiler>`_ for guidance.
 
 
 Carbon tracking
--- a/docs/source/user_guide/troubleshooting.rst
+++ b/docs/source/user_guide/troubleshooting.rst
@@ -2,6 +2,45 @@
 Troubleshooting
 ===============
 
+Installation
+------------
+
+When installing ``janus-core``, CMake errors can occur when building ``phonopy``, such as:
+
+.. collapse:: Example CMake error
+
+    .. code-block:: bash
+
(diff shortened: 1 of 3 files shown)


docs/source/user_guide/troubleshooting.rst
```diff
@@ -2,6 +2,45 @@
...
+
+    .. code-block:: bash
+
+        CMake Error at /tmp/pip-build-env-j2jx9pp5/overlay/lib/python3.11/site-packages/nanobind/cmake/nanobind-config.cmake:243 (target_compile_features):
+            target_compile_features The compiler feature "cxx_std_17" is not known to
+            CXX compiler
+
+            "GNU"
+
+            version 4.8.5.
```
oerc0122, line 20: Are we really concerned about trying to compile with a 10 year old release of GCC? This is not our failure and should probably be on phonopy's docs, not ours.

oerc0122 (changes requested):
Not sure this should be here.

docs/source/user_guide/troubleshooting.rst
ElliottKasoar: As I say in the issue, I agree to a large extent it is an issue for phonopy, but it occurs when running `pip install janus-core`, and for the user who encountered this, it wasn't obvious how to resolve it, so I don't think it hurts to have this.

The solution is generally very simple, but if you're not used to cmake/compilation, which I imagine is true for a lot of Python users, it may not be immediately obvious, as it's not something you often need to worry about.

Also just to note, there are two similar issues, which I've tried to cover together, although maybe I shouldn't:

1. Old GCC etc. In this case, it was the default for the user on their HPC system, which I don't think will be uncommon, unfortunately.
2. CXX being unset. This happened to me at one point when I changed environments, and it took me a minute to understand why installing had stopped working, when it had been fine up to then

oerc0122: What's the scope of PR-ing this into phonopy's docs on their "installing phonopy" page and just linking it/directing them there. 

I have no idea how responsive phonopy are to PRs, but this seems a fairly simple thing. 

https://phonopy.github.io/phonopy/install.html#building-using-pip-install

alinelena: in general are responsive

ElliottKasoar: > What's the scope of PR-ing this into phonopy's docs on their "installing phonopy" page and just linking it/directing them there.
> 
> I have no idea how responsive phonopy are to PRs, but this seems a fairly simple thing.
> 
> https://phonopy.github.io/phonopy/install.html#building-using-pip-install

I did wonder that. I think part of my concern is that a user doesn't necessarily immediately assume phonopy is at fault - I don't think the user who encountered the error did.

It's obvious in retrospect, but if you're somewhat used to battling dependency issues etc, your first thought may well be to blame us.

I'd also argue that we want to be relatively clear what requirements there are for installing janus-core, without needing to dive into the details of every sub-dependency. Clearly there are limits, but for a slightly unusual issue with a relatively simple solution, I'm not entirely convinced it's out of scope.

I'll open an issue with phonopy too either way.

ElliottKasoar: It is a known thing: https://github.com/phonopy/phonopy/issues/439

Maybe a different way to look at the problem is that the CMake error does more or less describe what you need to do in either case (update your compiler or make sure it's set). If you're installing phonopy directly, this sort of error is relatively expected, since they describe the need for a compiler, and so in that context, I'm not sure they need to elaborate much on the issue.

When installing janus-core, on the other hand, it's not something a user naturally expects in the same way, and without that context I don't think the solution is as clear.

oerc0122:
```
Failed to build phonopy
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (phonopy)
```
Seems fairly indicative if you're aware of phonopy it's not our problem. Then again, I'm not 100% sure people know what is happening when they pip install. 

I'm fine with it going in in principle, but I do wonder whether we (instead of explaining it) link against that issue? Maybe that's worse. I trust your judgement.

ElliottKasoar:
(Rebased)

@oerc0122's suggestion from recent discussions, which I agree with, is to:

1. Merge this
2. Open a PR with `phonopy` suggesting the issue is added to their troubleshooting, including similar issues e.g. CXX being unset
3. If/when that is merged, replace this text with a link to their troubleshooting

alinelena:
now is mergeed

ElliottKasoar:
> now is mergeed

Yep, only in dev so far, so we still need to wait for it to make it into a release to appear in the docs, but hopefully that'll be pretty soon.

ElliottKasoar:
Rebased, and now links to their updated docs

oerc0122 (approved):

PR #382, opened 2025-01-15
Fix release

ElliottKasoar:
I think by setting the `permissions` explicitly for `id-token`, it removed write permissions, which blocked the release action (previously this was successful, but publishing to PyPI was not).

Adding this back in seemed to work on my fork.

### DIFF
--- a/.github/workflows/publish-on-pypi.yml
+++ b/.github/workflows/publish-on-pypi.yml
@@ -15,6 +15,8 @@ jobs:
     permissions:
       # For PyPI's trusted publishing.
       id-token: write
+      # For release.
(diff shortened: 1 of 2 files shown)


oerc0122 (approved):
Has this deleted the botched release? Or was that a draft?

ElliottKasoar:
> Has this deleted the botched release? Or was that a draft?

There was a botched release yesterday, which I got rid of - it created the release on GitHub, but didn't publish to PyPI. With the change that I hope fixes PyPI, it failed to even generate the release, so there's a bit less mess (just the tag to delete, which I've done).

PR #383, opened 2025-01-16
Test and document external calculators

### LINKED ISSUE #380: Document external calculators?
I'm not proposing we actively work towards more support from other calculators, but it may be useful to note that most janus calculations (should) already work with any ASE-compatible calculator. For example:

```python
from janus_core.calculations.single_point import SinglePoint
from ase.calculators.lj import LennardJones
from ase.io import read

struct = read("tests/data/NaCl.cif")
struct.calc = LennardJones()

single_point = SinglePoint(
    struct=struct,
)

print(single_point.run())
```

returns

```python
{'energy': -0.05900093815771919, 'forces': array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]]), 'stress': array([0.00211953, 0.00211953, 0.00211953, 0.        , 0.        ,
       0.        ])}
```

This might help make a comparison with DFT/classical force fields, or enable the use of a currently unsupported MLIP, so perhaps work documenting?

ElliottKasoar:
Resolves #380

### DIFF
--- a/docs/source/user_guide/python.rst
+++ b/docs/source/user_guide/python.rst
@@ -57,3 +57,30 @@ will return
     If running calculations with multiple MLIPs, ``arch`` and ``mlip_model`` will be overwritten with the most recent MLIP information.
     Results labelled by the architecture (e.g. ``mace_mp_energy``) will be saved between MLIPs,
     unless the same ``arch`` is chosen, in which case these values will also be overwritten.
+
+
+Additional Calculators
+======================
+
+Although ``janus-core`` only directly supports the MLIP calculators listed in :doc:`Getting started </getting_started/getting_started>`,
+any valid `ASE calculator <https://wiki.fysik.dtu.dk/ase/ase/calculators/calculators.html>`_
+can be attached to a structure, including calculators for currently unsupported MLIPs.
+
+This structure can then be passed to ``janus-core`` calculations, which can be run as usual.
+
+For example, performing geometry optimisation using the (`ASE built-in <https://wiki.fysik.dtu.dk/ase/ase/calculators/others.html#lennard-jones>`_) Lennard Jones potential calculator:
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


harveydevereux (commented):
I think this is a good addition for the docs, is there any reason why this would not work?

Obviously if the calc depends on something and you don't have it installed (lammps) it won't, which is fine.

But are there breaking code changes we or ASE might make, justifying testing subbing e.g. ```LennardJones``` into ```struct.calc``` et al.?

ElliottKasoar:
> I think this is a good addition for the docs, is there any reason why this would not work?
> 
> Obviously if the calc depends on something and you don't have it installed (lammps) it won't, which is fine.
> 
> But are there breaking code changes we or ASE might make, justifying testing subbing e.g. `LennardJones` into `struct.calc` et al.?

Yes it inevitably requires the user to set up the calculator correctly, which may well require other installations, and I'm not sure exactly how it would behave with the less native calculators e.g. [CP2K](https://wiki.fysik.dtu.dk/ase/ase/calculators/cp2k.html#cp2k) requires running something like:

```
CP2K.command="env OMP_NUM_THREADS=2 mpiexec -np 4 cp2k_shell.psmp"
```

I knew that every calculation after `SinglePoint` was built in a way that it could (initially it had to) take `SinglePoint.struct` inputs, which meant it wouldn't try to reattach a calculator if it was already there.

The only thing I wasn't sure about was if anywhere we assume things like `struct.calc.parameters` having certain values, but I've always tried to make sure that sort of thing is optional.

I think in general the ASE calculator interface could change, but I don't think anything would break something like `LennardJones` that wouldn't break MACE etc., so perhaps it would be sensible to test for it.

oerc0122 (approved):
Nice and simple, if it does the job, happy with it.

PR #384, opened 2025-01-17
Update MLIPs

ElliottKasoar:
Updates MACE, ALIGNN and SevenNet

Currently, the energy predicted by ALIGNN seems to have changed (see https://github.com/usnistgov/alignn/issues/178). If this is expected, I'll update the value, but otherwise I may revert that bump, and just update MACE and SevenNet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,11 @@ Tools for machine learnt interatomic potentials
 
 - Python >= 3.10
 - ASE >= 3.23
-- mace-torch = 0.3.8
+- mace-torch = 0.3.9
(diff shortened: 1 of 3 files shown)


ElliottKasoar:
(Reverted ALIGNN version bump so we can get MACE in particular updated without waiting for a response)

oerc0122 (approved):
Is it necessary to update both together? 

If I was being particularly finicky, I'd suggest updating them in separate PRs.

ElliottKasoar:
(Rebased)

PR #385, opened 2025-01-17
Tidy PyPI info

ElliottKasoar:
Since the migration to uv, our Python badge doesn't work. I think we need to add these classifiers - see https://github.com/badges/shields/issues/5550.

Also capitalises the links that appear on [PyPI](https://pypi.org/project/janus-core/).


oerc0122 (approved):
Weird it needs title case... 

Looks fine!

PR #388, opened 2025-01-21
Add Correlator CLI and defaults

harveydevereux:
- Adds CLI arguments for requesting correlations defined in ```janus_core.processing.observables```.
- Adds default values for ```blocks=1```, ```points=1```, ```averaging=1``` and ```update_frequency=1``` in```janus_core.processing.correlator.Correlation```.
- Add default for ```Velocity``` (all components, all atoms)
- Fixes docstring in ```md.py```.

Last thing to do is add text in the user_guide 

---

Currently this is example functionality

```shell
janus md \
--ensemble nve --struct tests/data/NaCl.cif \
--steps 10 --correlation-kwargs \
"{'vaf1': {'a': 'velocity', 'points': 10}, 'vaf_x': {'a': 'velocity', 'update_frequency': 2, 'points': 10, 'a_kwargs': {'components': ['x']}, 'b_kwargs': '.'}}"
```
Syntax
- If one of 'a' and 'b' is omitted it is duplicated. 
- The special "ditto" kwarg value "." indicates copying the kwargs.
- 'blocks', 'points' etc have defaults which are not thought out yet.

it results in
```shell
cat NaCl-nve-T300.0-cor.dat
vaf1:
  lags: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0]
  value: [0.0005823628861742493, 0.0005829660336725057, 0.0005832862278350698, 0.0005833234400626741,
    0.0005830778925578795, 0.0005825500577781349, 0.0005817406577431897, 0.0005806506631988626,
    0.0005792812926375899, 0.0005776340111746314]
vaf_x:
  lags: [0.0, 1.0, 2.0, 3.0, 4.0, 5.0]
  value: [0.0005445815119304412, 0.0005449602181734653, 0.0005449448511073212, 0.0005445360209632227,
    0.0005437350402279519, 0.0005425439175493092]

### DIFF
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -335,7 +335,7 @@ The RDF is also computed from the trajectory file and the options ``rdf_start``,
 On-the-fly Processing
 +++++++++++++++++++++
 
-Alongside post-processing correlations may be calculate during MD. This means that, for example, the VAF may be computed without storing trajectory data. To compute the VAF at runtime the following options can be passed:
+Alongside post-processing, correlations may be calculated during MD. This means that, for example, the VAF may be computed without storing trajectory data. To compute the VAF at runtime the following options can be passed:
 
 .. code-block:: bash
 
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -332,6 +332,33 @@ will compute the RDFs for ``Na`` and ``Cl`` atoms. Seperately for each paring, u
 
 The RDF is also computed from the trajectory file and the options ``rdf_start``, ``rdf_stop``, and ``rdf_step`` may be used to control which trajectory frames are utilised.
 
+On-the-fly Processing
++++++++++++++++++++++
+
+Alongside post-processing correlations may be calculate during MD. This means that, for example, the VAF may be computed without storing trajectory data. To compute the VAF at runtime the following options can be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --correlation-kwargs "{'vaf': {'a': 'Velocity', 'points': 100, 'correlation_frequency': 2}}"
+
+This would result in the file ```janus_results/NaCl-nve-T300.0-cor.dat``` containing the combined VAF for Na and Cl atoms correlated every other step, meaning 50 correlation lag times.
+
+The option ``a`` specifies the Observable to be correlated. Possible values are ``Velocity``, ``Stress``, ``StressHydrostatic``, and ``StressShear``. The latter two stresses combine the diagonal and off-diagonal components of the stress tensor. When ``b`` is not specified it is set to a copy of ``a`` to form an auto-correlation.
+
+Correlation observables may also specify their own keyword arguments. For example to specify the components of stress to correlated over (with the correlations averaged) the following options may be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --correlation-kwargs "{'saf': {'a': 'Stress', 'points': 100, 'a_kwargs': {'components': ['xy', 'yz', 'zx']}}}"
+
+Resulting in the stress auto-correlation function :math:`\frac{1}{3}(\langle\sigma_{xy}\sigma_{xy}\rangle+\langle\sigma_{yz}\sigma_{yz}\rangle+\langle\sigma_{zx}\sigma_{zx}\rangle)`, calculated every step for 100 correlation lag times.
+
+The Velocity observable may also be computed over specific components (it defaults to all) and atom slices. To compute over odd indexed atoms (Na here) the following options may be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --correlation-kwargs "{'vaf': {'a': 'Velocity', 'points': 100, 'a_kwargs': {'atoms_slice': (0, None, 2)}}}"
+
 Heating
 -------
 
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -156,7 +156,7 @@ class MolecularDynamics(BaseCalculation):
         Keyword arguments to pass to `output_structs` when saving trajectory and final
         files. Default is {}.
     post_process_kwargs
-        Keyword arguments to control post-processing operations.
+        Keyword arguments to control post-processing operations. Default is None.
     correlation_kwargs
         Keyword arguments to control on-the-fly correlations. Default is None.
     seed
@@ -324,9 +324,9 @@ class MolecularDynamics(BaseCalculation):
             Keyword arguments to pass to `output_structs` when saving trajectory and
             final files. Default is {}.
         post_process_kwargs
-            Keyword arguments to control post-processing operations.
+            Keyword arguments to control post-processing operations. Default is None.
         correlation_kwargs
-            Keyword arguments to control on-the-fly correlations.
+            Keyword arguments to control on-the-fly correlations. Default is None.
         seed
             Random seed used by numpy.random and random functions, such as in Langevin.
             Default is None.
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -157,8 +157,8 @@ class MolecularDynamics(BaseCalculation):
         files. Default is {}.
     post_process_kwargs
         Keyword arguments to control post-processing operations.
-    correlation_kwargs : list[CorrelationKwargs] | None
-        Keyword arguments to control on-the-fly correlations.
+    correlation_kwargs
+        Keyword arguments to control on-the-fly correlations. Default is None.
     seed
         Random seed used by numpy.random and random functions, such as in Langevin.
         Default is None.
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -157,7 +157,7 @@ class MolecularDynamics(BaseCalculation):
         files. Default is {}.
     post_process_kwargs
         Keyword arguments to control post-processing operations.
-    correlation_kwargs
+    correlation_kwargs : list[CorrelationKwargs] | None
         Keyword arguments to control on-the-fly correlations.
     seed
         Random seed used by numpy.random and random functions, such as in Langevin.
--- a/janus_core/cli/md.py
+++ b/janus_core/cli/md.py
@@ -12,6 +12,7 @@ import yaml
 from janus_core.cli.types import (
     Architecture,
     CalcKwargs,
+    CorrelationKwargs,
     Device,
     EnsembleKwargs,
     FilePrefix,
@@ -24,7 +25,7 @@ from janus_core.cli.types import (
     Summary,
     WriteKwargs,
 )
-from janus_core.cli.utils import yaml_converter_callback
+from janus_core.cli.utils import parse_correlation_kwargs, yaml_converter_callback
 
 app = Typer()
 
@@ -221,6 +222,7 @@ def md(
     ] = None,
     write_kwargs: WriteKwargs = None,
     post_process_kwargs: PostProcessKwargs = None,
+    correlation_kwargs: CorrelationKwargs = None,
     seed: Annotated[
         int | None,
         Option(help="Random seed for numpy.random and random functions."),
@@ -350,7 +352,9 @@ def md(
         files. Default is {}.
     post_process_kwargs
         Kwargs to pass to post-processing.
-    seed
+    correlation_kwargs : Optional[CorrelationKwargs]
+        Kwrag to pass for on-the-fly correlations.
+    seed : Optional[int]
         Random seed used by numpy.random and random functions, such as in Langevin.
         Default is None.
     log
@@ -379,7 +383,6 @@ def md(
 
     # Check options from configuration file are all valid
     check_config(ctx)
-
     [
         read_kwargs,
         calc_kwargs,
@@ -397,6 +400,7 @@ def md(
             post_process_kwargs,
         ]
     )
+    correlation_kwargs = parse_correlation_kwargs(correlation_kwargs)
 
     if ensemble not in get_args(Ensembles):
         raise ValueError(f"ensemble must be in {get_args(Ensembles)}")
@@ -482,6 +486,7 @@ def md(
         "temp_time": temp_time,
         "write_kwargs": write_kwargs,
         "post_process_kwargs": post_process_kwargs,
+        "correlation_kwargs": correlation_kwargs,
         "seed": seed,
     }
 
--- a/janus_core/cli/md.py
+++ b/janus_core/cli/md.py
@@ -351,9 +351,9 @@ def md(
         Keyword arguments to pass to `output_structs` when saving trajectory and final
         files. Default is {}.
     post_process_kwargs
-        Kwargs to pass to post-processing.
-    correlation_kwargs : Optional[CorrelationKwargs]
-        Kwrag to pass for on-the-fly correlations.
+        Keyword arguments to pass to post-processing. Default is None.
+    correlation_kwargs
+        Keyword arguments to pass for on-the-fly correlations. Default is None.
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: janus_core/cli/utils.py [hunk 1]
```diff
@@ -309,3 +311,88 @@ def check_config(ctx: Context) -> None:
...
+
+    Returns
+    -------
+    Observable
+        The selected observable.
+    """
+    if name.lower() == "velocity":
+        return Velocity(**kwargs)
+    if name.lower() == "stress":
+        return Stress(**kwargs)
```
ElliottKasoar, line 335: Did we not want `getattr` here to get anything defined in `janus_core.processing.observables`, rather than manually listing them?

harveydevereux: so ```getattr(janus_core.processing.observables, name)```?

ElliottKasoar: Yes, I think so. It means we don't have to revisit that list every time a new observable is added, and in theory a user could add their own to the module as well.

janus_core/cli/md.py
```diff
@@ -289,7 +291,9 @@ def md(
         files. Default is {}.
     post_process_kwargs
         Kwargs to pass to post-processing.
-    seed
+    correlation_kwargs : Optional[CorrelationKwargs]
+        Kwrag to pass for on-the-fly correlations.
```
ElliottKasoar, line 295:
```suggestion
        Keyword arguments to pass for on-the-fly correlations.
```
+ default (also post_process_kwargs)

janus_core/cli/md.py
```diff
@@ -336,6 +339,9 @@ def md(
         ]
     )
 
+    if correlation_kwargs:
+        correlation_kwargs = parse_correlation_kwargs(correlation_kwargs)
```
ElliottKasoar, line 343: Maybe add a comment for why this is parsed differently to other dicts?

janus_core/cli/utils.py
```diff
@@ -309,3 +311,58 @@ def check_config(ctx: Context) -> None:
...
+    Parse CLI CorrelationKwargs to md correlation_kwargs.
+
+    Parameters
+    ----------
+    kwargs : CorrelationKwargs
+        CLI correlation keyword options.
+
+    Returns
+    -------
+    List[dict]
```
ElliottKasoar, line 327:
```suggestion
    list[dict]
```

janus_core/cli/utils.py
```diff
@@ -309,3 +311,58 @@ def check_config(ctx: Context) -> None:
...
             raise ValueError(f"'{option}' in configuration file is not a valid option")
+
+
+def parse_correlation_kwargs(kwargs: CorrelationKwargs) -> list[dict]:
+    """
+    Parse CLI CorrelationKwargs to md correlation_kwargs.
+
+    Parameters
+    ----------
+    kwargs : CorrelationKwargs
```
ElliottKasoar, line 322: I think you can drop the type hints for parameters like this in the doctoring

janus_core/cli/utils.py
```diff
@@ -309,3 +311,58 @@ def check_config(ctx: Context) -> None:
...
+
+    Returns
+    -------
+    List[dict]
+        The parsed correlation_kwargs for md.
+    """
+    parsed_kwargs = []
+    for name, cli_kwargs in kwargs.value.items():
+        if "a" not in cli_kwargs and "b" not in cli_kwargs:
+            raise ValueError("At least on observable must be supplied as 'a' or 'b'")
```
ElliottKasoar, line 333:
```suggestion
            raise ValueError("At least one observable must be supplied as 'a' or 'b'")
```

janus_core/cli/utils.py
```diff
@@ -309,3 +311,58 @@ def check_config(ctx: Context) -> None:
...
+
+        # Accept on Observable to be replicated.
+        if "b" not in cli_kwargs:
+            a = cli_kwargs["a"]
+            b = a
```
ElliottKasoar, line 338: Is it fine not being a copy?

janus_core/cli/utils.py
```diff
@@ -309,3 +311,58 @@ def check_config(ctx: Context) -> None:
...
+        else:
+            a = cli_kwargs["a"]
+            b = cli_kwargs["b"]
+
+        a_kwargs = cli_kwargs["a_kwargs"] if "a_kwargs" in cli_kwargs else {}
+        b_kwargs = cli_kwargs["b_kwargs"] if "b_kwargs" in cli_kwargs else {}
+
+        # Accept "." in place of one kwargs to repeat.
+        if a_kwargs == "." and b_kwargs == ".":
+            raise ValueError("a_kwargs and b_kwargs cannot 'ditto' eachother")
```
ElliottKasoar, line 351:
```suggestion
            raise ValueError("a_kwargs and b_kwargs cannot 'ditto' each other")
```

janus_core/processing/correlator.py
```diff
@@ -253,13 +253,13 @@ class Correlation:
...
+    blocks : int, default 1
         Number of correlation blocks.
-    points
+    points : int, default 1
         Number of points per block.
-    averaging
+    averaging : int, default 1
         Averaging window per block level.
-    update_frequency
+    update_frequency : int, default 1
```
ElliottKasoar, line 262: I think these probably should be

```
blocks
    Number of correlation blocks. Default is  1.
```
etc

janus_core/processing/correlator.py
```diff
@@ -285,13 +285,13 @@ def __init__(
             Observable for b.
         name
             Name of correlation.
-        blocks
+        blocks : int, default 1
```
ElliottKasoar, line 288: Same as above

tests/test_correlator.py
ElliottKasoar, line 1: We should probably test the set up quite carefully e.g. dittoing etc. Assuming it's set up on init, this can probably be done without even running the MD in a lot of cases - just checking the kwargs, observables etc behave as expected

ElliottKasoar (commented):
Mostly minor comments/tidying up, but looks good so far, thanks!

janus_core/calculations/md.py
```diff
@@ -154,7 +154,7 @@ class MolecularDynamics(BaseCalculation):
         files. Default is {}.
     post_process_kwargs
         Keyword arguments to control post-processing operations.
-    correlation_kwargs
+    correlation_kwargs : list[CorrelationKwargs] | None
         Keyword arguments to control on-the-fly correlations.
```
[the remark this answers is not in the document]
harveydevereux: I updated these to just list the default and not the type hint as that is the style of the others (also post-process)

janus_core/cli/utils.py
harveydevereux: maybe there is a case, being safe wont be onerous anyway

ElliottKasoar (commented):
A couple of places where I don't think the type hints are needed anymore, but generally looks good.

Could you add an example to the CLI docs (with a line or two describing roughly the composition, since it's not obvious)?

janus_core/cli/utils.py
```diff
@@ -309,3 +312,58 @@ def check_config(ctx: Context) -> None:
...
+            b = deepcopy(a)
+        elif "a" not in cli_kwargs:
+            a = cli_kwargs["b"]
+            b = deepcopy(a)
+        else:
+            a = cli_kwargs["a"]
+            b = cli_kwargs["b"]
+
+        a_kwargs = cli_kwargs["a_kwargs"] if "a_kwargs" in cli_kwargs else {}
+        b_kwargs = cli_kwargs["b_kwargs"] if "b_kwargs" in cli_kwargs else {}
```
oerc0122, line 348: Does *this* want to use the dict or None function?

janus_core/cli/utils.py
```diff
@@ -309,3 +312,58 @@ def check_config(ctx: Context) -> None:
...
+
+        cor_kwargs = {
+            "name": name,
+            "a": getattr(observables, a)(**a_kwargs),
+            "b": getattr(observables, b)(**b_kwargs),
+        }
+
+        for optional in ["blocks", "points", "averaging", "update_frequency"]:
+            if optional in cli_kwargs:
+                cor_kwargs[optional] = cli_kwargs[optional]
```
oerc0122, line 366: Is `cli_kwargs` a dict? If so:
```python
for optional in cli_kwargs.keys() & {"blocks", "points", "averaging", "update_frequency"}:
    cor_kwargs[optional] = cli_kwargs[optional]
```

janus_core/processing/correlator.py
```diff
@@ -269,10 +269,10 @@ def __init__(
...
         b: Observable,
         name: str,
-        blocks: int,
-        points: int,
-        averaging: int,
-        update_frequency: int,
+        blocks: int = 1,
+        points: int = 1,
+        averaging: int = 1,
+        update_frequency: int = 1,
```
oerc0122, line 275: Is 1 a sensible default for all of these?

tests/test_md_cli.py
```diff
@@ -68,6 +68,7 @@ def test_md(ensemble):
     traj_path = Path(f"{file_prefix[ensemble]}traj.extxyz").absolute()
     rdf_path = Path(f"{file_prefix[ensemble]}rdf.dat").absolute()
     vaf_path = Path(f"{file_prefix[ensemble]}vaf.dat").absolute()
+    cor_path = Path(f"{file_prefix[ensemble]}cor.dat").absolute()
```
oerc0122, line 71: For all of the above:
```python
paths = {key: file_prefix[ensemble] + name
for key, name in [("final", "final.extxyz"),
                  ("restart", "res-2.extxyz"),
                  ("stats", "stats.dat"),
                  ...
}

for file in paths.values():
     assert not file.exists()
...
for file in paths.values():
      assert file.exists()
      file.unlink(missing_ok=True)
```

janus_core/processing/correlator.py
harveydevereux: Points should be not defaulted, its the main lever so should be specified each time. The others are reasonable.

update frequency is arguable but less defaults will lead to much longer cli options. If you want them all specified each time we can do that though.

janus_core/cli/utils.py
harveydevereux: Dict I think, since it is expanded as kwargs later ```getattr(observables, a)(**a_kwargs)```

alinelena:
will be nice to have this in for training

docs/source/user_guide/command_line.rst
```diff
@@ -332,6 +332,33 @@ will compute the RDFs for ``Na`` and ``Cl`` atoms. Seperately for each paring, u
 
 The RDF is also computed from the trajectory file and the options ``rdf_start``, ``rdf_stop``, and ``rdf_step`` may be used to control which trajectory frames are utilised.
 
+On-the-fly Processing
++++++++++++++++++++++
+
+Alongside post-processing correlations may be calculate during MD. This means that, for example, the VAF may be computed without storing trajectory data. To compute the VAF at runtime the following options can be passed:
```
ElliottKasoar, line 338:
```suggestion
Alongside post-processing, correlations may be calculated during MD. This means that, for example, the VAF may be computed without storing trajectory data. To compute the VAF at runtime the following options can be passed:
```

janus_core/cli/utils.py
```diff
@@ -306,3 +309,66 @@ def check_config(ctx: Context) -> None:
...
+    list[dict]
+        The parsed correlation_kwargs for md.
+    """
+    parsed_kwargs = []
+    for name, cli_kwargs in kwargs.value.items():
+        if "a" not in cli_kwargs and "b" not in cli_kwargs:
+            raise ValueError("At least one observable must be supplied as 'a' or 'b'")
+
+        if "points" not in cli_kwargs:
+            raise ValueError("Correlation keyword argument points must be specified")
```
ElliottKasoar, line 334:
```suggestion
            raise ValueError("Correlation keyword argument 'points' must be specified")
```

janus_core/cli/utils.py
```diff
@@ -306,3 +309,66 @@ def check_config(ctx: Context) -> None:
...
+            a_kwargs = b_kwargs
+
+        cor_kwargs = {
+            "name": name,
+            "points": cli_kwargs["points"],
+            "a": getattr(observables, a)(**a_kwargs),
+            "b": getattr(observables, b)(**b_kwargs),
+        }
+
+        for optional in cli_kwargs.keys() & {"blocks", "averaging", "update_frequency"}:
```
ElliottKasoar, line 370: Do we want to raise an error if any keys aren't in this set?

ElliottKasoar (changes requested):
In the summary, a couple of the outputs haven't been properly parsed by the tuple and dict conversion utils. Can you try to have a look? E.g.

```
  correlation_kwargs: !!python/object:janus_core.cli.types.TyperDict
    value:
      vaf:
        a: Velocity
        a_kwargs: &id001
          atoms_slice: !!python/tuple
          - 0
          - null
          - 2
        b_kwargs: *id001
        points: 100
```

(Paths should be converted to strings, and tuples to lists)

Also worth rebasing again, to check the output file changes don't clash.

janus_core/cli/md.py
```diff
@@ -375,6 +378,10 @@ def md(
         ]
     )
 
+    # Handle separately to process short-hands, and Observables.
+    if correlation_kwargs:
+        correlation_kwargs = parse_correlation_kwargs(correlation_kwargs)
```
ElliottKasoar, line 406: For the summary, I think you need to add a copy of `correlation_kwargs to `all_kwargs`, ideally before this parsing if possible. The idea is `all_kwargs` makes a copy of the commands you actually passed, before any processing etc.

janus_core/cli/utils.py
harveydevereux: Depends how you want defaults to be handled. Currently these have defaults in the correlator. They could be added here if not present instead I suppose.

Otherwise all those options will need to be specified on the cli which becomes long, I suppose it already is though. 

If that's the preference then these optional values can be required and an error raised.

ElliottKasoar: I'm happy with defaults to be set wherever. The question was more than this looks like it silently throws away `cli_kwargs` keywords if they're not in `{"blocks", "averaging", "update_frequency"}`, unless I'm misunderstanding, which might be confusing.

janus_core/cli/utils.py
```diff
@@ -17,12 +18,14 @@
...
     from typer import Context
 
-    from janus_core.cli.types import TyperDict
+    from janus_core.cli.types import CorrelationKwargs, TyperDict
     from janus_core.helpers.janus_types import (
         MaybeSequence,
         PathLike,
     )
 
+from janus_core.processing import observables
```
ElliottKasoar, line 27: This should perhaps be moved into `parse_correlation_kwargs`, otherwise I think it and other things it depends on e.g. ase may be imported earlier than we need.

janus_core/cli/md.py
```diff
@@ -409,9 +412,14 @@ def md(
         "ensemble_kwargs": ensemble_kwargs.copy(),
         "write_kwargs": write_kwargs.copy(),
         "post_process_kwargs": post_process_kwargs.copy(),
+        "correlation_kwargs": correlation_kwargs.value.copy(),
```
ElliottKasoar, line 415:
```suggestion
        "correlation_kwargs": correlation_kwargs.copy(),
```

janus_core/cli/utils.py
ElliottKasoar: I'd be inclined to raise as error, unless there's a reason not to? Otherwise you end up running an entire simulation  while not calculating what you actually wanted.

harveydevereux: Yeah its raising an error now e.g. in tests for typo: 

```python
        (
            {"vaf": {"points": 10, "a": "Velocity", "a_kwarg": {}}},
            "correlation_kwargs got unexpected argument",
        ),
```

ElliottKasoar: Ah was looking for it nearer here, nice, thanks!

ElliottKasoar (approved):

PR #389, opened 2025-01-22
Update ASE

ElliottKasoar:
The CI has been running 3.24 since its release in December, but this makes sure it's updated, which means we can drop checks for the Bussi thermostat, as well as include any other [improvements made](https://wiki.fysik.dtu.dk/ase/releasenotes.html#releasenotes).


alinelena (approved):

ElliottKasoar:
(Rebased)

PR #393, opened 2025-01-29
Add NEB

### LINKED ISSUE #375: Add Nudge Elastic Band
See [proof of concept](https://gitlab.com/drFaustroll/lavello/-/blob/main/neb.py) and [tutorial examples](https://github.com/stfc/janus-tutorials/tree/main/neb)

ElliottKasoar:
Resolves #375

Starting point for NEB.

It's a bit fiddly, as most calculations are designed to take in a single input structure, whereas this requires two.

Would it also be useful to allow the full image band as an input, in which case interpolation could be skipped? 

To do:

- [x] Add tests
- [x] Fix images/bands, possibly via pymatgen's `AseAtomsAdaptor`, as used in the [tutorial](https://github.com/stfc/janus-tutorials/blob/main/neb/LiFePO4-NEB-mace_mp-small.ipynb)
- [x] Add CLI
- [x] Update documentation

~Note: needs rebasing after #392~


alinelena:
for the neb, since interpolate is a form of magic, I suggest we implement 3 solutions..

[ ] ase way
[ ] pymatgen way
[ ] user way, user provides the images

janus_core/calculations/neb.py
```diff
@@ -0,0 +1,569 @@
...
+        # Set default neb_kwargs
+        if isinstance(self.neb_class, ASE_NEB):
+            neb_defaults = {"k": 0.1, "climb": True, "method": "string"}
+        elif isinstance(self.neb_class, DyNEB):
+            neb_defaults = {
+                "fmax": 0.1,
+                "dynamic_relaxation": True,
+                "climb": True,
+                "scale_fmax": 1.2,
+            }
```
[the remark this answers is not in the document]
alinelena: yes

janus_core/helpers/janus_types.py
```diff
@@ -121,6 +121,7 @@ class CorrelationKwargs(TypedDict, total=True):
 Ensembles = Literal["nph", "npt", "nve", "nvt", "nvt-nh", "nvt-csvr"]
 Properties = Literal["energy", "stress", "forces", "hessian"]
 PhononCalcs = Literal["bands", "dos", "pdos", "thermal"]
+Interpolators = Literal["ase", "pymatgen"]
```
alinelena, line 124: is not a third category, user provided?

janus_core/calculations/neb.py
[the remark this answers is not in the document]
ElliottKasoar: Just to check, yes we should use class defaults?

janus_core/helpers/janus_types.py
ElliottKasoar: Yes, I did have something like `"custom"` or `None` originally as its own option, but since we can assume user provided/no interpolation if `band_structs` is passed (and it will raise errors if the initial/final structure are also passed), it felt a bit unnecessary to also make a user specify this too?

ElliottKasoar:
> overall looks good, shall we add the examples to this both cli and jupyter?

There are a couple of examples in the user guide CLI docs, which is where all the CLI examples moved a while ago (other than a single `janus singlepoint` example in the README).

I think it might make sense to add the notebook tutorial separately, since it's already a lot of changes

janus_core/calculations/neb.py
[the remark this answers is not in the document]
alinelena: yes

janus_core/helpers/janus_types.py
alinelena: what I am going at is that if we use user provided data, what will be the state of Interpolators? if None I am happy, if one of the two, I am not.

alinelena:
> > overall looks good, shall we add the examples to this both cli and jupyter?
> 
> There are a couple of examples in the user guide CLI docs, which is where all the CLI examples moved a while ago (other than a single `janus singlepoint` example in the README).
> 
> I think it might make sense to add the notebook tutorial separately, since it's already a lot of changes

happy for notebook to come separate, open an issue. also we need full working configs... and probably need to be tested too... but this will be a separate issue too.

janus_core/helpers/janus_types.py
ElliottKasoar: Ah ok it was slightly inconsistent. The CLI set interpolator to be None, but actually the Python still required the user to choose None.

It now should set interpolator to None on the init for the Python or CLI, and docstrings should reflect this.

alinelena (approved):

alinelena:
@oerc0122 can you have a look at this to see if I missed anythign?

PR #400, opened 2025-02-04
Adds post-process cli guide, specifying vafs by element name

harveydevereux:
I found I could not easily request partial VAFs on the CLI by element names (or StartStopStep likes). But instead would have to write all the indices down, which can be automated in shell but is quite unwieldy. So I implemented selecting vaf atoms by element name. Hopefully a useful addition.

Partial VAFs can be obtained in the CLI like so,

```shell
janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10\
 --post-process-kwargs "{'vaf_compute': True, 'vaf_atoms': (('Na',), ('Cl',)), 'vaf_output_files': ('vaf_na.dat', 'vaf_cl.dat')}"
```

where 'Na' and 'Cl' are expanded to indices of those elements in the trajectory.

Also add some documentation for post-processing.

### DIFF
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -283,7 +283,7 @@ To compute the VAF across all atoms from the command line the following options 
 
    janus md --ensemble nve --struct tests/data/NaCl.cif --steps 1000 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_start': 10, 'vaf_stop': None, 'vaf_step': 2}"
 
-In this case a file ``NaCl-nve-T300.0-vaf.dat`` would be created containing correlation values and lag times. The file will contain correlation 45 values with an implied time step of 20 fs. That is starting at time step 100 and correlating every 20 steps.
+In this case a file ``NaCl-nve-T300.0-vaf.dat`` would be created containing correlation values and lag times. The file will contain 45 correlation values with an implied time step of 20 fs. That is starting at time step 100 and correlating every 20 steps.
 
 The VAF is computed from the trajectory file, therefore ``--traj-every`` controls the correlation possible temporal resolutions. The precise steps used in the VAF may be controlled by the ``vaf_start``, ``vaf_stop``, and ``vaf_step`` options.
 
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -295,6 +295,14 @@ To compute partial VAFs of Na and Cl the following kwargs may be passed:
 
 Where ``'vaf_atoms'`` is a ``Sequence`` of ``Sequence`` of element names (or atom indices) included in each VAF. The output files must also be specified in the case of multiple VAFs, by default defined relative to the working directory.
 
+Computing the RDF is similar, for example:
+
+.. code-block:: bash
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10 --post-process-kwargs "{'rdf_compute': True, 'rdf_rmax': 2.0, 'rdf_elements': ('Na', 'Cl'), 'rdf_by_elements': True}"
+
+will compute the RDFs for ``Na`` and ``Cl`` atoms. Seperately for each paring, up to a maximum cutoff distance of ``2.0`` Angstroms. These will be written to three seperate files ``NaCl-nve-T300.0-Na_Na-rdf.dat``, ``NaCl-nve-T300.0-Cl_Cl-rdf.dat``, and ``NaCl-nve-T300.0-Na_Cl-rdf.dat``. Setting ``'rdf_by_elements': False`` will generate one RDF of all atoms saved to ``NaCl-nve-T300.0-rdf.dat``.
+
+The RDF is also computed from the trajectory file and the options ``rdf_start``, ``rdf_stop``, and ``rdf_step`` may be used to control which trajectory frames are utilised.
 
 Heating
 -------
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -275,7 +275,7 @@ For all options, run ``janus md --help``.
 Post-processing
 +++++++++++++++
 
-After MD velocity autocorrelation functions (VAFs) and radial distribution functions (RDFs) may be calculated as a post-processing step. It is possible to build partial VAFs and RDFs from arbitrary sequences of atom indices or species names.
+After MD, velocity autocorrelation functions (VAFs) and radial distribution functions (RDFs) may be calculated as a post-processing step. It is possible to build partial VAFs and RDFs from arbitrary sequences of atom indices or species names.
 
 To compute the VAF across all atoms from the command line the following options could be passed:
 
@@ -291,7 +291,7 @@ To compute partial VAFs of Na and Cl the following kwargs may be passed:
 
    janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_atoms': (('Na',), ('Cl',)), 'vaf_output_files': ('vaf_na.dat', 'vaf_cl.dat')}"
 
-Where ``'vaf_atoms'`` is a ``Sequence`` of ``Sequence`` of element names (or atom indices) included in each VAF. The output files  must also be specified in the case of multiple VAFs, these are by default relative to the working directory.
+Where ``'vaf_atoms'`` is a ``Sequence`` of ``Sequence`` of element names (or atom indices) included in each VAF. The output files must also be specified in the case of multiple VAFs, by default defined relative to the working directory.
 
 
 Heating
--- a/docs/source/user_guide/command_line.rst
+++ b/docs/source/user_guide/command_line.rst
@@ -285,7 +285,7 @@ To compute the VAF across all atoms from the command line the following options 
 
 In this case a file ``NaCl-nve-T300.0-vaf.dat`` would be created containing correlation values and lag times. The file will contain 45 correlation values with an implied time step of 20 fs. That is starting at time step 100 and correlating every 20 steps.
 
-The VAF is computed from the trajectory file, therefore ``--traj-every`` controls the correlation possible temporal resolutions. The precise steps used in the VAF may be controlled by the ``vaf_start``, ``vaf_stop``, and ``vaf_step`` options.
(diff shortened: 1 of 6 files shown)


janus_core/calculations/md.py
```diff
@@ -979,13 +980,18 @@ def _post_process(self) -> None:
...
             use_vel = self.post_process_kwargs.get("vaf_velocities", False)
             fft = self.post_process_kwargs.get("vaf_fft", False)
 
-            out_path = self._build_filename(
-                "vaf.dat", self.param_prefix, filename=file_name
-            )
+            if not isinstance(file_names, Sequence):
+                file_names = (file_names,)
+
+            out_paths = [
```
ElliottKasoar, line 990: Pretty minor but any reason to make this a list rather than a tuple?

ElliottKasoar (commented):
Looks helpful, thanks!

Added a few minor comments from a quick first look.

Quick question that's related, but not new to this: do you think we may want to rename `filter_atoms` to `atoms_filter`? To me, `filter_atoms` sounds more like a boolean.

janus_core/processing/post_process.py
```diff
@@ -165,7 +165,7 @@ def compute_vaf(
     use_velocities: bool = False,
     fft: bool = False,
     index: SliceLike = (0, None, 1),
-    filter_atoms: MaybeSequence[MaybeSequence[int | None]] = ((None),),
+    filter_atoms: MaybeSequence[MaybeSequence[int | str | None]] = ((None),),
```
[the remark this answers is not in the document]
harveydevereux: yes those commas are a pain

harveydevereux:
> Looks helpful, thanks! The documentation is great!
> 
> Added a few minor comments from a quick first look.
> 
> Quick question that's related, but not new to this: do you think we may want to rename `filter_atoms` to `atoms_filter`? To me, `filter_atoms` sounds more like a boolean.



> Looks helpful, thanks! The documentation is great!
> 
> Added a few minor comments from a quick first look.
> 
> Quick question that's related, but not new to this: do you think we may want to rename `filter_atoms` to `atoms_filter`? To me, `filter_atoms` sounds more like a boolean.

I can see that yes so might as well unless anyone can think of a better name

docs/source/user_guide/command_line.rst
```diff
@@ -272,6 +272,27 @@ running an additional 1000 steps, and appending the statistics and trajectory fi
 
 For all options, run ``janus md --help``.
 
+Post-processing
++++++++++++++++
+
+After MD velocity autocorrelation functions (VAFs) and radial distribution functions (RDFs) may be calculated as a post-processing step. It is possible to build partial VAFs and RDFs from arbitrary sequences of atom indices or species names.
```
ElliottKasoar, line 278:
```suggestion
After MD, velocity autocorrelation functions (VAFs) and radial distribution functions (RDFs) may be calculated as a post-processing step. It is possible to build partial VAFs and RDFs from arbitrary sequences of atom indices or species names.
```

docs/source/user_guide/command_line.rst
```diff
@@ -272,6 +272,27 @@ running an additional 1000 steps, and appending the statistics and trajectory fi
...
+
+To compute the VAF across all atoms from the command line the following options could be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 1000 --traj-every 10 --post-process-kwargs "{'vaf_compute': True}"
+
+In this case a file ``NaCl-nve-T300.0-vaf.dat`` would be create containing correlation values and lag times. In this case there will be 100 values with an implied time step of 10 fs.
```
ElliottKasoar, line 286: Worth saying here or above explicitly that the post processing uses the trajectory file (assuming I'm remembering correctly that is does)?

docs/source/user_guide/command_line.rst
```diff
@@ -272,6 +272,27 @@ running an additional 1000 steps, and appending the statistics and trajectory fi
...
+
+To compute partial VAFs of Na and Cl the following kwargs may be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_atoms': (('Na',), ('Cl',)), 'vaf_output_files': ('vaf_na.dat', 'vaf_cl.dat')}"
+
+Where ``'vaf_atoms'`` is a ``Sequence`` of ``Sequence`` of element names (or atom indices) included in each VAF. The output files  must also be specified in the case of multiple VAFs, these are by default relative to the working directory.
```
ElliottKasoar, line 294:
```suggestion
Where ``'vaf_atoms'`` is a ``Sequence`` of ``Sequence`` of element names (or atom indices) included in each VAF. The output files must also be specified in the case of multiple VAFs, by default defined relative to the working directory.
```

ElliottKasoar (changes requested):
I think we might want to handle invalid symbols more nicely. If I ask for "abc" instead of "Cl":


```python
janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_atoms': (('abc',), ('Cl',)), 'vaf_output_files': ('vaf_na.dat', 'vaf_cl.dat')}"
```

I get `TypeError: 'numpy.float64' object is not iterable`, which is quite strange.

docs/source/user_guide/command_line.rst
harveydevereux: yes and it does look to be the case as in md.py

```python
data = read(self.traj_file, index=":")
# ...
compute_vaf(
                data,
                out_paths,
                use_velocities=use_vel,
                fft=fft,
                index=slice_,
                atoms_filter=self.post_process_kwargs.get("vaf_atoms", None),
            )
```

harveydevereux (commented):
fix user guide typos

harveydevereux:
> I think we might want to handle invalid symbols more nicely. If I ask for "abc" instead of "Cl":
> 
> ```python
> janus md --ensemble nve --struct tests/data/NaCl.cif --steps 100 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_atoms': (('abc',), ('Cl',)), 'vaf_output_files': ('vaf_na.dat', 'vaf_cl.dat')}"
> ```
> 
> I get `TypeError: 'numpy.float64' object is not iterable`, which is quite strange.

now checked, and also in tests should recieve

```bash
ValueError: {'abc'} not allowed in VAF, allowed symbols are {'Cl', 'Na'}
```

tests/test_post_process.py
```diff
@@ -212,3 +216,21 @@ def test_vaf(tmp_path):
...
+    data = read(DATA_PATH / "NaCl-traj.xyz", index=":")
+    lags, vaf = post_process.compute_vaf(
+        data, atoms_filter=vaf_filter, filenames=[tmp_path / name for name in vaf_names]
+    )
+    expected = np.loadtxt(tmp_path / "vaf-Na-by-indices.dat")
+    actual = np.loadtxt(tmp_path / "vaf-Na-by-element.dat")
+
+    assert expected == approx(actual, rel=1e-9)
+
+    with pytest.raises(ValueError):
```
ElliottKasoar, line 235:
```suggestion
    # Check invalid symbol
    with pytest.raises(ValueError):
```

This maybe ought to be its own test, but probably a comment is enough to give it some context

janus_core/processing/post_process.py
```diff
@@ -247,12 +248,32 @@ def compute_vaf(
...
+            )
+        elif all(isinstance(a, int) for a in atoms):
+            filtered_atoms.append(atoms)
+        else:
+            raise ValueError(
+                "Cannot mix element symbols and indices in vaf atoms_filter"
+            )
 
-    used_atoms = {atom for atoms in filter_atoms for atom in atoms}
+    used_atoms = {atom for atoms in filtered_atoms for atom in atoms}
```
ElliottKasoar, line 276: Hasn't `filter_atoms` been replaced by `atoms_filter` generally? Is this still just renaming, or actually changing something? (also below)

docs/source/user_guide/command_line.rst
```diff
@@ -272,6 +272,29 @@ running an additional 1000 steps, and appending the statistics and trajectory fi
...
+
+To compute the VAF across all atoms from the command line the following options could be passed:
+
+.. code-block:: bash
+
+   janus md --ensemble nve --struct tests/data/NaCl.cif --steps 1000 --traj-every 10 --post-process-kwargs "{'vaf_compute': True, 'vaf_start': 10, 'vaf_stop': None, 'vaf_step': 2}"
+
+In this case a file ``NaCl-nve-T300.0-vaf.dat`` would be created containing correlation values and lag times. The file will contain 45 correlation values with an implied time step of 20 fs. That is starting at time step 100 and correlating every 20 steps.
+
+The VAF is computed from the trajectory file, therefore ``--traj-every`` controls the correlation possible temporal resolutions. The precise steps used in the VAF may be controlled by the ``vaf_start``, ``vaf_stop``, and ``vaf_step`` options.
```
ElliottKasoar, line 288:
```suggestion
The VAF is computed from the trajectory file, so ``--traj-every`` controls the possible temporal resolution of the correlations. The precise steps used in the VAF may be controlled by the ``vaf_start``, ``vaf_stop``, and ``vaf_step`` options.
```

ElliottKasoar, line 288: I'm not sure if there's a clearer way of saying this, but it could be read as a contradictory to say that the resolution depends on the trajectory, but also that the "precise steps" can be controlled. Maybe I'd just drop "precise"?

A second question: how does the `vaf_step` interact with the `traj_every`? Is it the step, within the steps saved in the trajectory, or MD steps? If the latter, what happens if `vaf_step` is not a multiple of `traj_every`?

harveydevereux: Yes it is clunky. The trajectory saving being $> 1$ implies a sub-sampling of the timestep, by saving a subset of simulation steps as trajectory frames. 

If you then have a $> 1$ ``vaf_step`` then that in turn sub-samples the sub-sampled timestep, by taking a sub-set of trajectory frames as input. 

I don't think it matters if it is not a multiple. 

Perhaps we just state that ``vaf_step`` is the rate of stepping through the saved trajectory and leave it at that.

janus_core/processing/post_process.py
ElliottKasoar: I don't really mind in terms of implementation (this seems reasonable), just checking it's intentional

docs/source/user_guide/command_line.rst
ElliottKasoar: > Perhaps we just state that `vaf_step` is the rate of stepping through the saved trajectory and leave it at that.

If that's what it's doing, then yeah I think that (still mentioning the start and stop) is enough, and then yeah being a multiple shouldn't matter either.

PR #401, opened 2025-02-04
Specify Python version in dev docs

ElliottKasoar:
By default, uv will try to install Python 3.13 on a system with no existing Python 3 (e.g. scarf, before modules have been loaded).

Specifying different Python versions with `-p` is discussed in the section below, but since people may copy and paste these to get started, it's probably safest to specify this here explicitly.

### DIFF
--- a/docs/source/developer_guide/get_started.rst
+++ b/docs/source/developer_guide/get_started.rst
@@ -21,17 +21,17 @@ This provides a number of useful features, including:
 
 Dependencies useful for development can then be installed by running::
 
-    uv sync
+    uv sync -p 3.12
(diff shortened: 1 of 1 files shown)


alinelena:
i suspect best in this case is to advise users to have a python already installed via conda/mamba or micros in this way they avoid the issue.

ElliottKasoar:
> i suspect best in this case is to advise users to have a python already installed via conda/mamba or micros in this way they avoid the issue.

Why not let uv install it for you? One less thing to worry about (this is only for developers, for now, so uv is expected either way)

alinelena (approved):

PR #402, opened 2025-02-04
Add progress bars to single point calculations

### LINKED ISSUE #324: Use progress bar for additional calculations
> (One other program that could use a progress bar is `singlepoint`, when applied to a file containing many structures.)

_Originally posted by @ajjackson in https://github.com/stfc/janus-core/pull/319#discussion_r1789751435_

(Dependent on merge of #319)

### LINKED ISSUE #403: Use progress bar for single point calculations

k-harris27:
Closes #403

Adds progress bars identical to #319 to single point calculations when applied to multiple frames:

Loading bars

Because each property is calculated in a separate loop, there is one progress bar per property for now. I'm happy to try to make it one overall progress bar if that's preferred.

### DIFF
--- a/janus_core/calculations/single_point.py
+++ b/janus_core/calculations/single_point.py
@@ -22,7 +22,7 @@ from janus_core.helpers.janus_types import (
 )
 from janus_core.helpers.mlip_calculators import check_calculator
 from janus_core.helpers.struct_io import output_structs
-from janus_core.helpers.utils import none_to_dict
+from janus_core.helpers.utils import none_to_dict, track_progress
 
 
 class SinglePoint(BaseCalculation):
@@ -69,6 +69,9 @@ class SinglePoint(BaseCalculation):
     write_kwargs
         Keyword arguments to pass to ase.io.write if saving structure with results of
         calculations. Default is {}.
+    enable_progress_bar
+        Whether to show a progress bar when applied to a file containing many
+        structures. Default is False.
 
     Attributes
     ----------
@@ -94,6 +97,7 @@ class SinglePoint(BaseCalculation):
         properties: MaybeSequence[Properties] = (),
         write_results: bool = False,
         write_kwargs: OutputKwargs | None = None,
+        enable_progress_bar: bool = False,
     ) -> None:
         """
         Read the structure being simulated and attach an MLIP calculator.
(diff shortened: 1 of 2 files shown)


ElliottKasoar (approved):
This looks great, thanks @k-harris27!

If there is a trivial way to have a single bar, that might be preferable, but I suspect that would require restructuring the way properties are calculated, so I think what you've done is probably the best solution.

If no one has any objections soon, I think this is ready to merge. I'll split the issue into sub issues for single point, descriptors and MD, so we can properly resolve the first of these with this,

janus_core/calculations/single_point.py
```diff
@@ -247,7 +260,10 @@ def _get_forces(self) -> MaybeList[ndarray]:
             Forces of structure(s).
         """
         if isinstance(self.struct, Sequence):
-            return [struct.get_forces() for struct in self.struct]
+            struct_sequence = self.struct
+            if self.enable_progress_bar:
+                struct_sequence = track_progress(struct_sequence, "Computing forces...")
```
oerc0122, line 265: Is this actually computing forces at this point? If they're already there I think it's just reading them so this may be misleading.

alinelena:
if my understanding is correct, energy is computed,  forces and stresses are just returned(inherited from energy calculation), and hessian is calculated. this shall be simple to test by running on a big trajectory.

alinelena:
@k-harris27 what I will add to this will be a screenshot, edit your top comment and add it...

janus_core/calculations/single_point.py
ElliottKasoar: I did wonder this too, but I think there are probably too many variables to handle this much more nicely:

- A user may request only one of these properties explicitly, and we don't know which
- Some properties (energy, forces, and stress) are normally calculated at the same time, but others e.g. hessian are not
- This will also work with non-MLIP calculators (or MLIPs that somehow separated things out)

All that happens is that the first progress bar is the most meaningful, and relatively slow, while the following bars (if they are already calculated) are very quick, just reading the calculated results as you say.

This happens in a fairly intuitive order, rather than, for example, the slow progress bar somewhere in the middle, so I'm not sure it hurts the experience too much.

Do you think there's an obvious better alternative?

oerc0122: If it's `tqdm` under the hood, they provide a `delay` kw, which means it'll only show up if it's taking longer than e.g. 1s. Also, changing from "Computing" to "Retrieving" just makes it unambiguous

k-harris27: I've had a look at the `rich` documentation and haven't found a `delay` option. I can change potential energies, forces and stresses to "Retrieving" if that's clearer.

oerc0122: Ah, `rich` doesn't have it from what I can see. I think "Retrieving" makes more sense to me, I defer to Elliott though.

ElliottKasoar: The issue is that the first time one of them is called (it could be any of them), it is calculating, not retrieving, so I don’t think we want it to be changed in all cases.

I think the only way to do this would be some sort of check for whether the results are already stored, but I’m not sure the added complexity is necessarily worth it?

oerc0122: The reason I chose "retrieving" was that could imply calculation. I'm not sure it's all that important in the end.

ElliottKasoar: To me, "calculating" makes more sense when the model is actually being run, so yeah, I think I'd leave it as is, unless we did want to separate out the different processes.

oerc0122 (approved):

PR #406, opened 2025-02-04
Update codecarbon

ElliottKasoar:
We require at least 2.7, as of #316, to use the `allow_multiple_runs` option that was added, but our requirements were not updated to match this at the time.

I've bumped it slightly higher, to include other new features, unless this causes any problems.


alinelena (dismissed):

alinelena (approved):

harveydevereux (approved):
:rocket:

PR #410, opened 2025-02-07
Add isotropic NPT MD

### LINKED ISSUE #398: Add isotropic NPT MD
Now available from ASE: https://gitlab.com/ase/ase/-/merge_requests/3550

Would require test guards to skip if not on ASE main/import fails.

k-harris27:
Resolves #398

Will write tests if this looks OK so far.


janus_core/calculations/md.py
```diff
@@ -1206,6 +1206,7 @@ def __init__(
 
         (ensemble_kwargs,) = none_to_dict(ensemble_kwargs)
         self.ttime = thermostat_time * units.fs
+        self.ptime = barostat_time * units.fs if barostat_time is not None else None
```
ElliottKasoar, line 1209: Is this supposed to be here?

ElliottKasoar (commented):
Looks great so far!

janus_core/calculations/md.py
k-harris27: Ah, in hindsight it's probably better to just have `thermostat_time` and `barostat_time` as kwargs for the `NPT_MTK.__init__` itself. I was trying to use the ones from the inherited `NPT.__init__`.

ElliottKasoar: Ah that makes a lot of sense.

The Bussi thermostat was probably an unhelpful example in this sense, since in the other cases that are subclasses of another ensemble, rather than `MolecularDynamics`, they still share the same `self.dyn`, and just use it with particular parameters to change the behaviour.

In these two cases, `self.dyn` is actually different, so while they do share a lot of parameters, it's a bit confusing to define, then overwrite this.

In retrospect, I'd probably do it differently, so yeah, while I think you could do it via NPT, it probably is better not to.

janus_core/cli/md.py
```diff
@@ -40,12 +40,25 @@ def md(
     thermostat_time: Annotated[
         float,
         Option(
-            help="Thermostat time for NPT, NVT Nosé-Hoover, or NPH simulation, in fs."
+            help=(
+                """
+            Thermostat time for NPT, NPT-MTK, NVT Nosé-Hoover, or NPH simulation, in fs.
+            Default is 50 fs for NPT and NVT Nosé-Hoover, or 100 fs for NPT-MTK.
+            """
```
ElliottKasoar, line 47:
```suggestion
            """
            Thermostat time for NPT, NPT-MTK, NVT Nosé-Hoover, or NPH simulation, in fs.
            Default is 50 fs for NPT, NVT Nosé-Hoover, and NPH, or 100 fs for NPT-MTK.
            """
```

janus_core/cli/md.py
```diff
@@ -40,12 +40,25 @@ def md(
...
     barostat_time: Annotated[
-        float, Option(help="Barostat time for NPT simulation, in fs.")
-    ] = 75.0,
+        float,
+        Option(
+            help=(
+                """
+            Barostat time for NPT, NPT-MTK or NPH simulation, in fs.
+            Default is 75 fs for NPT and NPH, or 1000 fs for NPT-MTK.
+            """
```
ElliottKasoar, line 58:
```suggestion
            """
            Barostat time for NPT, NPT-MTK or NPH simulation, in fs.
            Default is 75 fs for NPT and NPH, or 1000 fs for NPT-MTK.
            """
```

tests/test_md.py
```diff
@@ -25,6 +33,8 @@
     (NVT_NH, "nvt-nh"),
     (NPH, "nph"),
 ]
+if not MTK_IMPORT_FAILED:
+    test_data.append((NPT_MTK, "npt-mtk"))
```
ElliottKasoar, line 37: Probably better to use `pytest.skip` if possible

ElliottKasoar (commented):
A few minor things, but generally looking great!

I can't comment on the lines, but the doctoring describing the defaults for `thermostat_time` and `barostat_time` also need updating (similarly to the help) in janus_core/cli/md.py

tests/test_md.py
```diff
@@ -15,6 +16,13 @@
 from janus_core.helpers.stats import Stats
 from tests.utils import assert_log_contains
 
+if hasattr(ase.md.nose_hoover_chain, "IsotropicMTKNPT"):
+    from janus_core.calculations.md import NPT_MTK
+
+    MTK_IMPORT_FAILED = False
+else:
+    MTK_IMPORT_FAILED = True
+
```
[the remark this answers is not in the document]
k-harris27: Originally I tried that, but ruff wouldn't allow it because `IsotropicMTKNPT` never gets used, even if imported `as _`. I figured `hasattr` makes it a bit more obvious what we're actually checking for too.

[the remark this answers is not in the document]
ElliottKasoar: That would be true for `IsotropicMTKNPT`, although not `NPT_MTK` I imagine? I suppose you would have the problem for both in the CLI tests though, so maybe best to be consistent.

Seems sensible, let's leave it as is.

ElliottKasoar (approved):
Thanks again, @k-harris27, this is really useful!

Issue #397, opened 2025-01-31
Add user installation/upgrade instructions


ElliottKasoar:
As requested by @alinelena and others, we may want some example scripts/tutorials describing how to upgrade MACE, PyTorch, ASE (e.g. https://gitlab.com/drFaustroll/ase.git@npt_triangular), etc, as sometimes users require the latest features that are not yet available in published releases.

We may also want to consider user `uv` instructions, similar to #396, e.g. `uv pip install git+https://github.com/stfc/janus-core.git`. While most users don't need all of the details added in that PR, it is very fast, and can be useful for installing Python, particularly on HPC systems,  similar to micromamba, conda etc.

ElliottKasoar:
This could probably include a few things for GPU acceleration e.g. links to https://mace-docs.readthedocs.io/en/latest/guide/cuda_acceleration.html, particularly as MACE no longer installs cuequivariance by default.

ElliottKasoar:
Note: particularly if we do suggest `uv`, it would be useful to have some comments as part of this with interactions with conda/conda-forge installed things (e.g. PLUMED, in #526).

Issue #328, opened 2024-10-09
Update torch/dgl


ElliottKasoar:
Currently we pin torch to 2.2.0 to resolve issues with `dgl`, which is required by `matgl` (M3GNet), but increasingly becoming a problem.

If we only update torch to 2.4.0, poetry installs everything happily enough, but I get errors such as 

```
FileNotFoundError: Cannot find DGL C++ graphbolt library at /home/ek/PSDI/venv_janus_11_new/lib/python3.11/site-packages/dgl/graphbolt/libgraphbolt_pytorch_2.4.0.so
``` 

This is probably because `dgl` needs to be updated too, but they no longer publish to PyPI, and adding their custom link doesn't seem to work nicely with poetry (possibly related to https://github.com/python-poetry/poetry/issues/9385, which is resolved but not in a release yet).

One solution may be custom `dgl` URLs for each Python version, which I think would work

Alternatively, we may just want to remove automatic testing of M3GNet, and add instructions for installation separately. This will be necessary for Windows/MacOS anyway, as they now require building from source.

ElliottKasoar:
chgnet 0.4.0, which I think is required to resolve the MacOS testing issue (#214) also depends on torch 2.4.1. Unfortunately it also drops support for Python 3.9, which complicates things for us...

alinelena:
i will stay away from 2.4.1 stick to 2.4.0

ElliottKasoar:
What's the problem with 2.4.1? We won't be able to update chgnet for a while then

alinelena:
there is a mace issue and probably nequip

PR #415, opened 2025-02-11
Update torch

### LINKED ISSUE #312: Update chgnet to 0.4.0
See [changelog](https://github.com/CederGroupHub/chgnet/releases/tag/v0.4.0), but nothing obvious should break.

ElliottKasoar:
Resolves #312 and #328

Following https://github.com/astral-sh/uv/pull/11323, we are able to properly set the MLIPs that require `dgl` as conflicting with `chgnet`.

By default, this currently means that torch will be updated to 2.5.1, (I added `torch<2.6` for now, as a few things seemed to break), and we only revert to the pinned `torch==2.2` when either `matgl` or `alignn` are requested.

I also changed `all` to exclude `matgl` or `alignn`, as there is no way to include the conflicting MLIPs, and I hope others we add soon will be compatible with newer torch versions, but the name could potentially be clearer, and we may want an equivalent for `dgl` MLIPs.

This also adds a bit more installation guidance, most of which is relevant to describing how you would go about upgrading `torch`/`dgl`, if you wanted to use all of the MLIPs still, but in theory some of that section could be split out into a separate PR, addressing #397.

To do:

- [x] Fix tests
  - We'll need to install and test the dgl dependencies separately
- [x] Add more developer guidance e.g. for testing?


docs/source/user_guide/installation.rst
```diff
@@ -0,0 +1,77 @@
...
+.. code-block:: bash
+
+    python3 -m pip install -U torch
+
+
+or a specific version:
+
+.. code-block:: bash
+
+    python3 -m pip install torch==2.5.1
```
alinelena, line 26:
```suggestion
    python3 -m pip install -U torch==2.5.1
```

ElliottKasoar: Does this actually change the behaviour? I know it'll change the `torch` version either way, but I suppose sub-dependencies may not update to their highest compatible version otherwise?

ElliottKasoar:
Note: will require resolution of https://github.com/astral-sh/uv/issues/11479

docs/source/user_guide/installation.rst
alinelena: good point... in general I do it without the version... 
some deps may be updated depends of the package constraints, if you want to update deps too --upgrade-strategy eager

.github/workflows/ci.yml
```diff
@@ -17,17 +17,28 @@ jobs:
...
         uses: astral-sh/setup-uv@v4
         with:
-          version: "0.5.7"
+          version: "0.6.2"
           python-version: ${{ matrix.python-version }}
 
-      - name: Install dependencies
-        run: uv sync --all-extras
+      - name: Install "all" dependencies
+        run: uv sync --extra all
```
alinelena, line 24: since nothing is extra anymore... shall we change this to the mlip or something like that.

ElliottKasoar:
I rewrote the developer tutorial to focus on orb, since it's a bit more standard than alignn now, and also did a quick pass to update places in the docs which assumed MACE was installed (it's still required for testing, though).

alinelena (approved):

PR #421, opened 2025-02-13
Add fairchem architectures

### LINKED ISSUE #420: add support for fairchem
meta has their own model https://github.com/FAIR-Chem/fairchem

alinelena:


ElliottKasoar:
I think this is the one that'll conflict with mace due to e3nn

pyproject.toml
```diff
@@ -55,11 +57,19 @@ m3gnet = [
 sevennet = [
     "sevenn == 0.10.3",
 ]
+fairchem = [ 
+    "fairchem-core == 1.4.0",
```
ElliottKasoar, line 61: If we want the version with stresses, run `uv add git+https://github.com/FAIR-Chem/fairchem#subdirectory=packages/fairchem-core --rev omat24` (this will need to be resolved by uv, so we need to deal with torch/e3nn first, but I can also figure out what it would look like it it helps)

ElliottKasoar:
(Quoting #420 for visibility)

> (For context, fairchem "requires" torch==2.4, which will conflict with chgnet==0.4.0 but it seems to work with 2.5.1)
> 
> Ah, we can also manually specify the metadata: https://docs.astral.sh/uv/concepts/resolution/#dependency-metadata
> 
> See also: [astral-sh/uv#2686](https://github.com/astral-sh/uv/issues/2686)

ElliottKasoar:
Also to do:

ElliottKasoar:
Installation problems are related to https://github.com/astral-sh/uv/issues/2252 and comments in https://github.com/astral-sh/uv/issues/7722.

Essentially, `torch-scatter` and `torch-sparse` don't declare `torch` as a build-time dependency (see https://github.com/rusty1s/pytorch_scatter/issues/455).

[Disabling build isolation](https://docs.astral.sh/uv/concepts/projects/config/#build-isolation) is the solution to this, which is what I added.

However, it doesn't seem possible to do this as a single-step process yet, so the solution is something along the lines of running `uv pip install setuptools torch`, followed by `uv sync`.

This is particularly annoying, since these two steps are necessary even when not installing `fairchem` (as everything must be resolved), although the required dependencies are at least always present, so switching between extras won't be a problem.

Since `torch-scatter` and `torch-sparse` aren't actually declared as requirements by `fairchem` yet, the other alternative is raise an error when trying to use it if they're not installed, so removing issues with their dependency resolution.

ElliottKasoar:
eSEN will be [added soon](https://github.com/FAIR-Chem/fairchem/pull/1090), so we may want to hold off and incorporate it at the same time.

pyproject.toml
ElliottKasoar: Now that the conflicts have been resolved, I think we'll keep the standard PyPI installation, but add a note in the docs for installing this branch, along with notes about the torch extras needed.

ElliottKasoar:
> eSEN will be [added soon](https://github.com/FAIR-Chem/fairchem/pull/1090), so we may want to hold off and incorporate it at the same time.

This has been released now, but there remains a slight barrier since the pre-trained eSEN models all require manual approval (https://huggingface.co/facebook/OMAT24), and then setting a Hugging Face token, assuming you are approved.

When I last discussed this with @alinelena, I think his preference was to have a separate `arch` for each model. This makes sense in that they are different architectures, but they do also share the same `OCPCalculator`.

I'm wondering if, from that perspective, it's better just to keep a general `arch==fairchem`, and then perhaps have our own custom labels (`model=esen`/`model=equiformer`?). Otherwise, we end up duplicating the entire section to set the same calculator, and the individual names are perhaps less obvious.

I think the main downside is that arguably neither model is obvious as a default, but we can document it, or not have a default.

It's also different to `mace_off`/`mace_mp` in that, as well as having their own calculators/kwargs, they also conflict in terms of model names (e.g. `small`).

Any thoughts?

alinelena:
the approval itself is quite automatic. let us speak about it when you come from holiday.

janus_core/helpers/mlip_calculators.py
```diff
@@ -305,6 +305,27 @@ def choose_calculator(
...
+
+        model_name = None
+        checkpoint_path = None
+
+        if isinstance(model_path, Path) and model_path.exists():
+            checkpoint_path = str(model_path)
+        else:
+            model_name = str(model_path)
+
+        kwargs.setdefault("local_cache", "pretrained_models")
```
ElliottKasoar, line 322: Shall we change this to be somewhere a bit more hidden? Maybe `~/.cache/fairchem`?

alinelena: you mean the default? yes but be sure it works in the sense creates the folder hierarchy

ElliottKasoar:
(Note: fairchem 2.0.0 was released today, but depends on numpy>2, so will require further changes.)

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+                model = "loaded_Potential"
+            elif isinstance(model, Path):
+                if model.is_file():
+                    model = model.parent
+                potential = load_model(model)
+            elif isinstance(model, str):
+                potential = load_model(model)
+            else:
+                model = "M3GNet-MP-2021.2.8-DIRECT-PES"
+                potential = load_model(model)
```
oerc0122, line 184: I think you could also rewrite this as a `match`:
```suggestion
            match model:
                case Potential():
                    potential = model
                    model = "loaded_Potential"
                case Path():
                    if model.is_file():
                        model = model.parent
                    potential = load_model(model)
                case str():
                    potential = load_model(model)
                case _:
                    model = "M3GNet-MP-2021.2.8-DIRECT-PES"
                    potential = load_model(model)
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+            if isinstance(model, CHGNet):
+                loaded_model = model
+                model = "loaded_CHGNet"
+            elif isinstance(model, Path):
+                loaded_model = CHGNet.from_file(model)
+            elif isinstance(model, str):
+                loaded_model = CHGNet.load(model_name=model, use_device=device)
+            else:
+                model = __version__
+                loaded_model = None
```
oerc0122, line 208:
```suggestion
            match model:
                case CHGNet():
                    loaded_model = model
                    model = "loaded_CHGNet"
                case Path():
                    loaded_model = CHGNet.from_file(model)
                case str():
                    loaded_model = CHGNet.load(model_name=model, use_device=device)
                case _:
                    model = __version__
                    loaded_model = None
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
-        torch.set_default_dtype(torch.float32)
+            # Set default path to directory containing config and model location
+            if isinstance(model, Path):
+                if model.is_file():
+                    model = model.parent
+            # If a string, assume referring to model_name e.g. "v5.27.2024"
+            elif isinstance(model, str):
+                model = get_figshare_model_ff(model_name=model)
+            else:
+                model = default_path()
```
oerc0122, line 230:
```suggestion
            match model:
                case Path():
                    if model.is_file():
                        model = model.parent
                # If a string, assume referring to model_name e.g. "v5.27.2024"
                case str():
                    model = get_figshare_model_ff(model_name=model)
                case _:
                    model = default_path()
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+            from sevenn.sevennet_calculator import SevenNetCalculator
+            import torch
+
+            # Set before loading model to avoid type mismatches
+            torch.set_default_dtype(torch.float32)
+
+            if isinstance(model, Path):
+                model = str(model)
+            elif not isinstance(model, str):
+                model = "SevenNet-0_11July2024"
```
oerc0122, line 245:
```suggestion
            match model:
                case Path() | str():
                    model = str(model)
                case _:
                    model = "SevenNet-0_11July2024"
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+                loaded_model = model
+                model = "loaded_DirectForcefieldRegressor"
+            else:
+                try:
+                    loaded_model = getattr(orb_ff, model.replace("-", "_"))()
+                except AttributeError as e:
+                    raise ValueError(
+                        "`model` must be a `DirectForcefieldRegressor`, pre-trained "
+                        "model label (e.g. 'orb-v2'), or `None` (uses default, orb-v2)"
+                    ) from e
```
oerc0122, line 303:
```suggestion
            match model:
                case DirectForcefieldRegressor():
                    loaded_model = model
                    model = "loaded_DirectForcefieldRegressor"
                case str() if hasattr(orb_ff, model.replace("-", "_")):
                    loaded_model = getattr(orb_ff, model.replace("-", "_"))()
                case None:
                    # Default model
                    model = "orb_v3_conservative_20_omat"
                case _:
                    raise ValueError(
                        "`model` must be a `DirectForcefieldRegressor`, pre-trained "
                        "model label (e.g. 'orb-v2'), or `None` (uses default, orb-v2)"
                    )
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+            model = model if model else "mattersim-v1.0.0-5M"
+
+            if isinstance(model, Module):
+                potential = model
+                model = "loaded_Module"
+            else:
+                potential = None
+
+            if isinstance(model, Path):
+                model = str(model)
```
oerc0122, line 322:
```suggestion
            potential = None
            match model:
                case Module():
                    potential = model
                    model = "loaded_Module"
                case Path() | str():
                    model = str(model)
                case None:
                    model = "mattersim-v1.0.0-5M"
```

janus_core/helpers/mlip_calculators.py
```diff
@@ -116,223 +123,253 @@ def choose_calculator(
...
+        case "equiformer" | "esen":
+            from fairchem.core import OCPCalculator, __version__
 
-    elif arch == "grace":
-        from tensorpotential.calculator import grace_fm
+            # Default model
+            if arch == "equiformer":
+                model = model if model else "EquiformerV2-31M-S2EF-OC20-All+MD"
+            if arch == "esen":
+                model = model if model else "eSEN-30M-OMAT24"
```
oerc0122, line 348:
```suggestion
            match arch, model:
                case ("equiformer", None):
                    model = "EquiformerV2-31M-S2EF-OC20-All+MD"
                case ("esen", None):
                    model = "eSEN-30M-OMAT24"
                case _:
                    pass
```

oerc0122 (approved):
Potential given we're replacing this block with a match to replace more with matches. Unsure whether they all give extra clarity, but add suggestions for them all for internal consistency, add those you think are better, but otherwise happy for this to go in.

oerc0122 (approved):
If it works, I'm happy.

PR #427, opened 2025-02-14
add mattersim calculator fixes #425

### LINKED ISSUE #425: add mattersim
https://github.com/microsoft/mattersim

alinelena:


pyproject.toml
```diff
@@ -55,11 +55,15 @@ m3gnet = [
...
 ]
+mattersim = [
+    "mattersim == 1.1.1",
+]
 all = [
     "janus-core[alignn]",
     "janus-core[chgnet]",
     "janus-core[m3gnet]",
     "janus-core[sevennet]",
+      "janus-core[mattersim]",
```
ElliottKasoar, line 66:
```suggestion
    "janus-core[mattersim]",
```

ElliottKasoar:
Also to do:

ElliottKasoar (approved):

tests/test_single_point.py
```diff
@@ -40,7 +40,10 @@
 )
 def test_potential_energy(struct, expected, properties, prop_key, calc_kwargs, idx):
     """Test single point energy using MACE calculators."""
+    skip_extras("mace")
```
harveydevereux, line 43: are these mace skips needed everywhere?

oerc0122: Using `pytest.mark` and `pytest.param` properly, you can mark each of the `arch`s and skip appropriately using `-k "not mace and not nequip"` for example.

ElliottKasoar: > Using `pytest.mark` and `pytest.param` properly, you can mark each of the `arch`s and skip appropriately using `-k "not mace and not nequip"` for example.

Sorry, I'm not quite sure I understand how this would work?

`skip_extras` is generally used so we can give it the architecture, and it skips if the relevant module can't be imported. I think that's generally more compact for the tests where we test every `arch`, rather than skip statements for every `pytest.param`?

skip_extras("mace"), or equivalent, in theory could be almost everywhere, since in general we assume mace is imported to run the tests (otherwise every value would change), and in the past this was a safe assumption.

Since making mace optional, that's not really a practical/useful change, so instead I've added in to test_single_point, and test_mlip_calculators, where necessary, so they can be run in full with or without mace, which is what is tested by the CI.

ElliottKasoar: We used to have markers for the extras (`--run-extra-mlips`), but we ended up with duplicated code, and assumptions about which extras would be installed together, which may change now with more conflicts.

ElliottKasoar: Ah, sorry, maybe it's a misleading function name, particularly for a single architecture.

`skip_extras("mace")` doesn't skip importing MACE, it skips the test if MACE can't be imported.

This is why the general case is `skip_extras(arch)` - we can specify all the valid architectures on equal terms, but it will skip any tests that can't be run based on the currently installed models.

In terms of MACE becoming optional, you're right that it wasn't introduced here (it was in #415). The key difference is that until now, all models were still compatible with MACE. This meant that all tests could successfully be run with MACE installed, and given that the majority of tests still require it, that's wasn't a problem. (Even better, it meant tests were run against a couple of versions of torch).

MatterSim conflicts with MACE, so in order to be able to run the new tests added, I needed to make sure they could run without MACE. I could narrow down the scope of the MatterSim CI tests to only the functions explicitly using MatterSim, but this felt like a more general approach, allowing us to run against the full files, even if most of it is skipped.

I found it particularly useful for local testing, since specifying the correct combination of files and function names depending on the models installed would get quite tedious.

harveydevereux (approved):

tests/test_mlip_calculators.py
```diff
@@ -131,7 +134,7 @@ def test_invalid_arch():
 def test_invalid_model_path(arch, model_path):
     """Test error raised for invalid model_path."""
     skip_extras(arch)
-    with pytest.raises((ValueError, RuntimeError, KeyError)):
+    with pytest.raises((ValueError, RuntimeError, KeyError, AssertionError)):
```
oerc0122, line 137: Ok, I'm thinking this wants to be a bit more specific, I think adding a `pytest.raises` (hopefully with `match`) into the `parametrize` isn't asking too much. 

Probably for another PR.

PR #428, opened 2025-02-14
add dpa3 calculator fixes #411

### LINKED ISSUE #411: add dpa model
dpa is at alpha available here 
https://github.com/deepmodeling/deepmd-kit/tree/dpa3-alpha/deepmd in a separate branch.

alinelena:
still to clarify gpu/cpu

### DIFF
--- a/janus_core/helpers/mlip_calculators.py
+++ b/janus_core/helpers/mlip_calculators.py
@@ -244,14 +244,15 @@ def choose_calculator(
         from deepmd import __version__
         from deepmd.calculator import DP
 
-        if isinstance(model_path, Path):
-            model_path = str(model_path)
-        else:
+        # No default `model_path`
+        if model_path is None:
             raise ValueError(
                 "Please specify `model_path`, as there is no "
-                f"default model for {arch}"
-                f"eg https://bohrium-api.dp.tech/ds-dl/dpa3openlam-74ng-v3.zip"
+                f"default model for {arch} "
+                "e.g. https://bohrium-api.dp.tech/ds-dl/dpa3openlam-74ng-v3.zip"
             )
+
+        model_path = str(model_path)
 
         calculator = DP(model=model_path, **kwargs)
 
--- a/janus_core/helpers/mlip_calculators.py
+++ b/janus_core/helpers/mlip_calculators.py
@@ -240,6 +240,21 @@ def choose_calculator(
             model_path=model_path, device=device, **kwargs
         )
 
+    elif arch == "dpa3":
+        from deepmd import __version__
+        from deepmd.calculator import DP
+
+        if isinstance(model_path, Path):
+            model_path = str(model_path)
+        else:
+            raise ValueError(
+                "Please specify `model_path`, as there is no "
+                f"default model for {arch}"
+                f"eg https://bohrium-api.dp.tech/ds-dl/dpa3openlam-74ng-v3.zip"
+            )
+
+        calculator = DP(model=model_path, **kwargs)
+
     else:
         raise ValueError(
             f"Unrecognized {arch=}. Suported architectures "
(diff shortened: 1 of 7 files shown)


ElliottKasoar:
Also to do:

janus_core/helpers/mlip_calculators.py
```diff
@@ -224,6 +224,21 @@ def choose_calculator(
...
         calculator = SevenNetCalculator(model=model_path, device=device, **kwargs)
 
+    elif arch == "dpa3":
+        from deepmd import __version__
+        from deepmd.calculator import DP
+
+        if isinstance(model_path, Path):
+            model_path = str(model_path)
+        else:
+            raise ValueError(
```
ElliottKasoar, line 234: There's not necessarily a perfect way to handle this, but via the command line, something like 

```
janus singlepoint --arch dpa3 --struct tests/data/NaCl.cif --model-path ./invalid/path
```

gives raises error (the CLI input is a string, and this module only converts it to a `Path` if the path exists), which is misleading, as it makes it sound like `model_path` hasn't been set.

Perhaps we do want our own `Path.exists()` check, or just modify the error message to include making sure the path exists?

tests/test_mlip_calculators.py
ElliottKasoar, line 1: Would be good to include a similar check to the other MLIPs for an invalid path

ElliottKasoar (changes requested):
Generally looks great! Just minor changes, and documentation needed

ElliottKasoar (approved):
Looks good to me.

The only question is about setting the device. Arguably it's confusing to accept the device, if it's not used?

The only thing I could see was the the [integration with ASE](https://docs.deepmodeling.com/projects/deepmd/en/v3.0.0/third-party/ase.html) links to their page on [runtime variables](https://docs.deepmodeling.com/projects/deepmd/en/v3.0.0/env.html), but it's not particularly explicit.

alinelena:
i have a feeling that is tensorflow... way

janus_core/helpers/mlip_calculators.py
```diff
@@ -240,6 +240,22 @@ def choose_calculator(
...
+    elif arch == "dpa3":
+        from deepmd import __version__
+        from deepmd.calculator import DP
+
+        # No default `model_path`
+        if model_path is None:
+            raise ValueError(
+                "Please specify `model_path`, as there is no "
+                f"default model for {arch} "
+                "e.g. https://bohrium-api.dp.tech/ds-dl/dpa3openlam-74ng-v3.zip"
```
harveydevereux, line 252: Is this really the official release? Also naively Googling, I can't find any context on this zip link. 

Is it likely to go out of date and then we forget to update it?

oerc0122 (approved):
Looks ok to me once you've addressed @harveydevereux 's concerns.

janus_core/helpers/mlip_calculators.py
harveydevereux: ok I see it now, guess that's just how it is

harveydevereux (approved):

janus_core/helpers/mlip_calculators.py
oerc0122: I do think it's worth adding that link in a comment.

janus_core/helpers/mlip_calculators.py
```diff
@@ -240,6 +240,22 @@ def choose_calculator(
             model_path=model_path, device=device, **kwargs
         )
 
+    elif arch == "dpa3":
+        from deepmd import __version__
+        from deepmd.calculator import DP
+
+        # No default `model_path`
+        if model_path is None:
+            raise ValueError(
```
ElliottKasoar, line 249:
```suggestion
            # From https://matbench-discovery.materialsproject.org/models/dpa3-v1-mptrj
            raise ValueError(
```

oerc0122 (approved):

PR #437, opened 2025-02-20
Prevent temp steps less than timestep

### LINKED ISSUE #436: Temperature ramp step time shorter than 1 timestep allowed
Currently, temperature ramp step times are allowed to be less than 1 timestep and are rounded down to 0, even though a step time of 0 is not allowed. This leads to step 0 being repeated at different temperatures, without any dynamics.

k-harris27:
Fixes #436 

Adds a check to raise an error if the temperature step time is shorter than 1 timestep. Also prints a debug message if the temperature step time is rounded significantly.

Tests have been added to verify the error is raised correctly, and existing tests using temperature step times shorter than 1 timestep have been updated.

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -429,7 +429,7 @@ class MolecularDynamics(BaseCalculation):
                     "Temperature ramp requested for ensemble with no thermostat"
                 )
 
-            # Check validate start and end temperatures
+            # Validate start and end temperatures
             if self.temp_start < 0 or self.temp_end < 0:
                 raise ValueError("Start and end temperatures must be positive")
 
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -423,14 +423,20 @@ class MolecularDynamics(BaseCalculation):
                 "heating to run",
                 stacklevel=2,
             )
-        if self.ramp_temp and self.ensemble in ("nve", "nph"):
-            raise ValueError(
-                "Temperature ramp requested for ensemble with no thermostat."
-            )
+        if self.ramp_temp:
+            if self.ensemble in ("nve", "nph"):
+                raise ValueError(
+                    "Temperature ramp requested for ensemble with no thermostat"
+                )
 
-        # Check validate start and end temperatures
-        if self.ramp_temp and (self.temp_start < 0 or self.temp_end < 0):
-            raise ValueError("Start and end temperatures must be positive")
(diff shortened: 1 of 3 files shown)


janus_core/calculations/md.py
```diff
@@ -423,14 +423,20 @@ def __init__(
...
-        if self.ramp_temp and (self.temp_start < 0 or self.temp_end < 0):
-            raise ValueError("Start and end temperatures must be positive")
+            # Check validate start and end temperatures
+            if self.temp_start < 0 or self.temp_end < 0:
+                raise ValueError("Start and end temperatures must be positive")
+
+            if self.temp_time < self.timestep:
+                raise ValueError(
+                    "Temperature ramp step time cannot be less than 1 timestep"
+                )
```
ElliottKasoar, line 439: I think when we discussed this, @oerc0122 was suggesting it might be better to warn, rather than fail.

My concern with that is, most people probably won't see the warning (which is captured by the log).

I think it really depends whether there are any circumstances where just having step 0 for each temperature may be useful.

janus_core/calculations/md.py
```diff
@@ -1106,6 +1112,13 @@ def _run_dynamics(self) -> None:
         if self.ramp_temp:
             heating_steps = int(self.temp_time // self.timestep)
 
+            if self.logger and not np.isclose(self.temp_time % self.timestep, 0.0):
```
ElliottKasoar, line 1115: Just to note, this can give slightly strange results e.g. `100 % 0.1` - > `0.09999999999999445`, which so this returns `False` with the default tolerances.

It's not necessarily a problem, since the log info doesn't hurt, and for most sensible time steps, this doesn't seem to be an issue.

janus_core/calculations/md.py
```diff
@@ -423,14 +423,20 @@ def __init__(
...
+        if self.ramp_temp:
+            if self.ensemble in ("nve", "nph"):
+                raise ValueError(
+                    "Temperature ramp requested for ensemble with no thermostat"
+                )
 
-        # Check validate start and end temperatures
-        if self.ramp_temp and (self.temp_start < 0 or self.temp_end < 0):
-            raise ValueError("Start and end temperatures must be positive")
+            # Check validate start and end temperatures
```
ElliottKasoar, line 432:
```suggestion
            # Validate start and end temperatures
```
Not your typo but while we're at it

ElliottKasoar (approved):

PR #444, opened 2025-02-24
Add MD progress bar with class

### LINKED ISSUE #405: Use progress bar for MD

k-harris27:
Resolves #405

Reworks the progress bar approach to extend `rich.progress.Progress`, just providing default formatting. This way, we can access all of the features of the progress bar, rather than just the iterator mode. Previous progress bars for phonons and single-point calculations still work in essentially the same way.

The MD progress bar reads the number of completed steps each time it is updated, ensuring it remains accurate. The iterator mode with `Progress.track` was unstable in the case of e.g. an MD run of length 0. The frequency of updates of the progress bar can be chosen with the `progress_bar_update_every` argument.

e.g. 30 steps of temperature ramp, 7 steps of MD:
image

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -10,8 +10,12 @@ from math import isclose
 from os.path import getmtime
 from pathlib import Path
 import random
-from typing import Any
+from typing import TYPE_CHECKING, Any
 from warnings import warn
+
+# Only needed for type hints
+if TYPE_CHECKING:
+    from rich.progress import TaskID
 
 from ase import Atoms
 from ase.geometry.analysis import Analysis
@@ -43,7 +47,12 @@ from janus_core.helpers.janus_types import (
     PostProcessKwargs,
 )
 from janus_core.helpers.struct_io import input_structs, output_structs
-from janus_core.helpers.utils import none_to_dict, set_minimize_logging, write_table
+from janus_core.helpers.utils import (
+    ProgressBar,
+    none_to_dict,
+    set_minimize_logging,
+    write_table,
+)
 from janus_core.processing.correlator import Correlation
 from janus_core.processing.post_process import compute_rdf, compute_vaf
 
@@ -157,6 +166,11 @@ class MolecularDynamics(BaseCalculation):
     seed
         Random seed used by numpy.random and random functions, such as in Langevin.
         Default is None.
+    enable_progress_bar
+        Whether to show a progress bar. Default is False.
+    update_progress_every
+        How many timesteps between progress bar updates.
+        Default is steps/100, rounded up.
 
     Attributes
     ----------
@@ -218,6 +232,8 @@ class MolecularDynamics(BaseCalculation):
         post_process_kwargs: PostProcessKwargs | None = None,
         correlation_kwargs: list[CorrelationKwargs] | None = None,
         seed: int | None = None,
+        enable_progress_bar: bool = False,
+        update_progress_every: int | None = None,
     ) -> None:
         """
         Initialise molecular dynamics simulation configuration.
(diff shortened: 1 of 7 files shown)


ElliottKasoar (commented):
I'm not sure this is quite working as expected to, e.g. if I run 

```
janus md --ensemble nvt --temp-time 20 --temp-start 0 --temp-end 20 --temp-step 10 --traj-every 1 --stats-every 1 --struct tests/data/NaCl.cif --steps 50
```

This pauses at 40/110 and seems to end at 90/110.

Ah, this is probably a quirk of how we treat a ramp starting at 0K, sp not a general issue, but still something to handle.

I do also wonder if we ought to separate out heating and MD, and maybe even different stages of heating?

k-harris27:
> Ah, this is probably a quirk of how we treat a ramp starting at 0K, so not a general issue, but still something to handle.

Ah thanks, I had missed that 0K steps were skipped. This should be handled properly now.

> I do also wonder if we ought to separate out heating and MD, and maybe even different stages of heating?

I felt like there should be an overall progress bar so that a user can get an estimate of the remaining total simulation time. I've added a second progress bar that tracks the current temperature step if applicable. Let me know what you think, I'm happy to adjust it of course.

janus_core/calculations/md.py
```diff
@@ -218,6 +231,8 @@ def __init__(
         post_process_kwargs: PostProcessKwargs | None = None,
         correlation_kwargs: list[CorrelationKwargs] | None = None,
         seed: int | None = None,
+        enable_progress_bar: bool = False,
+        progress_bar_update_every: int = 1,
```
ElliottKasoar, line 235: We may want to consider a higher default, assuming most realistic use cases will be doing thousands - millions of steps?

tests/test_md_cli.py
ElliottKasoar, line 1: We should probably also test both the progress bar for heating, and heating + MD.

janus_core/calculations/md.py
```diff
@@ -372,6 +391,8 @@ def __init__(
         self.post_process_kwargs = post_process_kwargs
         self.correlation_kwargs = correlation_kwargs
         self.seed = seed
+        self.enable_progress_bar = enable_progress_bar
+        self.progress_bar_update_every = progress_bar_update_every
```
ElliottKasoar, line 395: We may want a slightly more concise name if possible. `update_progress_every` maybe?

janus_core/calculations/md.py
```diff
@@ -1044,6 +1074,89 @@ def _set_target_temperature(self, temperature: float):
         else:
             raise ValueError("Temperature set for ensemble with no thermostat.")
 
+    def _init_progress_bar(self) -> ProgressBar:
+        """
+        Initialise MD progress bar.
+
+        Returns
+        -------
+        progress_bar
```
ElliottKasoar, line 1083:
```suggestion
        ProgressBar
```
See: https://numpydoc.readthedocs.io/en/latest/format.html#returns

ElliottKasoar (changes requested):
Looks good!

In terms of the specific format of the bars, perhaps we should discuss it tomorrow. My inclination would be a new bar for each stage, rather than a total bar and a current bar that changes what it's referring to, which is what I think it currently does.

> Also, I never managed to reproduce anything like this. Let me know if you're still seeing it and I can have another look.

Ah, so I think the pause was the switch from the heat ramp to the MD, which was correctly as step 40, and then the 50 MD steps took it to 90, so it was just the total that was wrong, as it included 20 steps at 0K.

k-harris27:
This will need rebasing and checking once #453 is merged. I'm happy to do that, even if it's sometime next week.

ElliottKasoar:
I'm not sure this quite works for a simulation that is only heating.

```
E.g. `janus md --ensemble nvt --temp-time 200 --temp-start 10 --temp-end 20 --temp-step 10  --struct tests/data/NaCl.cif  ```

start showing 0/400, and 0/200 correctly, but doesn't update, then finally updates to 400/400 and 0/0.

ElliottKasoar:
Superseded by #521

PR #445, opened 2025-02-24
Add CI test against development versions

k-harris27:
This should test against up-to-date development versions of ASE and MACE during CI.

It shouldn't fail the CI if the tests fail, so that we can have a warning about upcoming breaking changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,38 +37,6 @@ jobs:
           file: coverage.xml
           base-path: janus_core
 
-  test-dev-versions:
-    continue-on-error: true
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    strategy:
-      matrix:
-        python-version: ["3.12"]
-
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Install uv
-        uses: astral-sh/setup-uv@v4
-        with:
-          version: "0.5.7"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -37,6 +37,38 @@ jobs:
           file: coverage.xml
           base-path: janus_core
 
+  test-dev-versions:
```
alinelena, line 40: I will do this as a separate ci-dev.yml, and also make it really run only on the main repo no need to bother the forks with it.

.github/workflows/ci.yml
```diff
@@ -37,6 +37,38 @@ jobs:
...
+      matrix:
+        python-version: ["3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "0.5.7"
```
alinelena, line 54: also maybe will be good to use latest uv in here?

k-harris27:
As a note: we were unsure if `continue-on-error` would mark the workflow as "passed" even if the job failed. Testing on a minimal repo seems to suggest it does, so if this fails it shouldn't prevent any merges.

https://github.com/k-harris27/actions-test/actions/runs/13501767954

ElliottKasoar (approved):

PR #446, opened 2025-02-24
Combine struct and struct_path parameters

### LINKED ISSUE #412: Merge struct and struct_path parameters?
Currently, all calculations take either `struct` and `struct_path`, but if the wrong one is used, the error message isn't particularly helpful e.g.:

```python
E       AttributeError: 'Atoms' object has no attribute 'read'
```

We could check the types/catch errors like this, but perhaps we should simply combine them into a single `struct` parameter, and use the type check to determine how to handle them?

ElliottKasoar:
Resolves #412


janus_core/calculations/eos.py
```diff
@@ -104,8 +100,7 @@ class EoS(BaseCalculation):
 
     def __init__(
         self,
-        struct: Atoms | None = None,
-        struct_path: PathLike | None = None,
+        struct: Atoms | PathLike,
```
alinelena, line 103:
```suggestion
        struct: MaybeSequence[Atoms] | PathLike,
```

janus_core/calculations/geom_opt.py
```diff
@@ -95,8 +92,7 @@ class GeomOpt(BaseCalculation):
 
     def __init__(
         self,
-        struct: Atoms | None = None,
-        struct_path: PathLike | None = None,
+        struct: Atoms | PathLike,
```
alinelena, line 95: as above

janus_core/calculations/md.py
```diff
@@ -177,8 +174,7 @@ class MolecularDynamics(BaseCalculation):
 
     def __init__(
         self,
-        struct: Atoms | None = None,
-        struct_path: PathLike | None = None,
+        struct: Atoms | PathLike,
```
alinelena, line 177: as above

janus_core/calculations/eos.py
ElliottKasoar: Equation of state doesn't (currently) accept multiple input structures, so I think just `Atoms` is correct here

janus_core/calculations/geom_opt.py
ElliottKasoar: This also only works for a single structure (also true for the phonons/EoS/geom opt: you can give it a trajectory file, but it only reads one, whereas descriptors and single point read the whole file by default).

janus_core/calculations/md.py
ElliottKasoar: Same as the others, I think this is fine

alinelena (approved):

Issue #359, opened 2024-11-28
Add upload option


ElliottKasoar:
Integration with PSDI to upload results to repository.

This could link to the suggestion in #128, as we will need the list of output files in order to upload them.

alinelena:
we will also need the api for the psdi repo.. I will speak with Andy about

alinelena:
https://inveniordm.docs.cern.ch/reference/rest_api_index/

ElliottKasoar:
Discussions/proposals with @alinelena, @oerc0122 and @k-harris27:

- Create base Invenio class, based on @alinelena's proof of concept
  - Create subclass for Zenodo
  - Create subclass for PSDI
- Create `janus push` and `janus pull` options to upload and download respectively
  - Create `janus push --dry-run`, which only creates archives, with no upload
  - We could have an archive option for calculations, but in general archive/upload is for final sets of results, not intermediates
  - Include option to upload model, probably set to False by default 
- We need to figure out how to specify uploading results from multiple calculations
- We probably want to have an option to create a results directory for calculations
  - and/or list all results files in log, and extract from this

PR #452, opened 2025-02-27
Add file prefix options to singlepoint and geomopt

k-harris27:
Adds the option to provide a file prefix that is used for all of the files produced by single point and geometry optimisations. Since this can include a directory, it helps to allow the option to create a results directory for https://github.com/stfc/janus-core/issues/359#issuecomment-2684974380.

### DIFF
--- a/janus_core/calculations/geom_opt.py
+++ b/janus_core/calculations/geom_opt.py
@@ -58,6 +58,8 @@ class GeomOpt(BaseCalculation):
         Default is True if attach_logger is True, else False.
     tracker_kwargs
         Keyword arguments to pass to `config_tracker`. Default is {}.
+    file_prefix
+        Prefix for output filenames. Default is inferred from structure.
     fmax
         Set force convergence criteria for optimizer in units eV/Å. Default is 0.1.
     steps
@@ -85,6 +87,8 @@ class GeomOpt(BaseCalculation):
     write_kwargs
         Keyword arguments to pass to ase.io.write to save optimized structure.
         Default is {}.
+    write_trajectory
+        Whether to save a trajectory file of optimization frames.
     traj_kwargs
         Keyword arguments to pass to ase.io.write to save optimization trajectory.
         Must include "filename" keyword. Default is {}.
@@ -103,6 +107,7 @@ class GeomOpt(BaseCalculation):
         log_kwargs: dict[str, Any] | None = None,
         track_carbon: bool | None = None,
         tracker_kwargs: dict[str, Any] | None = None,
+        file_prefix: PathLike | None = None,
         fmax: float = 0.1,
         steps: int = 1000,
         symmetrize: bool = False,
@@ -114,6 +119,7 @@ class GeomOpt(BaseCalculation):
         opt_kwargs: ASEOptArgs | None = None,
         write_results: bool = False,
         write_kwargs: OutputKwargs | None = None,
+        write_trajectory: bool = False,
         traj_kwargs: OutputKwargs | None = None,
     ) -> None:
         """
@@ -146,6 +152,8 @@ class GeomOpt(BaseCalculation):
             Default is True if attach_logger is True, else False.
         tracker_kwargs
             Keyword arguments to pass to `config_tracker`. Default is {}.
+        file_prefix
+            Prefix for output filenames. Default is inferred from structure.
         fmax
             Set force convergence criteria for optimizer in units eV/Å. Default is 0.1.
         steps
@@ -173,9 +181,11 @@ class GeomOpt(BaseCalculation):
         write_kwargs
             Keyword arguments to pass to ase.io.write to save optimized structure.
             Default is {}.
+        write_trajectory
+            Whether to save a trajectory file of optimization frames.
         traj_kwargs
             Keyword arguments to pass to ase.io.write to save optimization trajectory.
-            Must include "filename" keyword. Default is {}.
+            "filename" keyword is inferred from file_prefix if not given. Default is {}.
         """
         read_kwargs, filter_kwargs, opt_kwargs, write_kwargs, traj_kwargs = (
             none_to_dict(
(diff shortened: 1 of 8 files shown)


ElliottKasoar:
Do we want to set a default file prefix, as part of this PR or otherwise?

In that case, we may want an additional option, to specify the output directory, which prepends the file prefix?

k-harris27:
> In that case, we may want an additional option, to specify the output directory, which prepends the file prefix?

I feel like an output directory option that doesn't affect the file name would be a good addition. I was thinking that making sure all the files work consistently with `file_prefix` should help make a directory option relatively easy to add.

janus_core/calculations/geom_opt.py
```diff
@@ -85,6 +87,8 @@ class GeomOpt(BaseCalculation):
     write_kwargs
         Keyword arguments to pass to ase.io.write to save optimized structure.
         Default is {}.
+    write_trajectory
```
ElliottKasoar, line 90: I'm open to either way, but maybe `write_traj`, since we have `traj_kwargs` etc.?

ElliottKasoar (commented):
Thanks, @k-harris27!

Could you:

- Summarise the changes to how the trajectory file can now be saved (in the comments here)
- Update the CLI docs (docs/source/user_guide/command_line.rst) to make sure the example is correct

k-harris27:
> Summarise the changes to how the trajectory file can now be saved (in the comments here)

In `geomopt`, the trajectory is now written if the `write_traj` option is `True`.

By default, the trajectory filename is inferred from `file_prefix` as with other output files, and only a .extxyz file is written.

The filename can be specified manually by setting `traj_kwargs["filename"]`.

The binary .traj file written by ASE can be retained by manually setting its output path using `opt_kwargs["trajectory"]`.

If `traj_kwargs["filename"]` or `opt_kwargs["trajectory"]` are given when `write_traj` is `False`, an error is raised.

janus_core/calculations/geom_opt.py
```diff
@@ -223,17 +226,37 @@ def __init__(
...
+                # Overwrite .traj binary with .extxyz by default.
+                self.opt_kwargs["trajectory"] = str(self.traj_kwargs["filename"])
+        else:
+            if self.traj_kwargs:
+                raise ValueError(
+                    "traj_kwargs given, but trajectory writing not enabled."
+                )
+            if "trajectory" in self.opt_kwargs:
+                raise ValueError(
+                    '"trajectory" given in opt_kwargs,'
```
ElliottKasoar, line 256:
```suggestion
                    '"trajectory" given in opt_kwargs, '
```

janus_core/cli/geomopt.py
```diff
@@ -194,11 +184,16 @@ def geomopt(
...
     out
         Path to save optimized structure, or last structure if optimization did not
         converge. Default is inferred from name of structure file.
-    traj
-        Path if saving optimization frames. Default is None.
+    write_traj
+        Whether to save a trajectory file of optimization frames.
+    traj_kwargs
+        Keyword arguments to pass to ase.io.write when saving trajectory.
+        If "filename" is not included, it is inferred from --file-prefix.
```
ElliottKasoar, line 196:
```suggestion
        If "filename" is not included, it is inferred from `file_prefix`.
```
I normally just refer to the Python variables

janus_core/calculations/geom_opt.py
```diff
@@ -223,17 +226,37 @@ def __init__(
...
 
         if not self.struct.calc:
             raise ValueError("Please attach a calculator to `struct`.")
 
-        # Set output file
+        # Set output files
         self.write_kwargs.setdefault("filename", None)
         self.write_kwargs["filename"] = self._build_filename(
             "opt.extxyz", filename=self.write_kwargs["filename"]
         ).absolute()
```
oerc0122, line 239: I know this isn't you, probably me. Why set default, then immediately overwrite it?
```suggestion
        self.write_kwargs["filename"] = self._build_filename(
            "opt.extxyz", filename=self.write_kwargs.get("filename")
        ).absolute()
```

janus_core/cli/geomopt.py
ElliottKasoar, line 1: How useful is it to be able to define `traj_kwargs` directly, rather than through `minimize_kwargs`, when we can't do the same for `opt_kwargs`, `filter_kwargs`, etc?

It does make changing the filename a bit easier, but equally, with `file_prefix` added as well, we have more control than we did in the past about how the trajectory is stored.

janus_core/calculations/geom_opt.py
ElliottKasoar: Ah I think this was me, actually.

This construction was added in various places when we were allowing the file builder to create directories, whether or not the filename has been given, but I think it was just because I hadn't really thought about using `.get`.

janus_core/cli/geomopt.py
```diff
@@ -48,6 +47,9 @@ def _set_minimize_kwargs(
     if "opt_kwargs" not in minimize_kwargs:
         minimize_kwargs["opt_kwargs"] = {}
 
+    if "traj_kwargs" not in minimize_kwargs:
+        minimize_kwargs["traj_kwargs"] = {}
```
oerc0122, line 51:
```suggestion
    minimize_kwargs.setdefault("opt_kwargs", {})
    minimize_kwargs.setdefault("traj_kwargs", {})
```

janus_core/calculations/geom_opt.py
```diff
@@ -223,17 +226,32 @@ def __init__(
...
 
+        if self.write_traj:
+            if "filename" not in self.traj_kwargs:
+                self.traj_kwargs["filename"] = self._build_filename(
+                    "traj.extxyz"
+                ).absolute()
+            if "trajectory" not in self.opt_kwargs:
+                # Overwrite .traj binary with .extxyz by default.
+                self.opt_kwargs["trajectory"] = str(self.traj_kwargs["filename"])
```
oerc0122, line 248:
```suggestion
            self.traj_kwargs.setdefault(
                "filename", 
                self._build_filename("traj.extxyz").absolute()
            )
            self.opt_kwargs.setdefault("trajectory", str(self.traj_kwargs["filename"]))
```

oerc0122 (approved):

tests/test_geomopt_cli.py
```diff
@@ -60,6 +60,7 @@ def test_geomopt():
     finally:
         # Ensure files deleted if command fails
         read_atoms(results_path)
+        results_path.unlink(missing_ok=True)
```
ElliottKasoar, line 63: I think this is handled by `read_atoms` anyway?

k-harris27: Oh, sure. I'm not sure what happened but I ended up with the results file surviving the test at some point, so the test was failing on the beginning assertion. Happy to remove that line if it doesn't help anything.

ElliottKasoar: I think the only thing that should happen is a file different to the expected file is created, or one is left over from runs outside of the testing, but doesn't matter too much.

We can leave it, it doesn't hurt.

janus_core/calculations/geom_opt.py
k-harris27: I haven't touched that in my edits yet, so that sounds good, cheers.

ElliottKasoar (approved):

PR #454, opened 2025-02-28
List outputs

ElliottKasoar:
Adds `output_files` to all calculations, which is recorded in the summary file generated by the CLI.

~Will need rebasing after #452~

To do:

- [x] Add tests


janus_core/calculations/neb.py
```diff
@@ -325,49 +327,71 @@ def __init__(
...
+            self.write_kwargs["filename"] if self.write_band else None
+        )
+        output_files["minimized_initial_structure"] = (
+            self.init_struct_min_path if self.minimize else None
+        )
+        output_files["minimized_final_structure"] = (
+            self.final_struct_min_path if self.minimize else None
+        )
+
+        return output_files
```
oerc0122, line 393:
```suggestion
        return {
            "log": self.log_kwargs["filename"] if self.logger else None,
            "results": self.results_file if self.write_results else None,
            "plot": self.plot_file if self.plot_band else None,
            "band": (
                     self.write_kwargs["filename"] if self.write_band else None
            ),
            "minimized_initial_structure": (
                self.init_struct_min_path if self.minimize else None
            ),
            "minimized_final_structure": (
                self.final_struct_min_path if self.minimize else None,
            ),
        }
```

janus_core/calculations/phonons.py
```diff
@@ -376,6 +387,80 @@ def calcs(self, value: MaybeSequence[PhononCalcs]) -> None:
...
+        output_files["thermal"] = (
+            self.thermal_file
+            if self.write_results and "thermal" in self.calcs
+            else None
+        )
+        output_files["minimized_initial_structure"] = (
+            self.minimize_kwargs["write_kwargs"]["filename"] if self.minimize else None
+        )
+
+        return output_files
```
oerc0122, line 462:
```suggestion
        return {
            "log": self.log_kwargs["filename"] if self.logger else None,
            "params": self.phonopy_file if self.write_results else None,
            "force_constants": (
                self.force_consts_file if self.force_consts_to_hdf5 else None
            ),
            "bands": (
                self.bands_file if self.write_results and "bands" in self.calcs else None
            ),
            "bands_plot": self.bands_plot_file if self.plot_to_file else None,
            "dos": (
                self.dos_file if self.write_results and "dos" in self.calcs else None
            ),
            "dos_plot": (
                self.dos_plot_file if self.plot_to_file and "dos" in self.calcs else None
            ),
            "band_dos_plot": (
                self.bands_dos_plot_file,
                if self.plot_to_file and "dos" in self.calcs and "bands" in self.calcs
                else None
            ),
            "pdos": (
                self.pdos_file if self.write_results and "pdos" in self.calcs else None
            ),
            "pdos_plot": (
                self.pdos_plot_file if self.plot_to_file and "pdos" in self.calcs else None
            ),
            "thermal": (
                self.thermal_file,
                if self.write_results and "thermal" in self.calcs
                else None
            ),
            "minimized_initial_structure": (
                self.minimize_kwargs["write_kwargs"]["filename"] if self.minimize else None
            ),
        }
```

janus_core/calculations/single_point.py
```diff
@@ -227,6 +226,25 @@ def properties(self, value: MaybeSequence[Properties]) -> None:
...
+            Output file labels and paths.
+        """
+        output_files = {}
+
+        output_files["log"] = self.log_kwargs["filename"] if self.logger else None
+        output_files["results"] = (
+            self.write_kwargs["filename"] if self.write_results else None
+        )
+
+        return output_files
```
oerc0122, line 246:
```suggestion
        return {
            "log": self.log_kwargs["filename"] if self.logger else None,
            "results": (
                self.write_kwargs["filename"] if self.write_results else None
            ),
        }
```

oerc0122 (commented):
I do wonder if at this point files ought to become descriptors. Intercepting their `__set__` with building the appropriate filenames and extensions and maybe even building the output files `dict`.

ElliottKasoar:
New summary looks something like:

```
command: janus singlepoint
inputs:
  attach_logger: true
  calc:
    arch: mace_mp
    calc_kwargs: {}
    device: cpu
    model_path: null
    read_kwargs: {}
  enable_progress_bar: true
  file_prefix: null
  log: /Users/elliottkasoar/Documents/PSDI/janus-core/NaCl-singlepoint-log.yml
  properties: null
  struct:
    formula: Cl4Na4
    n_atoms: 8
    struct_path: tests/data/NaCl.cif
  track_carbon: true
  write_kwargs: {}
  write_results: true
output_files:
  log: /Users/elliottkasoar/Documents/PSDI/janus-core/NaCl-singlepoint-log.yml
  results: /Users/elliottkasoar/Documents/PSDI/janus-core/NaCl-results.extxyz
  summary: /Users/elliottkasoar/Documents/PSDI/janus-core/NaCl-singlepoint-summary.yml
start_time: 03/03/2025, 16:51:11
emissions: 2.3386228257216414e-07
end_time: 03/03/2025, 16:51:23
```

In most cases, the output files are relatively simple to predict, so I write them before running the calculation.

The main exceptions are:

- Phonons, where there are multiple Python entry points, so we cannot guarantee the output files will be generated in general, but for the CLI, we can manually specify this
- MD, where the restart files generated depends on how far the simulation gets
  - We could predict the final set of results files generated, but instead I update the list after MD completes, so in general, it should be accurate, but won't help if errors occur.

ElliottKasoar:
> I do wonder if at this point files ought to become descriptors. Intercepting their `__set__` with building the appropriate filenames and extensions and maybe even building the output files `dict`.

What would be the advantage of this?

Also, would it work for filenames that depend on things we only easily know at initialisation?

oerc0122:
> > I do wonder if at this point files ought to become descriptors. Intercepting their `__set__` with building the appropriate filenames and extensions and maybe even building the output files `dict`.
> 
> What would be the advantage of this?
> 
> Also, would it work for filenames that depend on things we only easily know at initialisation?

The advantage would be that filenames are declared as extant on the class with a single definition and potentially avoiding a lot of the `build_file(file, **kwargs) if file else None` throughout.

The disadvantages are increased code complexity (comprehending descriptors), and a number of special cases which may not be helped and it may not save anything at all in the end.

When setting the value of the attribute the parent instance is known, but local function arguments (except the values) aren't. So it depends a little. 

It was more a passing thought I noted midway (and then submitted as my only comment making it seem more important).

janus_core/calculations/eos.py
```diff
@@ -246,27 +246,46 @@ def __init__(
...
+        """
+        Dictionary of output file labels and paths.
+
+        Returns
+        -------
+        dict[str, PathLike]
+            Output file labels and paths.
+        """
+        return {
+            "log": self.log_kwargs["filename"] if self.logger else None,
```
oerc0122, line 280: I certainly don't think this should be blocking, but I am half-wondering whether we want `@property`s for these which check if their respective trigger is valid to remove this pattern everywhere.

Not this PR, we should get this in.

janus_core/calculations/md.py
```diff
@@ -705,6 +739,82 @@ def _restart_file(self) -> str:
...
+        Define correlations file name.
+
+        Returns
+        -------
+        str
+           File name for correlations files.
+        """
+        return self._build_filename("cor.dat", self.param_prefix)
+
+    @property
```
oerc0122, line 754: Like this approach! Much cleaner.

janus_core/calculations/phonons.py
```diff
@@ -262,9 +262,9 @@ def __init__(
         self.displacement = displacement
         self.displacement_kwargs = displacement_kwargs
         self.mesh = mesh
-        self.symmetrize = symmetrize
```
oerc0122, line 265: For alphabetical ordering?

janus_core/cli/md.py
```diff
@@ -27,6 +28,28 @@
...
+    summary
+        Path to summary file.
+    restart_files
+        Restart files generated.
+    """
+    with open(summary, encoding="utf8") as file:
+        summary_info = yaml.safe_load(file)
+
+    summary_info["output_files"]["restarts"] = (
+        [str(file) for file in restart_files] if restart_files else None
```
oerc0122, line 46: Would the empty list here be OK?

oerc0122 (approved):
Looks good, happy for it to go in, just a few stylistic suggestions.

janus_core/calculations/phonons.py
ElliottKasoar: In general I try to set the attributes in the order the parameters are passed where possible. In this case, that should be before minimise and minimise_kwargs, so not completely sure what happened...

At some point I need to do a pass on Python and CLI parameter orderings anyway to bring more consistency again anyway, so I'll fix this properly then.

tests/utils.py
```diff
@@ -145,3 +145,34 @@ def skip_extras(arch: str):
...
+        Expected output files to compare with summary.
+    """
+    assert "output_files" in summary
+
+    for key, value in output_files.items():
+        output_file = summary["output_files"][key]
+        if isinstance(value, list):
+            for file in value:
+                if not file.exists():
+                    raise ValueError(f"{file} missing")
```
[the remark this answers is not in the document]
ElliottKasoar: What do you mean, given pytest? I think the output is similar either way:

```
FAILED tests/test_singlepoint_cli.py::test_summary - AssertionError: /private/var/folders/p8/6555_ljx3zx2m306t4nzkfp00000gn/T/pytest-of-elliottkasoar/pytest-326/test_summary0/summary.yml missing
```

or

```
FAILED tests/test_singlepoint_cli.py::test_summary - ValueError: /private/var/folders/p8/6555_ljx3zx2m306t4nzkfp00000gn/T/pytest-of-elliottkasoar/pytest-325/test_summary0/summary.yml missing
```

but happy either way.

janus_core/cli/md.py
ElliottKasoar: Yep, we end up with `restarts: null` etc in the summary, which I think is fine

tests/utils.py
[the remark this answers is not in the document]
oerc0122: Ah, sometimes using `assert` can give more info from `pytest` as it overloads certain `AssertionError` interpolation.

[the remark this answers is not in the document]
oerc0122: In one sense it makes it obvious it's a testing error not an internal one?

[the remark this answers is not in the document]
ElliottKasoar: Ah I see, I think it's fine then. I think I actually changed it from an `assert` originally because it didn't output anything useful, although that's before I added the error message, which of course you can include in the same way as in your suggestion.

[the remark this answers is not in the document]
ElliottKasoar: Ah true, hmm.

alinelena:
ok, what is left for this?

oerc0122 (approved):
Can update the description given it's now rebased.

PR #457, opened 2025-02-28
Show all MD progress bars during ramp

k-harris27:
This fundamentally works:

image

but still has some holes:
- Progress bar update frequencies that aren't a multiple of the task total won't complete the bars.
- Restarts have not been checked and definitely don't behave properly.

I'll leave it here as a draft so someone else can polish it if they want to.


ElliottKasoar:
#521 is essentially a continuation of this, but I thought it's better to open a new PR and keep this for reference, as some of the rebasing was quite involved.

PR #460, opened 2025-03-03
Fix publish

ElliottKasoar:
Reverts failed release and fixes installation for publishing


oerc0122 (approved):
Phew! This is a pretty big change. Are we sure they're all correct?

alinelena (approved):

Issue #128, opened 2024-05-03
md_summary prints null instead of default values.


alinelena:
```
  timestep: 1.0
  traj_append: false
  traj_every: 100
  traj_file: null
  restart_stem: null
  restarts_to_keep: 4
  rotate_restart: true
  seed: null
  ensemble: npt
  equil_steps: 0
  file_prefix: null
```

ElliottKasoar:
My initial thought was that things like `file_prefix` are not readily accessible, as it's not even a simple default value, but actually we can access all of the attributes of the ensemble when we create the summary, just before we call `.run`, so this shouldn't be too bad to fix.

Only issue is that it depends if we consider the summary to be from the CLI perspective (it doesn't set these) or the Python perspective.

It's not possible at the moment, but if, for example, we wanted to use the summary as a new config for a different structure (so overwrite the structure only), we may want to keep `file_prefix` as `None`, rather than using the file_prefix that's been created for the first structure.

ElliottKasoar:
Maybe what we should have is 

- `inputs`
  - All CLI options that you could take and use directly as a `--config`
  - We could even allow it to be read directly from summaries
- `info`
  - Extra details e.g. `formula`, `n_atoms`, start/end time etc.)
- Maybe also `outputs`?
  - Files created
  - Slightly trickier in the case of optional minimization etc.

The first two would basically be rearranging what we already have, so maybe worth doing on their own, then `outputs` might be nice as an alternative to outputting the inferred `file_prefix`.

ElliottKasoar:
> Maybe what we should have is
> 
> * `inputs`
>   
>   * All CLI options that you could take and use directly as a `--config`
>   * I still think this should probably be the CLI input/default, rather than anything that is inferred in the Python
>   * We could even allow it to be read directly from summaries
> * `info`
>   
>   * Extra details e.g. `formula`, `n_atoms`, start/end time etc.)
> * Maybe also `outputs`?
>   
>   * Files created
>   * Slightly trickier in the case of optional minimization etc.
> 
> The first two would basically be rearranging what we already have, so maybe worth doing on their own, then `outputs` might be nice as an alternative to outputting the inferred `file_prefix`.

I think this is all implemented, between #454 and #463.

PR #463, opened 2025-03-05
Save input configuration

ElliottKasoar:
This saves the input parameters to CLI commands in the summary.

As described in #128, I think it makes most sense to save the config before any defaults are applied.

This is partly because where defaults are set varies between the CLI and Python modules for each command, accessing the "final" value of each parameter would not be trivial.

It also ensures consistency if this is used for a new config file, since in some cases, whether the parameter is using a default or has been explicitly set change behaviours, which would no longer be distinguishable if we stored the default.

In combination with #454. hopefully this won't be a problem in addressing the original #128 issue.

E.g.:

```
janus geomopt --struct tests/data/NaCl.cif --arch mace_mp --model-path small --opt-cell-lengths --write-traj --minimize-kwargs "{'opt_kwargs':{'trajectory':'test.traj'}}"
```

produces

```
command: janus geomopt
config:
  arch: mace_mp
  calc_kwargs: {}
  device: cpu
  file_prefix: null
  filter_func: null
  fmax: 0.1
  log: null
  minimize_kwargs:
    opt_kwargs:
      trajectory: test.traj
  model_path: small
  opt_cell_fully: false
  opt_cell_lengths: true
  optimizer: LBFGS
  out: null
  pressure: 0.0
  read_kwargs: {}
  steps: 1000
  struct: tests/data/NaCl.cif
  summary: null
  symmetrize: false
  symmetry_tolerance: 0.001
  tracker: true
  write_kwargs: {}
  write_traj: true
info:
  struct:
    formula: Cl4Na4
    n_atoms: 8
    struct_path: tests/data/NaCl.cif
start_time: 05/03/2025, 19:44:02
emissions: 4.319631282771114e-07
end_time: 05/03/2025, 19:44:14
```

### DIFF
--- a/janus_core/cli/descriptors.py
+++ b/janus_core/cli/descriptors.py
@@ -79,8 +79,7 @@ def descriptors(
     calc_per_atom
         Whether to calculate descriptors for each atom. Default is False.
     arch
-        MLIP architecture to use for single point calculations.
-        Default is "mace_mp".
+        MLIP architecture to use for calculations. Default is "mace_mp".
     device
         Device to run model on. Default is "cpu".
     model_path
@@ -111,8 +110,9 @@ def descriptors(
         carbon_summary,
         check_config,
         end_summary,
+        get_config,
+        get_struct_info,
         parse_typer_dicts,
-        save_struct_calc,
         start_summary,
     )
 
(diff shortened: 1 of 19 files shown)


alinelena:
this looks good... I wonder, unless @oerc0122 has any objections shall go in.

alinelena (approved):

PR #469, opened 2025-03-13
Fix temperature saved during MD

### LINKED ISSUE #433: specify temperature and current temperature in atoms.info
at the moment we save target temperature and pressure in the atoms.info
under temperature and pressure. Names can be improved.. seems to be a a bug (noticeable when you do heating)... recorded temp seems to be one step behind.

would be nice to add also the current temperature for example, but only if is not recomputed.

ElliottKasoar:
Resolves #433

Saves temperature (which ASE gets from momenta, so requires minimal calculation), and, where relevant, target temperature, to `Atoms.info`.

This is then written out to the trajectory, results and final files.

Also removes (target) pressure from the final state info, to be consistent with other outputs. We may want to similarly add in the real and target pressure.

To do:

- [x] Check the output from this against the stats temperatures, to identify discrepancies.

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -784,7 +784,7 @@ class MolecularDynamics(BaseCalculation):
         }
 
     @property
-    def unit_info(self) -> dict[str, str]:
+    def stats_units(self) -> dict[str, str]:
         """
         Get units of returned statistics.
 
@@ -847,8 +847,8 @@ class MolecularDynamics(BaseCalculation):
             write_table(
                 "ascii",
                 file=stats_file,
-                units=self.unit_info,
-                **{key: () for key in self.unit_info},
+                units=self.stats_units,
+                **{key: () for key in self.stats_units},
             )
 
     def _write_stats_file(self) -> None:
@@ -863,7 +863,7 @@ class MolecularDynamics(BaseCalculation):
             write_table(
                 "ascii",
                 file=stats_file,
-                units=self.unit_info,
+                units=self.stats_units,
                 formats=self.default_formats,
                 print_header=False,
                 **stats,
@@ -1313,7 +1313,7 @@ class NPT(MolecularDynamics):
         return stats
 
     @property
-    def unit_info(self) -> dict[str, str]:
+    def stats_units(self) -> dict[str, str]:
         """
         Get units of returned statistics.
 
@@ -1322,7 +1322,7 @@ class NPT(MolecularDynamics):
         dict[str, str]
             Units attached to statistical properties.
         """
-        return super().unit_info | {
+        return super().stats_units | {
             "Target_P": JANUS_UNITS["pressure"],
             "Target_T": JANUS_UNITS["temperature"],
         }
@@ -1408,7 +1408,7 @@ class NVT(MolecularDynamics):
         return stats
 
     @property
-    def unit_info(self) -> dict[str, str]:
+    def stats_units(self) -> dict[str, str]:
         """
         Get units of returned statistics.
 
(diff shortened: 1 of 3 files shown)


ElliottKasoar:
I'll make it into a proper test, and try to check against heating/different ensembles too, but after running:

`janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 100 --stats-every 1 --traj-every 1`

This test passes:

```
def test_temperatures():

    atoms = read("NaCl-nvt-T300.0-traj.extxyz", index=":")
    temps = [a.info["temperature"] for a in atoms]
    data = Stats("NaCl-nvt-T300.0-stats.dat")

    T_col = data.labels.index("T")
    stats_temps = data.data[:, T_col]

    assert temps == pytest.approx(stats_temps, rel=1e5)
```

(Consecutive temperatures include `285.5082095881303, 280.95845823589696`, so outside of the tolerance).

@alinelena, if you can share any specific commands that give inconsistent results, that would be great.

janus_core/calculations/md.py
```diff
@@ -532,6 +532,13 @@ def _set_info(self) -> None:
         except ValueError:
             self.dyn.atoms.info["density"] = 0.0
 
+        e_kin = self.dyn.atoms.get_kinetic_energy() / self.n_atoms
+        current_temp = e_kin / (1.5 * units.kB)
+        self.struct.info["temperature"] = current_temp
+
+        if hasattr(self.dyn, "set_temperature") or isinstance(self, NVT_CSVR | NPT_MTK):
```
alinelena, line 539: should not be also NVT in there?

ElliottKasoar: That should be included by `hasattr(self.dyn, "set_temperature")`.

These are the same checks added in #414 to make sure `_set_target_temperature` is used appropriately.

(I did consider adding a `has_thermostat` property for this condition, but in `_set_target_temperature` `NVT_CSVR` and `NPT_MTK` are handled individually, so it would only be used here)

janus_core/calculations/md.py
```diff
@@ -727,6 +734,28 @@ def _write_correlations(self) -> None:
...
+        tuple[str]
+            Keys of Atoms.info.
+        """
+        return (
+            "energy",
+            "forces",
+            "stress",
+            "time",
+            "real_time",
+            "temperature",
```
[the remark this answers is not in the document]
ElliottKasoar: That's added for the relevant ensembles, otherwise we'd be defining units for unused variables.

Having said that, I should probably remove pressure, since it's never written out in the info.

alinelena:
> I'll make it into a proper test, and try to check against heating/different ensembles too, but after running:
> 
> `janus md --ensemble nvt --struct tests/data/NaCl.cif --steps 100 --stats-every 1 --traj-every 1`
> 
> This test passes:
> 
> ```
> def test_temperatures():
> 
>     atoms = read("NaCl-nvt-T300.0-traj.extxyz", index=":")
>     temps = [a.info["temperature"] for a in atoms]
>     data = Stats("NaCl-nvt-T300.0-stats.dat")
> 
>     T_col = data.labels.index("T")
>     stats_temps = data.data[:, T_col]
> 
>     assert temps == pytest.approx(stats_temps, rel=1e5)
> ```
> 
> (Consecutive temperatures include `285.5082095881303, 280.95845823589696`, so outside of the tolerance).
> 
> @alinelena, if you can share any specific commands that give inconsistent results, that would be great.

i had in mind target_temperature in heating was shifted by one

alinelena (approved):

PR #470, opened 2025-03-13
add weas for graphs

alinelena:
this is only singlepoint tutorial but can be used for inspiration


ElliottKasoar (approved):
Just added `weas-widget` to a `visualise` extra (as well as `all`), and a note about installing it in the docs.

Someone else will need to approve too, but I think this looks good, thanks!

harveydevereux (approved):
This looks good to me as well

PR #478, opened 2025-03-21
Add singlepoint and geomopt CLI tutorials

ElliottKasoar:
Tries to introduce:

- Getting help (janus and commands)
- Running single point calculations
- Building and using config files
- Running calculations with different MLIPs
  - Arguable could come later, but helps explain what info the results actually contains, so it felt useful here
- Summary and log files
- Reusing configurations via the summary
  - Again, in some ways not a priority, but I use it as a way to show different calculations and multiple structures
- Setting kwargs via --kwargs and, preferably, a config file

### DIFF
--- a/docs/source/tutorials/cli/geomopt.ipynb
+++ b/docs/source/tutorials/cli/geomopt.ipynb
@@ -1,0 +1,583 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text",
+    "id": "view-in-github"
+   },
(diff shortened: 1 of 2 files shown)


ElliottKasoar:
Also just a note, I've used `--no-tracker` everywhere, since otherwise these don't run for me (password prompt on Mac).

I'm undecided about whether they should be kept or not, since it'd be nice to show the carbon tracking too, but if anyone else is using a Mac, it becomes annoying having to add it in everywhere.

alinelena (approved):

PR #480, opened 2025-03-24
add phonons example

alinelena:


ElliottKasoar (approved):
I'll do another pass after our live testing, when we're thinking about what to do with these longer term, but generally looks great.

Issue #490, opened 2025-03-25
Update SevenNet


ElliottKasoar:
SevenNet 0.11.0 has been released, which I think is the final block for NumPy 2, and removes a block for torch 2.6.

However, it also depends on `e3nn>=0.5.0`, which means we'd no longer be able to use it at the same time as mace.

PR #492, opened 2025-03-25
Fix CLI tutorials for Colab

### LINKED ISSUE #485: Fix Google Colab installation
Google Colab now defaults to NumPy 2, which some models are incompatible with.

We need to uninstall it, and restart the session in order to avoid errors e.g. a workaround described in https://github.com/googlecolab/colabtools/issues/5115.

ElliottKasoar:
Resolves #485

This adds a few changes needed to run successfully in Colab:

- Downgrading NumPy, which is installed by default
  - This also requires restarting the session, which is done automatically as part of these changes
- Downgrade torch
  - MACE should work with torch 2.6, but our current version of SevenNet does not, which I think it mostly because MACE forces an old version of e3nn (MACE introduced a [workaround](https://github.com/ACEsuit/mace/issues/555), whereas SevenNet has recently upgraded e3nn  - see #490)
  - We also need to remove torchaudio (and also torchvision, to be safe), as this is pre-installed, but raises an error
- Ensures the `./data` exists for geomopt.ipynb
- Disables all carbon tracking, since this seems to throw errors in Colab (may be worth revisiting, at least once sudo is [hopefully removed for MacOS](https://github.com/mlco2/codecarbon/pull/752))
- Fixes data links

Also tides up a few cells in minor ways

### DIFF
--- a/docs/source/tutorials/cli/geomopt.ipynb
+++ b/docs/source/tutorials/cli/geomopt.ipynb
@@ -38,7 +38,10 @@
    "source": [
     "# import locale\n",
     "# locale.getpreferredencoding = lambda: \"UTF-8\"\n",
-    "# !python3 -m pip install janus-core[all] data-tutorials"
+    "\n",
+    "# ! pip uninstall torch torchaudio torchvision numpy -y\n",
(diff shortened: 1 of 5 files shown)


oerc0122 (approved):
Looks good!

PR #493, opened 2025-03-25
Use click choices in CLI

### LINKED ISSUE #464: Use Choices for CLI Literal parameters
Similar to NEB's `interpolator` (https://github.com/stfc/janus-core/blob/main/janus_core/cli/neb.py#L53), it would be useful in other cases where `Literal` is used, such as `arch`, `device`, etc.

Note: There may be native typer support for `Literal` eventually: https://github.com/fastapi/typer/pull/429

ElliottKasoar:
Resolves #464

This results in something like:

```
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  --struct                          PATH                                                                     Path of structure to simulate. [default: None] [required]               │
│    --arch                            [mace|mace_mp|mace_off|m3gnet|chgnet|alignn|sevennet|nequip|dpa3|orb|ma  MLIP architecture to use for calculations. [default: mace_mp]           │
│                                      ttersim]                                                                                                                                         │
│    --device                          [cpu|cuda|mps|xpu]                                                       Device to run calculations on. [default: cpu]                           │
│    --model-path                      TEXT                                                                     Path to MLIP model. [default: None]                                     │
│    --properties                      [energy|stress|forces|hessian]                                           Properties to calculate. If not specified, 'energy', 'forces' and       │
│                                                                                                               'stress' will be returned. 
```

### DIFF
--- a/janus_core/cli/eos.py
+++ b/janus_core/cli/eos.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Annotated, get_args
 
+from click import Choice
 from typer import Context, Option, Typer
 from typer_config import use_config
 
@@ -21,6 +22,7 @@ from janus_core.cli.types import (
     WriteKwargs,
 )
 from janus_core.cli.utils import yaml_converter_callback
+from janus_core.helpers.janus_types import EoSNames
 
 app = Typer()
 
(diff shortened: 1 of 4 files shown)


oerc0122 (approved):
Looks really good and helpful.

PR #496, opened 2025-03-26
Add Windows CI

### LINKED ISSUE #489: Test and document Windows installation
I think we will always recommend WSL2, but having some understanding of Windows installation requirements, and testing on Windows, would help with debugging some people's issues.

ElliottKasoar:
Resolves #489

This generally seems to set up and run ok, but there are a few errors relating to paths, which need fixing.

The docs should be updated to at least reference that tests run (and ideally pass) on Windows, but perhaps more detailed installation guidelines should be separate.


ElliottKasoar:
Most of the issues related to how we specified a few paths in our testing, but there were a few actual bugs, which are hopefully "fixed":

- [Orb](https://github.com/orbital-materials/orb-models/issues/22) and [MatterSim](https://github.com/microsoft/mattersim/issues/102) both currently seem incompatible with Windows
  - Given that the issues were both closed, there doesn't appear to be any intention to fix this, so I've aded a system tag so they're skipped for Windows.

- Windows paths saved with `\` break the log, when we try to read it back in as a yaml, so I've added an extra layer of log parsing to convert these to `/`, which generally still work with Windows anyway.
  - Currently, I do this to the `record.msg`, which fixes the bug for `preprocess`, but also the `record.args`, which fixes the bug for `train`, since we pass a placeholder `%s`, which isn't formatted otherwise. The alternative would be to format the path within `train`, which would preserve in more cases the lazy string conversion, but risks cases I haven't encountered producing the same bug. Any thoughts/preference?
  
- uv setup currently has a bug that means cleaning up the cache [fails](https://github.com/astral-sh/setup-uv/issues/226), but we can skip this for now.

alinelena (approved):

alinelena (approved):

harveydevereux (approved):
Looks good to me, assuming the test still pass

PR #501, opened 2025-04-02
Build docs only on release

### LINKED ISSUE #440: Tutorial notebooks overwritten
Currently, the notebooks are run on tagged releases, with the contents visible in the documentation (I believe).

However, on subsequent pushes to main, which also build the docs, but don't  execute the notebooks, the blank notebooks are instead copied over, which remotes the previously calculated outputs.

We could add something like `exclude_patterns`, but then I'm not sure if they would be included at all, and it raises various warnings.

Alternatively, we could include the notebooks in their run state, which are not overwritten by default, and can be forced to re-execute, but this would require uploading larger notebook files to the docs.

Other suggestions welcome.

ElliottKasoar:
Resolves #440

Deletes documentation workflow, so docs are only built and published on release. (Docs are still always tested.) 

As discussed with @alinelena, I'm still somewhat considering a `dev` branch, but the main initial motivation was a simple solution for the notebook tutorials being overwritten, which this also solves, with exactly the same documentation either way.

At various points we've also mentioned any push to main generating a new release, which would also lead to the same outcome.

For now, this does mean the docs may lag behind slightly, but I'd argue if anything needs updating substantially in the docs, that's usually worth a new release. The only exception would be if we have PRs which update docs specifically, which would feel strange to create a release for, but it's probably still fine.


oerc0122 (changes requested):
Just to double check the docs are built in the release scripts? 

I do think that we probably do want to test building the docs and also running any `doctest` and `nbmake` tests as part of the CI, I want to ensure that those are being done. 

We can make sure that the docs tests are only run on changes in a PR to `doc` or contains `doc` somewhere in the labels.

oerc0122 (approved):

Issue #442, opened 2025-02-21
using labels to access the stats object leads to bad matching


alinelena:
at the moment we can access data both by index or by label after we read the stats, label matching seems to be fishy even if @oerc0122 claims is fuzzy.

harveydevereux:
Whats the best way to do this and keep the fuzzy match?

at the moment ```stats["T"]``` will return ```Step``` as ```"T".lower() in "Step"```

and ```stats["Time"]``` will return ```"Real_time"``` because "Real_time" is before "Time". 

Could we check for exact matches first or something?

ElliottKasoar:
Do we actually want fuzzy matching at all? I think anything more than comparisons after `str.lower()` is probably going to lead to messy results.

harveydevereux:
> Do we actually want fuzzy matching at all? I think anything more than comparisons after `str.lower()` is probably going to lead to messy results.

I think you may be right, any chance I/we can quickly do that today?

oerc0122:
The reason I wanted fuzzy matching was because some of the columns had complicated names. But if it's causing problems, I'm happy for it to go.

PR #505, opened 2025-04-07
Check exact and fuzzy matches in Stats

harveydevereux:
This is perhaps one way to address #442

I think there are 3 points:

- 'Exact' (case insensitive) matches should always work e.g. ```t``` or ```T``` should match ```t``` (temperature) rather that ```Step```. Currently I don't think you can get temperature from a string since ```'t'``` always goes to step and the units are stripped so ```'T [K]'``` throws ```IndexError: T K not found in labels```.
- Fuzzy matching should be preserved, it is useful to keep 'vol' returning Volume I think.
- But what to do about multiple fuzzy matches?  Currently I return the shortest matching key (so ```'TI'``` matches ```'Time'``` not ```'Real Time'```.

Maybe there are better ways to handle multiple fuzzy matches?

### DIFF
--- a/janus_core/helpers/stats.py
+++ b/janus_core/helpers/stats.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterator
+from difflib import get_close_matches
 from functools import singledispatchmethod
 import re
 from typing import TypeVar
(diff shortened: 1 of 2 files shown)


oerc0122:
An alternative might involve [`difflib`](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches)

ElliottKasoar:
> An alternative might involve [`difflib`](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches)

That's quite cool, but we may need to play with the settings if we want to cover everything @harveydevereux's changes do e.g. 

```python
>>> get_close_matches('vol', ['ape', 'apple', 'peach', 'puppy', 'app', 'Volume'])
[]
```

(It finds it if the cases match, so we could still do the manual `lower()`, or we could lower the `cutoff`.)

@harveydevereux , do we definitely want the shortest label? The sorting by score from `get_close_matches ` is quite nice, although admittedly harder to explain to users.

ElliottKasoar:
> difflib does look like it will be better I'll take a look. Without knowing about it, shortest seemed at least somewhat reasonable, no reason otherwise.
> 
> On the score, even if is is more complex at least it's standard python we can point to. This is a best effort match at the end of the day.
> 
> The real question is if there are multiple decent matches should that throw a warning?

Yes, I think a warning seems pretty sensible.

janus_core/helpers/stats.py
```diff
@@ -71,18 +72,11 @@ def _getind(self, lab: T) -> T:
...
-            ),
-            None,
-        )
-        if index is None:
+        labels = [label.lower() for label in self.labels]
+        match = get_close_matches(lab.lower(), labels, n=1)
+        if len(match) == 0:
             raise IndexError(f"{lab} not found in labels")
-        return index
+        return labels.index(match[0])
```
ElliottKasoar, line 79: Should we maybe warn if there are multiple matches?

ElliottKasoar (commented):
Generally looks good, thanks @harveydevereux!

This will also need rebasing after #541.

janus_core/helpers/stats.py
oerc0122: Because of `n=1` we aren't getting those and some of the "closest" may be very far away.

harveydevereux: It will now, and warn if there are multiple. E.g.

```shell
from janus_core.helpers.stats import Stats; s = Stats("tests/data/md-stats.dat"); s['px']
UserWarning: multiple matches found for label px: ['pxz', 'pxy', 'pxx']
array([ 1.177e+03, -1.296e+03, -1.787e+03, -1.663e+03, -2.345e+03, ...
```

and

```shell
from janus_core.helpers.stats import Stats; s = Stats("tests/data/md-stats.dat"); s['t']
array([285.958, 280.568, 258.041, 291.922, 300.084, 277.877, 303.594, ...
```

PR #508, opened 2025-04-11
Update Orb

### LINKED ISSUE #507: Update Orb
Orb v3 models are out, and should also allow dropping `pynanoflann`, and removing another torch 2.6 blocker.

ElliottKasoar:
Resolves #507

Also applies a similar fix to MACE's for torch 2.6, by setting the `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD` environment variable, although I only set it if the user has not done so.

Otherwise we'd need to make Orb conflict with dependencies which do not have a fix yet (e.g. nequip, our current version of sevennet (#490)).

### DIFF
--- a/tests/test_single_point.py
+++ b/tests/test_single_point.py
@@ -80,7 +80,7 @@ def test_potential_energy(struct, expected, properties, prop_key, calc_kwargs, i
             "toluene.xyz",
             {"model_path": NEQUIP_PATH},
         ),
-        ("orb", "cpu", -27.08180809020996, "NaCl.cif", {}),
+        ("orb", "cpu", -27.08186149597168, "NaCl.cif", {}),
         ("orb", "cpu", -27.089094161987305, "NaCl.cif", {"model_path": "orb-v2"}),
         (
             "sevennet",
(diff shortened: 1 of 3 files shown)


tests/test_single_point.py
```diff
@@ -80,8 +80,8 @@ def test_potential_energy(struct, expected, properties, prop_key, calc_kwargs, i
             "toluene.xyz",
             {"model_path": NEQUIP_PATH},
         ),
-        ("orb", "cpu", -27.088973999023438, "NaCl.cif", {}),
-        ("orb", "cpu", -27.088973999023438, "NaCl.cif", {"model_path": "orb-v2"}),
+        ("orb", "cpu", -27.08186149597168, "NaCl.cif", {}),
+        ("orb", "cpu", -27.089094161987305, "NaCl.cif", {"model_path": "orb-v2"}),
```
oerc0122, line 84: Do we need them to this precision if comparison is rel=1e-5?, just do 6d.p.

ElliottKasoar: As a test which needs to be consistent across different systems, I don't think we can guarantee better precision than somewhere around rel=1e-5, but occasionally it is useful to compare to a higher level of precision for a specific test, so I don't think it hurts to have essentially a note of what I think it should be?

oerc0122 (approved):

PR #516, opened 2025-04-14
Fix catching BadZipFile

ElliottKasoar:
`BadZipFile` doesn't have a `.msg`, which this line was supposed to fix, but I implemented the fix in the wrong order.

### DIFF
--- a/tests/test_mlip_calculators.py
+++ b/tests/test_mlip_calculators.py
@@ -109,7 +109,7 @@ def test_mlips(arch, device, kwargs):
         assert calculator.parameters["version"] is not None
         assert calculator.parameters["model_path"] is not None
     except (BadZipFile, URLError) as e:
-        if "Connection timed out" in e.msg or isinstance(e, BadZipFile):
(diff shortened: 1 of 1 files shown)


oerc0122 (approved):

tests/test_mlip_calculators.py
```diff
@@ -109,7 +109,7 @@ def test_mlips(arch, device, kwargs):
         assert calculator.parameters["version"] is not None
         assert calculator.parameters["model_path"] is not None
     except (BadZipFile, URLError) as e:
-        if "Connection timed out" in e.msg or isinstance(e, BadZipFile):
+        if isinstance(e, BadZipFile) or "Connection timed out" in e.msg:
             pytest.skip("Model download failed")
         else:
             raise e
```
oerc0122, line 115:
```python
    except BadZipFile:
        pytest.skip("Model download failed")
    except URLError as err:
        if "Connection timed out" in e.msg:
            pytest.skip("Model download failed")
        raise
            
        else:
            raise e
```

PR #521, opened 2025-04-16
Add MD progress bar

### LINKED ISSUE #405: Use progress bar for MD

ElliottKasoar:
Resolves #405

See examples for the next week: https://asciinema.org/a/gQeyJo9JJPvSWFkuLVCkdnfHV

Or permanent example:
image

This combines aspects of #444 and #457.

Thanks to @k-harris27, we do now have the option to use ASE's `irun`, and wrap this with the `rich` progress tracker, but unfortunately this seems to provide limited options for reducing the frequency of updates, since each `next(irun)` causes the bar to refresh. This ends up having a noticeable (consistently 4-5% from tests up to around 25 min on my laptop) impact on performance, so I think we need to limit the frequency of updates, which requires a more manual process.

To simplify things, I've gone with a single progress bar, but the description changes according to the current step.

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -15,7 +15,7 @@ from warnings import warn
 
 # Only needed for type hints
 if TYPE_CHECKING:
-    from rich.progress import Progress, TaskID
+    from rich.progress import TaskID
 
 from ase import Atoms
 from ase.geometry.analysis import Analysis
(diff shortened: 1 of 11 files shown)


ElliottKasoar:
(Will need rebasing after #520 is merged)

janus_core/calculations/md.py
```diff
@@ -555,6 +574,135 @@ def output_files(self) -> None:
...
+    def heating_steps_per_temp(self) -> int:
+        """
+        Get number of steps to be run at each temperature in heating/cooling ramp.
+
+        Returns
+        -------
+        int
+            Number of steps to be run at each temperature in heating/cooling ramp.
+        """
+        return int(self.temp_time // self.timestep)
```
oerc0122, line 628: Shouldn't need to cast after floordiv (`//`)

janus_core/cli/phonons.py
```diff
@@ -182,6 +183,7 @@ def phonons(
     log: LogPath = None,
     tracker: Tracker = True,
     summary: Summary = None,
```
oerc0122, line 185: `Summary | None` ?

janus_core/cli/singlepoint.py
```diff
@@ -69,6 +70,7 @@ def singlepoint(
     log: LogPath = None,
     tracker: Tracker = True,
     summary: Summary = None,
```
oerc0122, line 72: `Summary | None`?

oerc0122 (approved):

janus_core/cli/phonons.py
ElliottKasoar:
```
Summary = Annotated[
    Path | None,
    Option(
        help=(
            "Path to save summary of inputs, start/end time, and carbon emissions. "
            "Default is inferred from `file_prefix`."
        ),
        rich_help_panel="Logging/summary",
    ),
]
```
So I think this is fine?

janus_core/calculations/md.py
```diff
@@ -555,6 +574,135 @@ def output_files(self) -> None:
...
+    def ramp_temps(self) -> list[float]:
+        """
+        Get list of temperatures in heating/cooling ramp.
+
+        Returns
+        -------
+        list[float]
+            Temperatures in heating/cooling ramp.
+        """
+        ramp_sign = 1 if (self.temp_end - self.temp_start) > 0 else -1
```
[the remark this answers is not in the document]
ElliottKasoar: That's quite neat but I think the current form makes the reason behind it slightly more explicit.

janus_core/cli/singlepoint.py
ElliottKasoar: Same as above

janus_core/calculations/md.py
ElliottKasoar: I can get a float (admittedly int-like) out:

```
>>> x = 1.15
>>> x // 1
1.0
```

Issue #518, opened 2025-04-15
Tutorial notebooks keep crashing


catastropiyush:
I was trying to run the colab notebooks available but when I run the installation commands the session crashes and gives the following output.
I am using the T4 GPU.

```
error: a value is required for '--prerelease <PRERELEASE>' but none was supplied
  [possible values: disallow, allow, if-necessary, explicit, if-necessary-or-explicit]

For more information, try '--help'.
{'status': 'ok', 'restart': True}
```

Thank you for your time,
Piyush

ElliottKasoar:
Hi, thanks for raising this!

This appears to be a new issue with the environment variables Google Colab sets: https://github.com/googlecolab/colabtools/issues/5237

Explicitly setting the `--prerelease` value is probably the simplest fix, which would look something like:


```
! uv pip install janus-core[all] data-tutorials torch==2.5.1 --system --prerelease  if-necessary-or-explicit
```

(or equivalent, depending on the tutorial you are in).

Other alternatives fixe include running `! unset UV_PRERELEASE`, or `! UV_PRERELEASE=if-necessary-or-explicit; uv pip install ...`, as discussed in the issue.

alinelena:
@catastropiyush did it help?

catastropiyush:
Yes its running now.
I used 
```
import locale
locale.getpreferredencoding = lambda: "UTF-8"

! pip uninstall torch torchaudio torchvision numpy -y
! uv pip install janus-core[all] data-tutorials torch==2.5.1 --system
get_ipython().kernel.do_shutdown(restart=True)
```

While running the first tutorial it also needs one more install.

`! pip install --upgrade huggingface-hub`

ElliottKasoar:
Yep, the Colab issue with `uv` seems to have been fixed on their end, but there is some new issue with `huggingface-hub`.

It's not strictly needed in the tutorial, but it's a dependency or orb via `cached-path`. `torchmetrics` is throwing the error, I think.

ElliottKasoar:
Resolved by #531

PR #523, opened 2025-04-23
Colab notebooks running

catastropiyush:
Not a major change. Tested out all the colab notebooks.
Just one single line is needed.
`!pip install --upgrade huggingface-hub`

Thank you

### DIFF
--- a/docs/source/tutorials/python/eos.ipynb
+++ b/docs/source/tutorials/python/eos.ipynb
@@ -1,347 +1,398 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "colab_type": "text",
-    "id": "view-in-github"
-   },
-   "source": [
-    "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/stfc/janus-core/blob/main/docs/source/tutorials/python/eos.ipynb)"
-   ]
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/catastropiyush/janus-core/blob/main/docs/source/tutorials/python/eos.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "4vcbRxmIhHLL"
+      },
+      "source": [
+        "# Equation of State"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "djnGmdzabfer"
+      },
+      "source": [
+        "## Set up environment (optional)\n",
+        "\n",
+        "These steps are required for Google Colab, but may work on other systems too:"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "TF-EiWxyuMc7",
+        "collapsed": true
+      },
+      "outputs": [],
+      "source": [
+        "import locale\n",
+        "locale.getpreferredencoding = lambda: \"UTF-8\"\n",
+        "\n",
+        "! pip uninstall torch torchaudio torchvision numpy -y\n",
+        "! uv pip install janus-core[all] data-tutorials torch==2.5.1 --system\n",
+        "get_ipython().kernel.do_shutdown(restart=True)"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "LTVcudWQbfet"
+      },
+      "outputs": [],
+      "source": [
+        "from ase import Atoms\n",
+        "from ase.io import read\n",
+        "from weas_widget import WeasWidget\n",
+        "\n",
+        "from data_tutorials.data import get_data\n",
+        "\n",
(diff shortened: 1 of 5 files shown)


alinelena:
since janus itself does not use huggingface, can you indicate which tutorial does not run wihout an explicit dependence? in my opinion huggingface is probably a dependency of the potential you use and probably fixed there.

catastropiyush:
> since janus itself does not use huggingface, can you indicate which tutorial does not run wihout an explicit dependence? in my opinion huggingface is probably a dependency of the potential you use and probably fixed there.

I tried running the tutorials in colab as it is and when I am starting any calculation this error comes up.
For example the first tutorial for single point energy calculation.
image


```
/usr/local/lib/python3.11/dist-packages/e3nn/o3/_wigner.py:10: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
[<ipython-input-4-fdccc97ea2b7>](https://localhost:8080/#) in <cell line: 0>()
----> 1 sp_mace = SinglePoint(
      2     struct=NaCl.copy(),
      3     arch="mace_mp",
      4     device="cpu",
      5     model_path="small",

20 frames
[/usr/local/lib/python3.11/dist-packages/transformers/utils/versions.py](https://localhost:8080/#) in _compare_versions(op, got_ver, want_ver, requirement, pkg, hint)
     42         )
     43     if not ops[op](version.parse(got_ver), version.parse(want_ver)):
---> 44         raise ImportError(
     45             f"{requirement} is required for a normal functioning of this module, but found {pkg}=={got_ver}.{hint}"
     46         )

ImportError: huggingface-hub>=0.30.0,<1.0 is required for a normal functioning of this module, but found huggingface-hub==0.27.1.
Try: `pip install transformers -U` or `pip install -e '.[dev]'` if you're working with git main
```

ElliottKasoar:
Thanks for sharing the new error, @catastropiyush!

I can reproduce it, but I would prefer to understand and address the root cause of this.

`huggingface-hub=0.3` is not actually consistent with our current set of dependencies (`cached-path`, which is a dependency of `orb`, specifies [>=0.8.1, <0.28.0](https://github.com/allenai/cached_path/blob/main/pyproject.toml#L27)), so it's most likely a change to the pre-installed modules from Colab that we should address.

catastropiyush:
@ElliottKasoar I also think it was weird that hugging face hub came in there. I shared this so that we can address the issue,hope it helps.

ElliottKasoar:
Ok, following on from my comment in #518, the issue is a little convoluted:

- `mace` depends on `torchmetrics`
- `torchmetrics` has a [check](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/utilities/imports.py#L39C53-L39C73) based on whether `transformers>=4.4.0`
- Since Colab has `transformers==4.51.3` pre-installed by default, `torchmetric` [attempts to import](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/functional/__init__.py#L144) from `transformers`
- `transformers` is [incompatible](https://github.com/huggingface/transformers/blob/main/setup.py#L120) with `huggingface-hub<0.30.0`, and checks against this, so the version installed for `orb` clashes

I think the simplest solution is to uninstall `transformers` at the same time as `torch`, `numpy` etc., which I think solves this.

@catastropiyush, can you confirm if:

```
! pip uninstall torch torchaudio torchvision transformers numpy -y
```
resolves this for you?

If so, I'm happy to do a new PR with this change (typically I would use `git add -p` to ensure only the relevant line is changed), or please do update this, if you think it makes sense.

catastropiyush:
@ElliottKasoar thank you for detailing out the issue with the libraries. I tried to run again the tutorials.

```
import locale
locale.getpreferredencoding = lambda: "UTF-8"

! pip uninstall torch torchaudio torchvision numpy -y
! uv pip install janus-core[all] data-tutorials torch==2.5.1 --system
get_ipython().kernel.do_shutdown(restart=True)
```

But still I get the previous error :
```
ImportError: huggingface-hub>=0.30.0,<1.0 is required for a normal functioning of this module, but found huggingface-hub==0.27.1.
Try: `pip install transformers -U` or `pip install -e '.[dev]'` if you're working with git main
```

catastropiyush:
@ElliottKasoar Thanks for providing the detailed issue with the libraries and for finding the source of error.
Its working now. I'll update the notebooks and then submit a PR. I think since its a minor change and you figured it out so maybe its not needed.

Thank you

ElliottKasoar:
Hi @catastropiyush, sorry, just to clarify, are you updating this PR to update the uninstall command?

If so, please try to change the minimum number of lines in each notebook to make the file changes more clear - ideally each notebook should only need one line changed.

If not, are you happy for me to close this and I will create a new PR with the changes we discussed?

catastropiyush:
@ElliottKasoar I have updated the uninstall command in the colab notebooks and submitted a [pull request](https://github.com/stfc/janus-core/pull/531).
Thank you

PR #526, opened 2025-04-23
Added plumed for metadynamics simulations

carlosbornes:
Hi @alinelena @ElliottKasoar I wanted to have an interface to plumed within janus. I did some changes and followed [this tutorial](https://www.plumed-tutorials.org/lessons/24/012/data/04_metaD.html) for testing the implementation. I did a fast simulation, around 6k points and then plotted the free energy profile as a function of ϕ and θ similar to what they did in the tutorial. I'm attaching the test I did.

If you think this is relevant I can push it to you repo. But yeah feel free to right this in maybe a more polish and consistent way.

[janus_results.zip](https://github.com/user-attachments/files/19869005/janus_results.zip)


janus_core/calculations/md.py
```diff
@@ -222,6 +228,8 @@ def __init__(
         write_kwargs: OutputKwargs | None = None,
         post_process_kwargs: PostProcessKwargs | None = None,
         correlation_kwargs: list[CorrelationKwargs] | None = None,
+        plumed_input: str | PathLike | None = None,
```
ElliottKasoar, line 231:
```suggestion
        plumed_input: PathLike | None = None,
```
`PathLike` is just a alias for `str | Path`

janus_core/calculations/md.py
```diff
@@ -463,6 +478,48 @@ def __init__(
...
 
+        # Wrap calculator with Plumed if input is provided
+        if self.plumed_input:
+            try:
+                from ase.calculators.plumed import Plumed
+            except ImportError:
+                raise ImportError(
+                    "ASE Plumed calculator not found. Install janus-core with "
+                    "'plumed' extra: pip install janus-core[plumed]"
+                )
```
ElliottKasoar, line 489: I'm not sure this ever actually generates an `ImportError`, so we may need to check `plumed` explicitly:

```
>>> from ase.calculators.plumed import Plumed
>>>
>>> from plumed import Plumed as pl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'plumed'
```

I probably wouldn't specify exactly how to install plumed either.

janus_core/calculations/md.py
```diff
@@ -463,6 +478,48 @@ def __init__(
...
+                plumed_input_source = "string"
+
+            # Calculate kT in ASE units (eV)
+            kT = self.temp * units.kB
+
+            self.plumed_log = self._build_filename(
+                "plumed.log", self.param_prefix, filename=self.plumed_log
+            )
+
+            build_file_dir(self.plumed_log)
```
ElliottKasoar, line 508: This should ideally only happen at the point the file is about to be written, rather than in the initialisation.

janus_core/calculations/md.py
```diff
@@ -463,6 +478,48 @@ def __init__(
...
+                    plumed_script = f.read()
+                plumed_input_source = str(plumed_input_path.resolve())
+            else:
+                plumed_script = self.plumed_input
+                plumed_input_source = "string"
+
+            # Calculate kT in ASE units (eV)
+            kT = self.temp * units.kB
+
+            self.plumed_log = self._build_filename(
```
ElliottKasoar, line 504: Can this be added to the `output_files` property as well?

janus_core/calculations/md.py
```diff
@@ -463,6 +478,48 @@ def __init__(
...
+                input=plumed_script.splitlines(),
+                timestep=self.timestep,
+                atoms=self.struct,
+                kT=kT,
+                log=str(self.plumed_log),
+            )
+            if self.logger:
+                self.logger.info("Wrapped calculator with ASE Plumed interface.")
+                self.logger.info("PLUMED input source: %s", plumed_input_source)
+                self.logger.info("PLUMED log file: %s", self.plumed_log)
```
ElliottKasoar, line 521: If this is added to `output_files`, we probably don't need this too.

janus_core/cli/md.py
```diff
@@ -133,6 +133,20 @@ def md(
     ] = 10,
     post_process_kwargs: PostProcessKwargs = None,
     correlation_kwargs: CorrelationKwargs = None,
+    plumed_input: Annotated[
+        str | None,
+        Option(
+            help="PLUMED input script content or path to PLUMED input file.",
+            rich_help_panel="Calculation",
+        ),
+    ] = None,
```
ElliottKasoar, line 142: Will this ever realistically be expressible as a plain string? Maybe should just be a `Path` here?

janus_core/calculations/md.py
```diff
@@ -463,6 +478,48 @@ def __init__(
...
+
+            build_file_dir(self.plumed_log)
+
+            self.struct.calc = Plumed(
+                calc=self.struct.calc,
+                input=plumed_script.splitlines(),
+                timestep=self.timestep,
+                atoms=self.struct,
+                kT=kT,
+                log=str(self.plumed_log),
```
ElliottKasoar, line 516: We probably need `restart` to be passed here too?

ElliottKasoar (changes requested):
Thanks for this, @carlosbornes, it's a great start!

I've left a few comments, but they're mostly polishing, so I wouldn't worry too much about them now. The biggest things we'll need first are:

- Tests, ideally covering all ensembles, and at least a couple of different structures
  - E.g. from our testing, `get_stats` can throw errors during MD, since the Plumed calculator does not implement stress, but you may not have observed this if you used structures without a defined cell
  - This will require getting it installed, ideally on all platforms, for the CI to run (but perhaps we should make them skippable too)
- Documentation, including (limited) installation instructions

janus_core/calculations/md.py
carlosbornes: How would you handle this? I did in such a way because I did not want to create some problem with ASE's plumed calculator. But yeah probably there is a smarter way to do it

ElliottKasoar: It depends when this log is written. The main reasoning is just that initialing the `MolecularDynamaics` class shouldn't usually change/create anything - this happens when `run()` is called, so probably just moving `build_file_dir` into `run()` is fine.

(Defining the filename should stay in the init, as this means we can inspect the expected output files before running the calculation)

carlosbornes:
I've tested the running NVT and NPT with ASE's python wrapper on a periodic system and yes it fails due to stresses. Maybe with from scratch Plumed compilation will work

alinelena:
> I've tested the running NVT and NPT with ASE's python wrapper on a periodic system and yes it fails due to stresses. Maybe with from scratch Plumed compilation will work
odd for npt i understand but not for nvt...

ElliottKasoar:
> I've tested the running NVT and NPT with ASE's python wrapper on a periodic system and yes it fails due to stresses. Maybe with from scratch Plumed compilation will work

Assuming it's the same issue we encountered, it's not entirely a Plumed issue. When we do a `try...except`, we only catch `ValueError`s: https://github.com/stfc/janus-core/blob/main/janus_core/calculations/md.py#L902

I think we just need to change that to `except (ValueError, NotImplementedError)`.

carlosbornes:
@ElliottKasoar is right, it is not due to Plumed itself, is more because janus tries to write stress for logging statistics with the implementation in md.py. Also, changing the line https://github.com/carlosbornes/janus-core/blob/ebab97ce2d7627c11d7b07ef4ea6024e44ed3343/janus_core/calculations/md.py#L958


from `except ValueError:` to `except (ValueError, NotImplementedError):` makes the NVT run well. I've pushed that change

 I'm attaching a short umbrella sampling nvt simulation from a water molecule jumping from one site to the other. It seems to work.
 
 [h2ojump.zip](https://github.com/user-attachments/files/19911532/h2ojump.zip)

janus_core/calculations/md.py
```diff
@@ -159,6 +159,12 @@ class MolecularDynamics(BaseCalculation):
         Keyword arguments to control post-processing operations. Default is None.
     correlation_kwargs
         Keyword arguments to control on-the-fly correlations. Default is None.
+    plumed_input
+        PLUMED input or path to a PLUMED input file. If provided,
+        the ASE Plumed calculator will be used to wrap the MLIP calculator.
+        Requires the 'plumed' extra to be installed. Default is None.
```
ElliottKasoar, line 165: I think this should just be a path. (It can probably stay as `PathLike`, but `Path(plumed_input)` should exist).

ElliottKasoar:
> from `except ValueError:` to `except (ValueError, NotImplementedError):` makes the NVT run well. I've pushed that change

Great to hear that it's working! Are you able to add some tests and set up so it runs in the CI?

janus_core/calculations/md.py
```diff
@@ -463,6 +478,46 @@ def __init__(
...
+                )
+
+            # Read input script from file if it's a path
+            plumed_input_path = Path(self.plumed_input)
+            if not plumed_input_path.is_file():
+                raise FileNotFoundError(
+                    f"PLUMED input file not found or is not a file: {plumed_input_path}"
+                )
+            with open(plumed_input_path, "r", encoding="utf-8") as f:
+                plumed_script = f.read()
```
oerc0122, line 499:
```suggestion
            plumed_script = plumed_input_path.read_text(encoding="utf-8")
```

janus_core/calculations/md.py
```diff
@@ -899,7 +955,7 @@ def get_stats(self) -> dict[str, float]:
                 -self.dyn.atoms.get_stress(include_ideal_gas=True, voigt=True)
                 / units.GPa
             )
-        except ValueError:
+        except (ValueError, NotImplementedError):
```
oerc0122, line 958: Bit worrying to silently except on `NotImplementedError`. Why?

oerc0122 (commented):
This will need a test or two before it can be merged.

janus_core/calculations/md.py
oerc0122: I would probably want a comment describing what you've described there.

carlosbornes: I've added a small comment

carlosbornes:
> Are you able to add some tests and set up so it runs in the CI?

I've added a very simple test. Regarding making sure that it runs in CI. Since this requires plumed installation I'm not sure how do you want to make it. I thought something like

```
 - name: Install PLUMED
    run: |
        sudo apt-get update
        sudo apt-get install -y plumed libplumed-dev
        uv pip install plumed
          
- name: Run PLUMED tests
   env:
       JANUS_TEST_PLUMED: 1
       PYTEST_ADDOPTS: "--durations=0"
     run: |
        uv run pytest --cov janus_core --cov-append tests/test_md_plumed.py || true
```

But you might want to force plumed installation with janus? I guess that's a bit more up to you. I gave you permissions to branch (also to @oerc0122) in case you want to easily change/add something

ElliottKasoar:
> > Are you able to add some tests and set up so it runs in the CI?
> 
> @ElliottKasoar I've added a very simple test. Regarding making sure that it runs in CI. Since this requires plumed installation I'm not sure how do you want to make it. I thought something like
> 
> ```
>  - name: Install PLUMED
>     run: |
>         sudo apt-get update
>         sudo apt-get install -y plumed libplumed-dev
>         uv pip install plumed
>           
> - name: Run PLUMED tests
>    env:
>        JANUS_TEST_PLUMED: 1
>        PYTEST_ADDOPTS: "--durations=0"
>      run: |
>         uv run pytest --cov janus_core --cov-append tests/test_md_plumed.py || true
> ```
> 
> But you might want to force plumed installation with janus? I guess that's a bit more up to you. I gave you permissions to branch (also to @oerc0122) in case you want to easily change/add something

Thanks, @carlosbornes. Sorry for the delay, I had some trouble getting it working on my Mac.

I've rebased and tidied up a few things. At the moment, I use a GitHub Action to build plumed: https://github.com/plumed/install-plumed. Does `sudo apt-get install -y plumed libplumed-dev` actually work? I can't test it particularly easily, but it doesn't seem to be listed in the docs as an installation method.

I also want to add a slightly add another at least one more test, but generally it seems to work as expected!

carlosbornes:
I've been playing around with this and you are right, sudo installing Plumed can work but it's a bit fussy. Do you have a working Github action to replace this? Please feel free to replace it, it is more clean

ElliottKasoar:
> @ElliottKasoar sorry I did not noticed you changed the Linux CI 🤦. I was trying to set it up according to Plumed official docs, but yes it is indeed very fiddly.

No worries, thanks for looking into it @carlosbornes!

I've (hopefully) fixed the restart option, added some docs, and added a few different tests.

Could you have a quick look over the changes I've made, and check it all seems reasonable? It would also be amazing if you could rerun your example and check I haven't broken anything! (We felt that the input should really be a file, so that will be at least one minor change).

I wouldn't necessarily add it as part of this PR, but it would also be great to add that notebook to our tutorials, if you're interested in having a go at that?

tests/test_md_plumed.py
```diff
@@ -0,0 +1,291 @@
...
+        # Raises runtime error if PLUMED_KERNEL environment variable not set
+        Plumed()
+    except ImportError:
+        pytest.skip("PLUMED not installed")
+    except RuntimeError as err:
+        if any("PLUMED not available" in arg for arg in err.args):
+            pytest.skip("PLUMED not configured")
+
+
+def test_nvt_plumed(tmp_path, plumed_input_file):
```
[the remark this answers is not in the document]
ElliottKasoar: That uses the calculated stress, so I don't think we can (since the Plumed calculator doesn't implement it).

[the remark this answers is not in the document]
alinelena: that is what I had in mind... a test to capture that.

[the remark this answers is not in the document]
alinelena: i mean that is not implemented.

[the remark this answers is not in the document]
ElliottKasoar: Ah I see, it will raise a `PropertyNotImplementedError`, but only at the point ASE tries to call `get_stress`, so I'll add a more useful (and earlier) catch, and then test it.

janus_core/calculations/md.py
```diff
@@ -649,6 +668,53 @@ def total_steps(self) -> int:
...
+            except ImportError as err:
+                raise ImportError(
+                    "PLUMED Python package not found. Please install PLUMED with its "
+                    "Python interface to use this feature."
+                ) from err
+            except RuntimeError as err:
+                if any("PLUMED not available" in arg for arg in err.args):
+                    raise RuntimeError(
+                        "Please ensure PLUMED_KERNEL environment variable is set"
+                    ) from err
```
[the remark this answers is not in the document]
ElliottKasoar: Ah yes, good point, thanks!

pyproject.toml
```diff
@@ -30,6 +30,7 @@ dependencies = [
     "codecarbon<3.0.0,>=2.8.4",
     "numpy<2.0.0,>=1.26.4",
     "phonopy<3.0.0,>=2.23.1",
+    "plumed<3.0.0,>=2.9.0; sys_platform != 'win32'",
```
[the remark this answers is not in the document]
ElliottKasoar: I was partly thinking that it's part of our "core" set of features, and the extras are almost all (but not exclusively, which I admittedly forgot) different calculators.

We can discuss it later, but maybe optional is slightly more sensible actually,

tests/test_md_plumed.py
```diff
@@ -0,0 +1,291 @@
...
+from janus_core.cli.janus import app
+from tests.utils import assert_log_contains
+
+DATA_PATH = Path(__file__).parent / "data"
+MODEL_PATH = Path(__file__).parent / "models" / "mace_mp_small.model"
+
+runner = CliRunner()
+
+
+@contextlib.contextmanager
```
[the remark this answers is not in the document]
ElliottKasoar: We still support 3.10, so I don't think we can use this yet?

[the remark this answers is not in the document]
oerc0122: I meant in terms of naming so the switch is trivial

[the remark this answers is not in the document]
ElliottKasoar: Ah I see, thanks, done.

oerc0122 (commented):
With the exception of whether this is to be an optional dependency I think this looks good.

ElliottKasoar (approved):
Someone else will need to approve too, and hopefully @carlosbornes has a chance to confirm he's happy, but I think everything we've discussed has been addressed.

carlosbornes:
> Someone else will need to approve too, and hopefully @carlosbornes has a chance to confirm he's happy, but I think everything we've discussed has been addressed.

Well it is way better then the rough version I sent, thanks a lot. I've tested on both the examples I've attached above and it works. One could use the two example I've attached for the tutorials but I guess I could find a smaller more simple zeolite to show the H2O jump and write some test that explains it a bit. I can try to draft something

ElliottKasoar:
> > Someone else will need to approve too, and hopefully @carlosbornes has a chance to confirm he's happy, but I think everything we've discussed has been addressed.
> 
> Well it is way better then the rough version I sent, thanks a lot. I've tested on both the examples I've attached above and it works. One could use the two example I've attached for the tutorials but I guess I could find a smaller more simple zeolite to show the H2O jump and write some test that explains it a bit. I can try to draft something

Amazing, thank you for all the work you've put into this! I've created a new issue to track this: #556

Issue #486, opened 2025-03-25
Consistency between CLI and Python interfaces.


oerc0122:
Currently, there are some cases (e.g. `temp_max`, `temp_min` vs. `temp_start`, `temp_end`) where python keywords do not align with CLI. These should be made uniform.

In order to rectify this, depending on how we feel about API breaking changes at this stage, we can do the following.

If CLI name adopted in Python:
- rename the positional argument to the new name
- Add the alias (old) term to KW only block
- Raise `DeprecationWarning` on using old kw.

If Python name adopted in CLI:
- Add new term as option
- Move the alias (old) term to a new argument group (c.f. #484) (Deprecated)
- Raise `DeprecationWarning` on using old term.

If we're not that concerned (code considered beta?) we can just rename the terms as is and update the tutorials accordingly.

alinelena:
+1 for model instead of model-path.

ElliottKasoar:
While we're breaking things, we should consider improving the name of `filter_func`, as discussed in #311, which doesn't need to be tied to #313, although it's worth considering the future inclusion of constraints.

ElliottKasoar:
I think `temp_max`, `temp_min` vs. `temp_start`, `temp_end` are only different between calculations (phonons vs MD, where they do slightly different things), rather than between the Python and the CLI for a given calculation.

I think @alinelena had a look through everything and didn't spot any obvious problems, so I may close this, unless there are any objections, @oerc0122?

PR #528, opened 2025-04-23
Rename model path

ElliottKasoar:
As discussed in #486, renames `model_path` to `model`, since in many cases it is not a path at all.

A couple of things to discuss, which I thought were better to do now that I've ensured all the tests pass:

- Do we want to deprecate `model_path`, rather than remove it completely?
  - If we keep it fully removed, is there an obvious place in the docs to flag this change?
- What do we want to allow through `calc_kwargs` still? Previously, `model_path` passed through `calc_kwargs` raised an error, so is it best to similarly also disallow `model` to be passed through `calc_kwargs` (as I've implemented currently, although I kept `model_path` disallowed too)?
  - The logic for allowing some is that `model` alternatives e.g. `path` are generic enough that it's possible it doesn't actually refer to a model for different MLIPs.


oerc0122 (dismissed):
Looks sensible. 

Given what we chatted about in the meeting too, I do think that we want to deprecate the `model_path` and `model` in `calc_kwargs`.

And I think that we're being as defensive as we need to be with regards to detecting duplicate passing while reducing our workload.

alinelena (dismissed):

ElliottKasoar:
I'll fix the rebase and then add back in `model_path` as a deprecated option, as discussed with @oerc0122.

ElliottKasoar:
> Looks sensible.
> 
> Given what we chatted about in the meeting too, I do think that we want to deprecate the `model_path` and `model` in `calc_kwargs`.
> 
> And I think that we're being as defensive as we need to be with regards to detecting duplicate passing while reducing our workload.

Small note, but since `model_path` in `calc_kwargs` already raised an error, I've kept this the case, so only `model` in `calc_kwargs` warns.

I've also changed it so `model_path` on its own warns, but raises an error if combined with `model`.

alinelena (approved):

PR #531, opened 2025-04-28
updated the uninstall command in all colab notebooks

catastropiyush:
Based on the following discussion [here](https://github.com/stfc/janus-core/pull/523)
I have updated the colab notebooks in the tutorials with the following command.
`! pip uninstall torch torchaudio torchvision transformers numpy -y`

Thank you,
Piyush

### DIFF
--- a/docs/source/tutorials/cli/geomopt.ipynb
+++ b/docs/source/tutorials/cli/geomopt.ipynb
@@ -39,7 +39,7 @@
     "# import locale\n",
     "# locale.getpreferredencoding = lambda: \"UTF-8\"\n",
     "\n",
-    "# ! pip uninstall torch torchaudio torchvision numpy -y\n",
+    "# ! pip uninstall torch torchaudio torchvision transformers numpy -y\n",
     "# ! uv pip install janus-core[all] data-tutorials torch==2.5.1 --system\n",
     "# get_ipython().kernel.do_shutdown(restart=True)"
    ]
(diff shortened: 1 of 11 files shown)


ElliottKasoar:
Thanks for this, @catastropiyush!

Looking at the [files changed tab](https://github.com/stfc/janus-core/pull/531/files), I can see that the uninstall lines looks like that we need, but the total diffs for each file are very large, likely due to your environment differing to where the notebooks were authored.

This makes it difficult to check that nothing else important has changed.

Do you mind if I push to this branch, to make it so that the diffs become only the line we're interested in?

If you would prefer to make the changes yourself, the simplest way is perhaps to restore the files to be equivalent to main again, make the change to the uninstall lines again, then use a partial git add, `git add -p [file]`, which lets you control changes you want to add, rather than adding all changes in a file together.

catastropiyush:
"Do you mind if I push to this branch, to make it so that the diffs become only the line we're interested in?"Yeah sure @ElliottKasoar 

I actually opened the files in colab itself and then saved it back to the repo, thats why the total diffs are large.

ElliottKasoar:
Updated to include the CLI tutorials, and rebased.

(Side note: we need to release for these to fully work now, maybe not until after #532 is merged, as the notebooks have been updated to swap `model_path` for `model`, but install from the last release).

oerc0122 (approved):
Once this is rebased, it looks fine to go in.

PR #532, opened 2025-04-28
Remove default arch

### LINKED ISSUE #525: Remove default architecture
Since mace is no longer always installed as part of janus-core, we may want to remove it as the default architecture.

This is also useful for the `set_calc` option when initialising calculations. By default, this will not overwrite an existing calculator, since a user may attach a custom calculator or run multiple calculations, but this also has the effect that when a file is read with calculation results using ASE, a new calculator will not be set by default, as ASE attaches one to make the results accessible.

We could, of course, simply document this slightly confusing behaviour, but removing the default architecture would solve this too, as `arch=None` would mean use a pre-attached calculator, rather than `mace_mp`, while `arch=mace` etc. would always (re)set the calculator, even if one is already attached.

ElliottKasoar:
Resolves #525

Removes default architecture from calculations and calculator helper.

As discussed with @alinelena / @oerc0122, in general, `arch=None` requires a calculator to already be attached to the `Atoms` object, but we catch and raise an error in the case of the fake ASE calculator (`SinglePointCalculator`).

Otherwise, with `arch` specified, the calculator will always be overwritten, as expected.


alinelena (dismissed):

janus_core/calculations/base.py
```diff
@@ -91,14 +89,13 @@ def __init__(
         struct: MaybeSequence[Atoms] | PathLike,
         *,
         calc_name: str = "base",
-        arch: Architectures = "mace_mp",
+        arch: Architectures | None = "None",
```
oerc0122, line 92:
```suggestion
        arch: Architectures | None = None,
```
?

ElliottKasoar: Ah yep, thanks!

janus_core/helpers/struct_io.py
```diff
@@ -203,6 +227,13 @@ def input_structs(
         )
         if logger:
             logger.info("Calculator attached to structure.")
+    else:
+        if struct.calc is None:
+            raise ValueError("A calculator must be attached to `struct`")
+        if isinstance(struct.calc, SinglePointCalculator):
+            raise ValueError(
+                "The attached calculator cannot be used for new calculations."
+            )
```
oerc0122, line 236:
```suggestion
    elif struct.calc is None:
        raise ValueError("A calculator must be attached to `struct`")
    elif isinstance(struct.calc, SinglePointCalculator):
        raise ValueError(
            "The attached calculator cannot be used for new calculations."
        )
```

oerc0122 (commented):
Apart from those two minor things, looks good.

oerc0122 (approved):

PR #539, opened 2025-05-13
Remove DGL MLIPs for Windows

ElliottKasoar:
MLIPs that depend on `dgl` haven't actually been working on Windows, but the tests pass due to the skips (e.g. https://github.com/stfc/janus-core/actions/runs/14931210938/job/41947679259).

```
Run uv sync --extra mace --extra m3gnet --extra alignn
Resolved 386 packages in 2ms
error: Distribution `dgl==2.1.0 @ registry+[https://pypi.org/simple`](https://pypi.org/simple%60) can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're on Windows (`win_amd64`), but `dgl` (v2.1.0) only has wheels for the following platforms: `manylinux1_x86_64`, `manylinux2014_aarch64`, `macosx_11_0_arm64`, `macosx_11_0_x86_64`
```

In theory there are other versions of dgl published for Windows, but since they've both stopped publishing on PyPI and stopped publishing even in their own repo for Windows (and MacOS), I think leaving it up to users is probably fine.

This includes a bit of a tidy up of the docs to reduce duplication - essentially moving from `dgl` specific info to a more general statement, as there are various other conflicts now too.

### DIFF
--- a/docs/source/user_guide/installation.rst
+++ b/docs/source/user_guide/installation.rst
@@ -62,16 +62,40 @@ Some libraries are not installed by default, but may improve performance, such a
 - `PyTorch implementation of DFTD3 <https://github.com/CheukHinHoJerry/torch-dftd.git>`_, which can be used by MACE calculations on GPU
 
 
-Updating DGL
-------------
+
+MLIP Incompatibilies
+====================
+
+Due to the different requirements of the MLIPs we support, it is not always possible to install all combinations of ``extras``.
+
+
+MLIPs requiring DGL
+------------------
 
 `DGL <https://github.com/dmlc/dgl>`_, which is a dependency of ``alignn`` and ``matgl``, no longer
 publishes to PyPI, and no longer publishes any packages for Windows or MacOS.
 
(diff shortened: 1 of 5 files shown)


docs/source/user_guide/installation.rst
```diff
@@ -62,16 +62,40 @@ Some libraries are not installed by default, but may improve performance, such a
...
+
+Several MLIP packages, including ``mattersim``, ``fairchem``, and newer versions of ``sevennet``,
+depend on versions of ``e3nn`` that are incompatible the version required by ``mace``. So these cannot
+be installed together.
+
+
+MLIPs with limited OS support
+-----------------------------
+
+Several MLIP packages have limited support on Windows We are currently unable to
```
oerc0122, line 99:
```suggestion
Several MLIP packages have limited support on Windows. We are currently unable to
```

oerc0122 (changes requested):
Tiny typo, otherwise happy.

docs/source/user_guide/installation.rst
ElliottKasoar: Thanks!

PR #540, opened 2025-05-13
Update Orb

ElliottKasoar:
Includes a few fixes, including a missing model URL.


oerc0122 (approved):
If it works, I'm happy.

PR #545, opened 2025-05-14
Fix docs warnings

ElliottKasoar:
Fixes warning from underline being too short, and missing mace imports, which have knock-on effects for training/pre-processing.


oerc0122 (approved):
Looks simple enough!

PR #546, opened 2025-05-16
Rename filter

### LINKED ISSUE #311: geomopt filter_func argument does not support ase.constraints as advertised
The docstring for `janus geomopt` includes

```
    filter_func : Optional[str]
        Name of filter function from ase.filters or ase.constraints, to apply
        constraints to atoms. If using --vectors only or --fully-opt, defaults to
        `FrechetCellFilter` if available, otherwise `ExpCellFilter`.
```

but the code does not actually seem to look for functions in ase.constraints. I would like to use the FixSymmetry constraint with a full geometry optimisation. Is there a way to make that work?

ElliottKasoar:
Resolves #311

Deprecating `filter_func` is a bit more complicated than previous deprecations, since it has both a non-`None` default, and we don't set it directly in the CLI.

For the CLI, the solution is a bit simpler, as we can disallow both, and treat the old `--filter_func` as before.

For the Python, to avoid changing the structure of the code too much e.g. with sentinels, which would make defaults more obscure, I've changed it so that `filter_func` will override `filter_class`, if it's set.

I think this has the behaviour we'd like, except in the case where a user explicitly sets both, but I think that's relatively unlikely.

I'll also open a new issue for adding constraints, which was part of #311, but is less destructive.


janus_core/calculations/geom_opt.py
```diff
@@ -114,7 +116,8 @@ def __init__(
         symmetrize: bool = False,
         symmetry_tolerance: float = 0.001,
         angle_tolerance: float = -1.0,
-        filter_func: Callable | str | None = FrechetCellFilter,
+        filter_class: Callable | str | None = FrechetCellFilter,
+        filter_func: Callable | str | None = None,
```
oerc0122, line 120: Do you want to relegate `filter_func` to a `**kwargs` so it's not visible in the function hinting?

janus_core/calculations/geom_opt.py
```diff
@@ -167,11 +170,13 @@ def __init__(
         angle_tolerance
             Angle precision for spglib symmetry determination, in degrees. Default is
             -1.0, which means an internally optimized routine is used to judge symmetry.
+        filter_class
+            Filter class, or name of class from ase.filters to wrap around atoms.
+            Default is `FrechetCellFilter`.
         filter_func
-            Filter function, or name of function from ase.filters to apply constraints
-            to atoms. Default is `FrechetCellFilter`.
+            Deprecated. Please use `filter_class`.
```
oerc0122, line 177: Numpydoc provides an [`Other Parameters`](https://numpydoc.readthedocs.io/en/latest/format.html#other-parameters) block to which we could relegate the deprecated args?

janus_core/calculations/geom_opt.py
```diff
@@ -233,6 +242,15 @@ def __init__(
         if not self.struct.calc:
             raise ValueError("Please attach a calculator to `struct`.")
 
+        # Warn for deprecated filter_func now that logging is set up
+        if filter_func:
+            warnings.warn(
+                "`filter_func` has been deprecated, but currently overrides "
+                "`filter_class`. Please only set `filter_class`.",
+                FutureWarning,
```
oerc0122, line 250: https://docs.python.org/3.11/library/exceptions.html#DeprecationWarning

janus_core/cli/geomopt.py
```diff
@@ -113,17 +113,23 @@ def geomopt(
...
             help=(
-                "Name of ASE filter/constraint function to use. If using "
+                "Name of ASE filter to wrap around atoms. If using "
                 "--opt-cell-lengths or --opt-cell-fully, defaults to "
                 "`FrechetCellFilter`."
             ),
             rich_help_panel="Calculation",
         ),
     ] = None,
+    filter_func: Annotated[
```
oerc0122, line 127: https://typer.tiangolo.com/tutorial/commands/help/#deprecate-a-command ?

oerc0122 (commented):
Just a couple of "more standard"(?) things to do with deprecation.

janus_core/calculations/geom_opt.py
ElliottKasoar: Surely this is intended to be seen by end users?

> exception FutureWarning
> Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

If I remember correctly, `DeprecationWarning` tended to be hidden by default (unless called via `__main__`), so wasn't really visible.

janus_core/cli/geomopt.py
ElliottKasoar: I did have a look at this, but I think it only applies to CLI commands, rather than their parameters. I think click has introduced a parameter deprecated option as of 8.2.0 (https://github.com/pallets/click/pull/2271), so when typer is updated for this, that may be useful in the future.

janus_core/calculations/geom_opt.py
oerc0122: Yep, looks like you're right.

janus_core/cli/geomopt.py
ElliottKasoar: I'll open an issue, but given that we don't plan to keep these as deprecated for very long, and it's unclear how long it'll take for typer to become compatible, I'm not sure it'll be worth it for these cases.

janus_core/calculations/geom_opt.py
ElliottKasoar: This almost seems to have the effect of making them (`filter_func`, alongside `model_path` if we went down this route) more prominent, since the section still shows up in the hints, but they're no longer next to their replacements, so I think I may leave it as is.

janus_core/calculations/geom_opt.py
oerc0122:
```python
if extras := kwargs.keys() - {"filter_func", "model_path"}:
    raise NameError("Invalid kwargs ({', '.join(extras)}) passed")
```
Maybe not though.

oerc0122 (approved):

janus_core/calculations/geom_opt.py
ElliottKasoar: What I was going to do was `kwargs.pop` (defaulting to `None`), then raise against a non-empty `kwargs`, since I want to keep how they're used/passed around otherwise, so the actual variables need to be set, but I think it's probably more trouble then it's worth.

ElliottKasoar:
(Draft while considering `--filter` as nicer CLI option)

oerc0122 (approved):
Looks good.

PR #552, opened 2025-05-27
Fix an incorrect pressure unit in the documentation

trdurrant:
The example md command given in the documentation gives the unit of pressure as bar, but this does not seem to be correct and I believe that it should be GPa


trdurrant:
Well there's no rush. I don't think I have the permissions to rerun the tests myself though

PR #557, opened 2025-06-15
enable hdf5 for phonons bands

alinelena:


janus_core/calculations/phonons.py
```diff
@@ -107,8 +107,8 @@ class Phonons(BaseCalculation):
         End temperature for thermal properties calculations, in K. Default is 1000.0.
     temp_step
         Temperature step for thermal properties calculations, in K. Default is 50.0.
-    force_consts_to_hdf5
```
ElliottKasoar, line 110: This probably ought to be deprecated first, given that we've had it a while?

janus_core/calculations/phonons.py
```diff
@@ -237,8 +237,8 @@ def __init__(
             End temperature for thermal calculations, in K. Default is 1000.0.
         temp_step
             Temperature step for thermal calculations, in K. Default is 50.0.
-        force_consts_to_hdf5
-            Whether to write force constants in hdf format or not. Default is True.
+        hdf5
```
ElliottKasoar, line 240: Is it definitely fine having these as a combined option, rather than two separate options?

ElliottKasoar (commented):
Generally looks good, thanks!

We should add a test for the hdf5 bands, and it looks like there are a couple of `LZMAError`s in the tests still.

janus_core/calculations/phonons.py
```diff
@@ -341,9 +341,15 @@ def __init__(
...
+            if hdf5:
+                self.bands_file = self._build_filename("bands.hdf5")
+            else:
+                self.bands_file = self._build_filename("bands.yml")
         else:
-            self.bands_file = self._build_filename("auto_bands.yml.xz")
+            if hdf5:
+                self.bands_file = self._build_filename("auto_bands.hdf5")
+            else:
+                self.bands_file = self._build_filename("auto_bands.yml")
```
oerc0122, line 352: Is this clearer?
```python
filename = "bands" + (".hdf5" if hdf5 else ".yml")
if not self.qpoint_file:
    filename = f"auto_{filename}"
self.bands_file = self._build_filename(filename)
```

oerc0122 (dismissed):
Looks ok codewise when it works, just one small potential clarification.

janus_core/calculations/phonons.py
```diff
@@ -159,7 +162,8 @@ def __init__(
         temp_min: float = 0.0,
         temp_max: float = 1000.0,
         temp_step: float = 50.0,
-        force_consts_to_hdf5: bool = True,
+        force_consts_to_hdf5: None = None,
```
ElliottKasoar, line 165:
```suggestion
        force_consts_to_hdf5: bool | None = None,
```

janus_core/calculations/phonons.py
```diff
@@ -278,11 +284,27 @@ def __init__(
         self.temp_max = temp_max
         self.temp_step = temp_step
         self.force_consts_to_hdf5 = force_consts_to_hdf5
```
ElliottKasoar, line 286:
```suggestion
```
I don't think we need this

janus_core/calculations/phonons.py
```diff
@@ -278,11 +284,27 @@ def __init__(
...
+        self.hdf5 = hdf5
         self.plot_to_file = plot_to_file
         self.write_results = write_results
         self.write_full = write_full
         self.enable_progress_bar = enable_progress_bar
 
+        # Handle deprecation
+        if force_consts_to_hdf5:
```
ElliottKasoar, line 294:
```suggestion
        if force_consts_to_hdf5 is not None:
```
Otherwise it won't catch setting it to `False`.

janus_core/calculations/phonons.py
```diff
@@ -278,11 +284,27 @@ def __init__(
...
+            if hdf5:
```
ElliottKasoar, line 295:
```suggestion
            if hdf5 is False:
```
Since the default for `hdf5` is `True`, I think we may want to check against it being set `False`, otherwise this triggers even if only `force_consts_to_hdf5` is set.

Alternatively, we could set the default `hdf5` to `None` initially, and set the actual default here, which would allow us to do the same `not None` check as above, and also catch setting `hdf5` to be `True`.

ElliottKasoar (commented):
Generally looks great, thanks!

Should we perhaps add a quick check that the new hdf5 can be loaded in at least one of the tests? All the places where we used to open the .xz file are now just yaml, so I don't think we test it anywhere.

ElliottKasoar (approved):

PR #558, opened 2025-06-16
Add example config files

### LINKED ISSUE #423: Add example config files
Similar to the configs in [janus-tutorials](https://github.com/stfc/janus-tutorials/tree/main/configs), we should add example config files, including for NEB, once #375 is resolved.

We should consider where the best place to put them is.

ElliottKasoar:
Resolves #423

### DIFF
--- a/docs/source/examples/descriptors.rst
+++ b/docs/source/examples/descriptors.rst
@@ -1,0 +1,11 @@
+===========
+Descriptors
+===========
+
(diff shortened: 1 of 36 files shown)


ElliottKasoar:
(New README link won't work until we next release, but it's probably worth putting in now given that the old link has examples which are very outdated. We could do an incredibly minor release if we want to avoid a dead link for a couple of weeks?)

PR #560, opened 2025-06-18
Update docs

ElliottKasoar:
Fixes the link to view/edit documentation, and adds a link to GitHub in the footer, based on what furo do themselves: https://github.com/pradyunsg/furo/blob/main/docs/conf.py#L90, since it's often something I look for/use in other code's docs.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -128,9 +128,8 @@ html_logo = "images/janus-core.png"
 html_theme_options = {
     "source_repository": "https://github.com/stfc/janus-core/",
     "source_branch": "main",
-    "source_directory": "docs/",
+    "source_directory": "docs/source",
 }
-
 
 # Add any paths that contain custom themes here, relative to this directory.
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -128,10 +128,21 @@
 html_theme_options = {
```
oerc0122, line 128: Does
```
html_context = {
    "display_github": True, # Integrate GitHub
    "github_user": "stfc", # Username
    "github_repo": "janus-core", # Repo name
    "github_version": "main", # Version
    "conf_py_path": "/docs/", # Path in the checkout to the docs root
}
```
work?

ElliottKasoar: I don't think it adds a GitHub link anywhere, if that's what you mean.

I was also hoping to set `icon_links` but it doesn't sound like the them is staying quite minimalistic: https://github.com/pradyunsg/furo/discussions/176

oerc0122 (approved):

PR #561, opened 2025-06-18
Update deps

ElliottKasoar:
Updates various pre-commit/CI dev dependencies.

Also removes the pin from uv in a few places, which was already inconsistent.


ElliottKasoar:
> Although we might want an exact pin on ruff as they add new checks between versions?

The pre-commit ruff is still pinned. I should really look at the automated code suggestions at some point, which would make a pinned version slightly less important.

Issue #570, opened 2025-07-11
Tidy tutorials


ElliottKasoar:
A few issues/improvements to address in the tutorial:

- A few data files are missing when running on Colab
- It might be worth splitting out the uninstall/reinstall of torch into new lines/cell, since to make it easier to skip if the current Colab torch is fine at a given time
- It would be useful to add an overall aim at the start of each tutorial, as we have in a few cases e.g. NEB
- We're missing an EoS CLI tutorial
- It might be worth changing `janus-core[all]` to `janus-core[mace]` and `janus-core[mace,sevennet]` etc.
  - This saves the need to unnecessary packages e.g. tensorflow
  - This also shows a bit more about how to install different models

PR #571, opened 2025-07-14
Fix Colab data and instructions

ElliottKasoar:
Address the quick parts of #570:

- Fixes missing data
- Splits up reinstallation of `torch` and `numpy`
- Reduces installed extras

Also adds more explanation of what the Colab installation is doing, and adds a way to see which version of `janus-core` is installed for quicker debugging.


ElliottKasoar:
Now also removes remaining references to `janus-tutorials`, including copying over missing data files.

alinelena (approved):

PR #573, opened 2025-07-23
Add calculator option for dispersion correction

### LINKED ISSUE #567: Add dispersion option for all calculators?
To be able to compare different calculators e.g. to develop a testing framework, it's useful to have the option to add D3 to any calculator, even if it's not directly supported by the calculator (as it is for MACE).

We could turn `dispersion` into an option interpreted when `choose_calculator` is called, similar to `dtype`, and if it's not supported by a given calculator, we could do the conversion to a `SumCalculator` ourselves. See also: https://github.com/joehart2001/mlipx/blob/main/mlipx/recipes/models.py.jinja2#L216

ElliottKasoar:
Resolves #567

- Adds `d3` optional extra
- Adds `dispersion` and `dispersion_kwargs` to `choose_calculator`, allowing any calculator to be converted into a `SumCalculator` composed of the original calculator, and the `TorchDFTD3Calculator`.
  - This is not currently exposed directly to calculations, but can be set via `calc_kwargs`. The option remains open to add these as full parameters to all calculations in the future, and it's documented as is for now
- Adds `_d3` label to `calc.parameters.arch` e.g. `mace_mp_d3_energy`. This is also the case for values that don't depend on the dispersion, e.g. descriptors, but I thought it might be better to be consistent with the `arch` label. I'm also very happy to revert to `mace_mp_descriptors` if this is preferable.
- For `mace_mp`, I use their default parameters, as it felt like the least worst option, as I'm not sure it makes sense to apply these defaults for all calculators, but it would also be confusing changing the default `mace_mp` parameters.

### DIFF
--- a/janus_core/helpers/mlip_calculators.py
+++ b/janus_core/helpers/mlip_calculators.py
@@ -11,7 +11,6 @@ from __future__ import annotations
 from os import environ
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, get_args
-import warnings
 
 from ase import units
 from ase.calculators.mixing import SumCalculator
@@ -205,13 +204,7 @@ def choose_calculator(
             model = model if model else "small"
             kwargs.setdefault("default_dtype", "float64")
 
-            # Remove dispersion (to be added later for any calculator)
-            if dispersion and not kwargs.get("dispersion", True):
-                warnings.warn(
-                    "`dispersion` set in two places. kwargs value will be used",
-                    stacklevel=2,
-                )
-            dispersion = kwargs.pop("dispersion", dispersion)
+            # Set mace_mp dispersion defaults
             dispersion_kwargs.setdefault("damping", kwargs.pop("damping", "bj"))
             dispersion_kwargs.setdefault("xc", kwargs.pop("dispersion_xc", "pbe"))
             dispersion_kwargs.setdefault(
--- a/janus_core/helpers/mlip_calculators.py
+++ b/janus_core/helpers/mlip_calculators.py
@@ -11,8 +11,12 @@ from __future__ import annotations
 from os import environ
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, get_args
+import warnings
 
+from ase import units
 from ase.calculators.mixing import SumCalculator
+from torch import get_default_dtype
+from torch_dftd.torch_dftd3_calculator import TorchDFTD3Calculator
 
 from janus_core.helpers.janus_types import Architectures, Devices, PathLike
 from janus_core.helpers.utils import none_to_dict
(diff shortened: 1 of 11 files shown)


janus_core/helpers/mlip_calculators.py
```diff
@@ -146,6 +205,19 @@ def choose_calculator(
             model = model if model else "small"
             kwargs.setdefault("default_dtype", "float64")
 
+            # Remove dispersion (to be added later for any calculator)
+            if dispersion and not kwargs.get("dispersion", True):
```
oerc0122, line 209: `dispersion` is an argument and thus can't be in `kwargs`?

tests/test_mlip_calculators.py
```diff
@@ -244,3 +244,24 @@ def test_invalid_device(arch):
...
+    assert calculator.parameters["model"] is not None
+    assert calculator.parameters["arch"] == "mace_mp_d3"
+
+
+def test_d3_manual():
+    """Test adding D3 dispersion calculator manually."""
+    skip_extras("mace_mp")
+
+    calculator = choose_calculator(arch="mace_mp")
+    calculator = add_dispersion(calculator)
```
oerc0122, line 264: What happens if you add a dispersion whose `device` is different to that of the calculator?

oerc0122 (changes requested):
Couple of concerns to look at.

tests/test_mlip_calculators.py
ElliottKasoar: I think that should just work, since it basically just runs each calculation independently, and the calculator is responsible preparing the data for its device.

I can't test easily on Mac since the mace models are tricky to run with mps, but e.g.

```
calculator = choose_calculator(arch="mace_mp", device="cuda")
calculator = add_dispersion(calculator, device="cpu")
```

seemed to be usable in a Colab notebook.

janus_core/helpers/mlip_calculators.py
ElliottKasoar: Good point! That simplifies things...

janus_core/helpers/mlip_calculators.py
```diff
@@ -146,6 +204,13 @@ def choose_calculator(
             model = model if model else "small"
             kwargs.setdefault("default_dtype", "float64")
 
+            # Set mace_mp dispersion defaults
+            dispersion_kwargs.setdefault("damping", kwargs.pop("damping", "bj"))
+            dispersion_kwargs.setdefault("xc", kwargs.pop("dispersion_xc", "pbe"))
+            dispersion_kwargs.setdefault(
+                "cutoff", kwargs.pop("dispersion_cutoff", 40.0 * units.Bohr)
+            )
```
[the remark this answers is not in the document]
ElliottKasoar: I think when this is used, they may be set somewhat often, e.g. since we're starting to move away from models just trained on PBE, but I think it's probably something to come back to (similarly to potentially exposing `dispersion`/`dispersion_kwargs` separately to `calc_kwargs`).

Not a blocker, but we would also need to be a little careful with defaults, to make sure we can still overwrite the defaults for mace_mp, while not overwriting user inputs.

Issue #572, opened 2025-07-21
few new potentials arrived


alinelena:
- [ ] nep
 - [ ] mad-pet
 - [ ] uma

we shall add them

ElliottKasoar:
The main NEP codebase doesn't seem to have an ASE calculator/is GPU only: https://github.com/brucefan1983/GPUMD, so it seems there are only externally maintained calculators: https://github.com/brucefan1983/GPUMD/issues/215#issuecomment-1184085972

Maybe the `calorine` one is more likely to be future-proof?

ajjackson:
Ah, good to know! I've been finding SevenNet geometry optimisation bottoms out a bit higher than the MACE-based potentials (around 5e-6 rather than < 1e-6); should be good enough for phonons but also seems more likely to break symmetry. Still, this is just anecdotal 😄

The best MLIP for foundation models is not necessarily the best foundation model for bespoke potentials. But it's nice to have something to fine-tune.

Here is the link to NEP89 by the way: https://github.com/brucefan1983/GPUMD/blob/master/potentials/nep/nep89_20250409/nep89_20250409.txt

ElliottKasoar:
@ajjackson It turns out there are actually very(!) recently released foundation models: https://www.nequip.net/models

I don't believe they're built into the package they provide though, so I'm not sure we can support them quite as easily as others.

ajjackson:
The Zenodo download links should be pretty stable if we fancy it https://zenodo.org/records/16980200

And I see a very familiar name on that dataset... Pinging @kavanase ! STFC maintains this package as a convenient front-end to lots of foundation models. Most of them will automatically download the data for the user. Would the data linked here be appropriate for inclusion, or would you prefer that we hold off for another version? (Or perhaps there is an existing auto-download shim for this data you can recommend?)

kavanase:
Hi @ajjackson!
Yes that data is appropriate for inclusion, though we will be releasing a v1 of auto-downloading via `nequip` from nequip.net very soon (potentially later today), which would be the best approach to use for future-proofing. Pinging @cw-tan as the lead on that.

We will likely have updated foundation potentials in the near-ish future (~1 month) as we converge some training parameter sweeps, but should only be relatively minor improvements and easy to handle (if any changes needed) with the `nequip` auto-download approach.

PR #575, opened 2025-07-23
Add PET-MAD

ElliottKasoar:
See: #572


oerc0122 (commented):
~Is this actually implemented?~ Ignore that, it was showing the last diff.

PR #581, opened 2025-08-04
Update issue templates

alinelena:


.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -0,0 +1,37 @@
...
+        A clear and concise description of the bug, including a minimal reproducible example.
+
+        Please include your full output with the complete error message.
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: Platform
+      description: What operating system and architecture are you using? (see `uname -orsm`)
```
oerc0122, line 18: Stating the obvious, but `uname` won't work on Windows. Perhaps a python sys.platform equivalent

.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -0,0 +1,37 @@
...
+      label: Version
+      description: What version of janus-core are you using? (see `janus --version`)
+      placeholder: "e.g., janus-core version: 0.8.3"
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: Python version
+      description: What version of Python are you using? (see `python3 --version`)
```
oerc0122, line 34: Also may not work with full virtualenvs like Conda/UV. Will need to ensure it's in the right environment

.github/ISSUE_TEMPLATE/3_question.yml
```diff
@@ -0,0 +1,26 @@
...
+    attributes:
+      label: Question
+      description: Describe your question in detail.
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: Platform
+      description: What operating system and architecture are you using? (see `uname -orsm`)
```
oerc0122, line 15: As above

.github/ISSUE_TEMPLATE/1_bug_report.yml
oerc0122: Well, if UV're happy with it, I'm mostly happy with it. I'm not sure how often Python version will really be relevant either.

ElliottKasoar: We've seen a few bugs from (sub) dependencies that only affect specific Python version, so it can be useful.

Maybe more importantly, we've had several people try to install things with Python 3.6 etc, so if the "bug" is being unable to install anything, that becomes much easier to catch.

.github/ISSUE_TEMPLATE/1_bug_report.yml
ElliottKasoar: This is also borrowed from `uv`. It will tell you about wsl, which is nice e.g. `Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 GNU/Linux`

`sys.platform` doesn't seem to give much information, at least for me (just `darwin`), so I'm not sure it's much more useful than people hopefully just telling us Windows version x?

PR #635, opened 2025-12-09
Adds compression for HDF5 force constants

### LINKED ISSUE #356: phonon bands compressed make it option
at the moment we face a dlimea for band when we write them full with eigenvectors. files are big for a mof for example we are looking at 30GB+ but fast to write, for compressed we have 4GB but takes hours to write and compress.

suggestion is to add an option
write-compressed: True/False to control this behaviour and choose the optimal one.

harveydevereux:
Closes #356 by adding support for the compression option in ```write_force_constants_to_hdf5```.

### DIFF
--- a/janus_core/calculations/phonons.py
+++ b/janus_core/calculations/phonons.py
@@ -249,7 +249,7 @@ class Phonons(BaseCalculation):
         hdf5
             Whether to write force constants and bands in hdf5 or not. Default is True.
         hdf5_compression
-            Compression scheme to use for force constants. Default is None.
+            Compression scheme to use for force constants, gzip or lzf. Default is None.
         plot_to_file
             Whether to plot various graphs as band stuctures, dos/pdos in svg.
             Default is False.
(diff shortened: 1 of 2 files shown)


janus_core/calculations/phonons.py
```diff
@@ -9,7 +9,9 @@
 from ase import Atoms
 from numpy import ndarray
 import phonopy
-from phonopy.file_IO import write_force_constants_to_hdf5
+from phonopy.file_IO import (
+    write_force_constants_to_hdf5,
+)
```
ElliottKasoar:
```suggestion
from phonopy.file_IO import write_force_constants_to_hdf5
```

janus_core/calculations/phonons.py
```diff
@@ -586,6 +593,8 @@ def write_force_constants(
         hdf5
             Whether to save the force constants separately to an hdf5 file. Default is
             self.hdf5.
+        compression
+            Compression scheme to use fo hdf5 file. Default is None.
```
ElliottKasoar:
```suggestion
            Compression scheme to use fo hdf5 file, gzip or lzf. Default is None.
```

PR #644, opened 2025-12-10
590 alphanet

### LINKED ISSUE #590: Add AlphaNet

JessieGould:
closes: #590


pyproject.toml
```diff
@@ -41,6 +41,14 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+alphanet = [
+    "alphanet @ git+https://github.com/zmyybc/AlphaNet.git",
+    "numpy<2.0",
```
ElliottKasoar: Does it not work with numpy 2+? I didn't see anything in their requirements etc. about numpy explicitly?

tests/test_mlip_calculators.py
```diff
@@ -49,6 +49,10 @@
 
 NEQUIP_PATH = MODEL_PATH / "toluene.nequip.pth"
 
+ALPHANET_PATH = MODEL_PATH / "alphanet"
+ALPHANET_CKPT = ALPHANET_PATH / "test_model.ckpt"
```
ElliottKasoar: Is there a way to download this as part of the test (normally to the local cache)? That is what we do for quite a few of the models, and ideally there should be some sort of function built into AlphaNet to do this by the relevant label.

janus_core/helpers/mlip_calculators.py
```diff
@@ -310,6 +310,49 @@ def choose_calculator(
...
+
+            # Get config path from kwargs or auto-detect
+            config_path = kwargs.pop("config", None)
+            if config_path is None and isinstance(model, Path):
+                # Auto-detect: try stem.json then parent_name.json
+                config_path = model.parent / f"{model.stem}.json"
+                if not config_path.exists():
+                    config_path = model.parent / f"{model.parent.name}.json"
+            
+            # If model is still a string (model name), config must be provided
```
ElliottKasoar: This seems unrelated to whether the model is a string?

ElliottKasoar (commented):
Can you add a list of things left to do e.g. if you add a space in between the square brackets below, it'll create a checkbox

- [] Update README

The developer guide may be slightly outdated in terms of what it suggests in terms of documenting a few model, but the main places will be the README and the user guide in `docs`.

We also need single point tests

tests/test_mlip_calculators.py
alinelena: in theory we can use the same approach from data-tutorials if there is no option on their side

PR #647, opened 2025-12-11
Rotate results directories

### LINKED ISSUE #472: Create new results directory for new calculations
Assuming #462 or similar functionality is merged, a single output directory, `janus_results` will be created, but repeated calculations may overwrite output files, which is consistent with current behaviour.

Ideally, we will have an (optional) way of specifying whether to use the existing output directory, or create new ones, e.g. `janus_results_1`, `janus_results_2`, etc. for each new calculation.

Some care will need to be taken about when the new directory name is determined and created, particularly in cases where calculations call other calculations, or where output files may be defined and written at different stages of the calculation, such as running different components of phonon calculations, which is why I think it's best to treat it separately to #462.

ElliottKasoar:
Resolves #472

Creates new results directories if one ore more already exist.

This is implemented similarly to [dlpoly-py](https://gitlab.com/ccp5/dlpoly-py/-/blob/devel/dlpoly/utility.py?ref_type=heads#L114), with a few small changes, but functionally I think it works more or less in the same way.

The scheme implemented is `janus_results` -> `janus_results_1` -> `janus_results_2`, etc. and will ignore output directories formatted differently e.g. `janus_results1` would not be considered.

As far as I can tell, the only place in the code that this impacts is our MD restart, so we now search for the potential last results directory to get the relevant restart file.

### DIFF
--- a/tests/test_filenamemixin.py
+++ b/tests/test_filenamemixin.py
@@ -131,14 +131,20 @@ def test_file_name_mixin_build(mixin_params, file_args, file_kwargs, file_name):
 @pytest.mark.parametrize(
     "existing_dirs, expected_new_dir",
     (
-        # Defaults to structure atoms from ASE
         ((), "janus_results"),
         (("janus_results",), "janus_results_1"),
-        (("janus_results_1",), "janus_results_2"),
         (("janus_results", "janus_results_1"), "janus_results_2"),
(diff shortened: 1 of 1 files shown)


oerc0122 (changes requested):
I still have a gut feeling the logic could be simpler, but with a numeric sort this should do the job nicely.

PR #653, opened 2026-01-13
Add nequip to train.

harveydevereux:
This will add nequip as a ```janus train [architecture]``` option, where the [architecture] is now a required argument to train. All ```Architectures``` other than ```"mace"```, ```"mace_mp"```, ```"mace_off"```, ```"mace_omol"```, and ```"nequip"``` raise an error.

It does work with these inputs [inputs.zip](https://github.com/user-attachments/files/24593835/inputs.zip)

Todo:

- [x] Testcase with ```janus train nequip```.
- [x] Fine-tuning (and test).
- [x] Documentation.

One thing I noticed is that ```janus train mace ...``` will currently output outside ```janus_results```, but the way I have it for nequip it goes into ```janus_results```. What's better? E.g. for mace,

```shell
checkpoints  config.yml  janus_results  logs  results  test_compiled.model  test.model
```

Or is that something that is control by the mace ```config.yml```? For nequip I've had to set it manually so I chose ```./janus_results```.

---

#### Details

Similar to ```mace.cli.run_train.run```, nequip has a ```nequip.scripts.train.main``` entrypoint - but it uses [Hydra](https://github.com/facebookresearch/hydra) to manage configuration [here](https://github.com/mir-group/nequip/blob/main/nequip/scripts/train.py#L31).

I've done similar to the current code and used the ```main``` function rather than a subprocess. This means using the [Compose API](https://hydra.cc/docs/advanced/compose_api/) which at least one [downsides](https://github.com/facebookresearch/hydra/issues/2456) - the ```hydra.runtime.output_dir``` is not set, but is required, and we have to manually set the HydraConfig singleton (as far as I've found...).

This does mean that we have a match that obtains the runner function and its arguments, seperating out modules to avoid conflicts (mace/nequip in fact is one).

### DIFF
--- a/janus_core/cli/train.py
+++ b/janus_core/cli/train.py
@@ -23,14 +23,14 @@ def train(
         bool, Option(help="Whether to fine-tune a foundational model.")
     ] = False,
     file_prefix: Annotated[
-        Path,
+        Path | None,
         Option(
             help=(
                 "Prefix for output files, including directories."
                 "Default directory is ./janus_results."
             )
         ),
-    ] = Path.cwd() / "janus_results",
+    ] = None,
     log: LogPath | None = None,
     tracker: bool = True,
     summary: Summary | None = None,
@@ -124,6 +124,9 @@ def train(
                 )
         case _:
             raise ValueError(f"Unsupported Architecture ({arch})")
+
+    if file_prefix is None:
+        file_prefix = Path.cwd() / "janus_results"
 
     if log is None:
         log = file_prefix / "train-log.yml"


ElliottKasoar:
Thank you so much for this!

How does nequip fine-tuning work?

> One thing I noticed is that janus train mace ... will currently output outside janus_results, but the way I have it for nequip it goes into janus_results. What's better? E.g. for mace,

Perhaps we should set the default value for mace's `--work_dir` (https://github.com/ACEsuit/mace/blob/main/mace/tools/arg_parser.py#L38) to janus-core, and then do the equivalent for any new architecture?

> This does mean that we have a match that obtains the runner function and its arguments, seperating out modules to avoid conflicts (mace/nequip in fact is one).

I think this was probably unavoidable given how different training entry points will be, so I think this is fine?

ElliottKasoar:
> > How does nequip fine-tuning work?
> 
> For fine-tuning it looks simple from what I can tell. For training the [config](https://github.com/mir-group/nequip-tutorial/blob/main/config.yaml) has this model section
> 
> ```yaml
>   model:
>     _target_: nequip.model.NequIPGNNModel
> # Then parameters
> ```
> 
> for fine-tuning it looks like [this](https://github.com/mir-group/nequip-tutorial/blob/main/config_finetuning.yaml)
> 
> ```yaml
>   model:
>     _target_: nequip.model.ModelFromPackage
>     package_path: ${model_package_path}
> ```
> 
> where further up the file the path is set
> 
> ```yaml
> model_package_path: /content/NequIP-OAM-L-0.1.nequip.zip
> ```

Ok cool, should we do a similar existence check for `model_package_path` if fine-tuning then as we do for mace?

janus_core/cli/train.py
```diff
@@ -8,65 +8,126 @@
...
+            # See nequip.model.__all__
+            if fine_tune and model not in (
+                "nequip.model.ModelFromCheckpoint",
+                "nequip.model.ModelFromPackage",
+            ):
+                raise ValueError(
+                    """Fine-tuning requested but there is no checkpoint or
+                    package specified in your config."""
+                )
+
```
oerc0122: `case _: raise`

janus_core/cli/train.py
```diff
@@ -8,65 +8,126 @@
...
+    file_prefix: Annotated[
+        Path,
+        Option(
+            help=(
+                "Prefix for output files, including directories."
+                "Default directory is ./janus_results."
+            )
+        ),
+    ] = Path("./janus_results"),
+    log: LogPath = None,
```
oerc0122: Violates the type hint.

tests/models/extra_models.sh
oerc0122: Shell script is not portable. Likewise filepath is not necessarily portable. Probably want Python script.

tests/test_train_cli.py
```diff
@@ -234,23 +311,30 @@ def test_fine_tune_invalid_foundation(tmp_path):
...
+        model_path = results_dir / "test.model"
+        compiled_path = results_dir / "test_compiled.model"
+        checkpoints_path = results_dir / "checkpoints"
+        logs_path = results_dir / "logs"
+
         log_path = tmp_path / "test.log"
         summary_path = tmp_path / "summary.yml"
 
-        config = write_tmp_config(DATA_PATH / "mlip_train.yml", Path())
+        config = write_tmp_config_mace(DATA_PATH / "mlip_train.yml", Path())
```
oerc0122:
```suggestion
        config = write_tmp_config_mace(DATA_PATH / "mlip_train.yml", Path.cwd())
```

janus_core/cli/train.py
```diff
@@ -8,65 +8,126 @@
...
+                raise ValueError(
+                    """Fine-tuning requested but there is no checkpoint or
+                    package specified in your config."""
+                )
+
+    if log is None:
+        log = file_prefix / "train-log.yml"
+
+    if summary is None:
+        summary = file_prefix / "train-summary.yml"
```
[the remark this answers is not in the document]
harveydevereux: That can't be done with both still depending on file_prefix arg

oerc0122 (approved):
Looks good if it works.

alinelena:
@oerc0122 @harveydevereux @ElliottKasoar anything blocking in here?

Issue #648, opened 2025-12-19
MD Restarts with MACE_MP+D3


trdurrant:
### Summary

I am seeing a problem restarting MACE_MP calculations, that include dispersion:

```
from janus_core.calculations.md import NVT

from ase.build import molecule
mol = molecule('H2O')

opt = NVT(
    struct=mol,
    arch="mace_mp",
    model="medium",
    device="cpu",
    calc_kwargs={"default_dtype":"float64", "dispersion":True},
    temp=300.0,
    timestep=1.0,
    steps=100,
    restart_every=100,
    stats_every=100,
    traj_every=100,
    restart=False,
    )

opt.run()

opt = NVT(
    struct=mol,
    arch="mace_mp",
    model="medium",
    device="cpu",
    calc_kwargs={"default_dtype":"float64", "dispersion":True},
    temp=300.0,
    timestep=1.0,
    steps=200,
    restart_every=100,
    stats_every=100,
    traj_every=100,
    restart=True,
    )

opt.run()
```

which produces the following error:

```
Traceback (most recent call last):
  File "/home/tom/janus_bug/md.py", line 23, in <module>
    opt = NVT(
          ^^^^
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/calculations/md.py", line 1706, in __init__
    super().__init__(*args, ensemble=ensemble, **kwargs)
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/calculations/md.py", line 545, in __init__
    self._prepare_restart()
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/calculations/md.py", line 822, in _prepare_restart
    self.struct, self.struct_path = input_structs(
                                    ^^^^^^^^^^^^^^
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/helpers/struct_io.py", line 221, in input_structs
    attach_calculator(
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/helpers/struct_io.py", line 98, in attach_calculator
    calculator = choose_calculator(
                 ^^^^^^^^^^^^^^^^^^
  File "/home/tom/janus_bug/local/lib/python3.12/site-packages/janus_core/helpers/mlip_calculators.py", line 487, in choose_calculator
    raise ValueError(
[...]

PR #657, opened 2026-01-30
Bugfix for MD restarts with D3+mace

trdurrant:
Resolves https://github.com/stfc/janus-core/issues/648


ElliottKasoar:
> this is correct in fairness I would prefer not to append _d3 in the first place but have something like arch=mace_mp dispersion=T in the info.

The downside to this (and I think why we didn't take this approach originally) is it becomes more ambiguous if you wanted to store multiple MLIP results

PR #660, opened 2026-02-02
Add qpoints phonons mode as available in phonopy

dmt4:
Solves issue 659


oerc0122 (approved):
If it works, looks good.

PR #662, opened 2026-02-03
Add SevenNet to train

harveydevereux:
- [x] train support for SevenNet```
- [x] fine-tuning
- [x] docs
- [x] test train
- [x] test fine-tune

The only curve ball with this one is that the atom attributes ```forces```, ```energy```, and ```stress``` must be called that, as  far as I can tell.


oerc0122 (dismissed):
Just some minor optional things, otherwise looks good.

docs/source/user_guide/command_line.rst
```diff
@@ -700,6 +700,11 @@ Configuration of Nequip training is outlined in the `Nequip user guide <https://
 The results directory contents depends on the options selected in the configuration file, but may typically contain model checkpoint, ``.ckpt``, files and a metrics directory.
 
 
+Training SevenNet MLIPS
```
ElliottKasoar:
```suggestion
Training SevenNet MLIPs
```
Very minor nitpick...

Should be the same for `Nequip MLIPS` which it won't let me suggest

janus_core/cli/train.py
```diff
@@ -122,6 +122,27 @@ def train(
...
+                raise ValueError(
+                    """Fine-tuning requested but there is no continue
+                    section in your config."""
+                )
+                model = continue_section.get("checkpoint")
+                if model is None:
+                    raise ValueError(
+                        """No model specified as a checkpoint for
+                        fine-tuning.
+                        """
```
ElliottKasoar:
```suggestion
                        "No model specified as a checkpoint for fine-tuning."
```
Not important but looks odd otherwise...

tests/test_train_cli.py
```diff
@@ -117,6 +128,59 @@ def write_tmp_config_nequip(
...
+    for dataset in config["data"].keys() & {"load_dataset_path", "load_validset_path"}:
+        files = config["data"][dataset]
+        for i, file in enumerate(files):
+            name = Path(file).name
+            path = DATA_PATH / name
+            if path.exists():
+                frames = read(path, index=":")
+                # There is currenlty no option to rename these.
+                rename_info = {"dft_energy": "energy", "dft_stress": "stress"}
+                rename_arrays = {"dft_forces": "forces"}
```
ElliottKasoar, line 163: Is this so we can use the same training data xyz files for all the models we train? Presumably this wouldn't be necessary if everything was configured for SevenNet?

tests/models/extra_models.py
ElliottKasoar, line 1: I know this isn't newly introduced, but I just realised this isn't run as part of the tests, but only in the CI? Ideally we'd probably encourage devs to run this and/or include downloading them (optionally) locally?

Otherwise it might be confusing to run the tests fine locally and then things fail in the CI?

tests/test_train_cli.py
```diff
@@ -101,6 +106,12 @@ def write_tmp_config_nequip(
         pth = DATA_PATH / Path(config["data"][file]).name
         if pth.is_file():
             config["data"][file] = str(pth)
+    if fine_tune:
+        model_dict = config["training_module"]["model"]
+        model = Path(model_dict[f"{model_type}_path"]).name
+        for pth in (model, f"extra/{model}"):
```
ElliottKasoar: Does the string `/` work for Windows?

tests/test_train_cli.py
```diff
@@ -227,7 +291,7 @@ def test_fine_tune(tmp_path):
         summary_path = tmp_path / "summary.yml"
         logs_path = results_dir / "logs"
 
-        config = write_tmp_config_mace(DATA_PATH / "mlip_fine_tune.yml", Path())
+        config = write_tmp_config_mace(DATA_PATH / "mlip_fine_tune.yml", Path.cwd())
```
ElliottKasoar, line 294: Why has this changed?

tests/test_train_cli.py
oerc0122: For `Path` objects it should.

ElliottKasoar: It definitely does if you have "x" / "y" but are you sure it does if you have "x/y"? I thought we ran into problems with this in the past

tests/test_train_cli.py
harveydevereux: Either it was suggested or suggested to be used in other situation so applied here for getting the CWD

ElliottKasoar:
Looks like some 503 errors, @harveydevereux. We may want some try...excepts for the download extras function, similar to what we have in some of the tests, since we can't really do anything about these.

ElliottKasoar (approved):

PR #663, opened 2026-02-05
Add grace to train

harveydevereux:
- [x] fine tuning
- [x] test fine tuning
- [ ] test training
- [ ] docs

Grace is quite slow it seems (60s for 1 epoch fine-tuning) and memory hungry (13Gb)


tests/data/mlip_train.pkl.gz
harveydevereux: This file would not be required if we upgraded to `tensorpotential 0.5.5` since that add `extxyz` support.

It is small though, and it requires various modifications to the xyz's we have as well

ElliottKasoar: We definitely should upgrade at some point, but it introduces conflicts with basically everything else via PyTorch/CUDA conflicts (see https://github.com/ICAMS/grace-tensorpotential/issues/23), if I remember correctly

This may be more tractable once we cut out some of the unsupported MLIPs from our extras

ElliottKasoar:
@harveydevereux I'm not expecting you to finish this off at this point, but if you have a second, can you remind me what still needs doing? Apart from now needing to rebase, was it mostly just the testing/docs (and trying to upgrade GRACE)?

PR #686, opened 2026-03-12
Optionally reset velocities for MD

### LINKED ISSUE #666: Initial velocities for MD non-restart simulations
### Question

For dynamics there is the following code (which will end up being called by `md.run()`:
```
    def _run_dynamics(self) -> None:
        """Run dynamics and/or temperature ramp."""
        [...]
        # Set velocities to match current temperature
        if not self.restart:
            self._set_velocity_distribution()
```
i.e. if the calculation is not a restart, set a Maxwell-Boltzmann distribution of velocities at `self.temp`

But this seems unexpected to me, as this means that user supplied velocities are discarded. This has a few effects:

- If I start a new MD run from the positions and velocities of an old run, the velocities will be re-initialised by this line
- For `NVE` calculations, it is not possible to for a user to set their own initial distribution of velocities, (if no temperature is directly set, then the default temperature of `300` is used)

In contrast, in the initialiser for the `NVT_CSVR` class, there is the following code:
```
        # Velocity distribution must be non-zero before dynamics is set
        if np.isclose(self.struct.get_kinetic_energy(), 0.0, rtol=0, atol=1e-12):
            if self.logger:
                self.logger.warning("Velocities modified during MD initialisation")
            self._set_velocity_distribution()
```

This is more in line with what I would have expected, but still in this case the velocity distribution set here will be replaced by the first code block, regardless of if this code triggers or not.

I would like to run `NVE` with a custom velocity distribution, but I don't see how to do this at the moment. Does it make sense to remove the initial block of code, and instead try and include something like the second block in each of the MD classes? Or perhaps I haven't understood this code correctly.

### Platform

_No response_

### Version

_No response_

ElliottKasoar:
Resolves #666

This is a slightly different fix to the discussion in the issue, to me it also seems like a reasonable solution that makes sense based on our current parameters (`rescale_velocities` in particular, since this already guarded an effectively duplicated `self._reset_velocities()` call):

- If `rescale_velocities` is `True` (default is `False`) then velocities are reset before heating and/or MD (assuming we're not restarting)
- If `rescale_velocities` is `False`, we don't reset the velocities, unless no velocities have been set, in the same way as `NVT_CSVR`. I'm also happy to leave the velocities at 0 if it makes more sense

I test that reading in a file with velocities (and `rescale_velocities` is `False`) keeps these velocities, so I think it addresses the main point of the issue.

There is also scope for debate about the additional resets we perform in between heating steps, and between heating and MD, but that could also be addressed separately.

How does this look/sound, @trdurrant?

### DIFF
--- a/janus_core/calculations/md.py
+++ b/janus_core/calculations/md.py
@@ -108,11 +108,13 @@ class MolecularDynamics(BaseCalculation):
     minimize_kwargs
         Keyword arguments to pass to geometry optimizer. Default is {}.
     rescale_velocities
-        Whether to rescale velocities. Default is False.
+        Whether to rescale velocities before dynamics and during equilibration.
+        Frequency during equilibration depends on `rescale_every`. Default is False.
     remove_rot
-        Whether to remove rotation. Default is False.
+        Whether to remove rotation when rescaling velocities. Default is False.
     rescale_every
-        Frequency to rescale velocities. Default is 10.
+        Frequency to rescale velocities and remove roation during equilibration.
+        Default is 10.
     file_prefix
         Prefix for output filenames. Default is inferred from structure, ensemble,
         and temperature.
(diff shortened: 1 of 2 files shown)


ElliottKasoar:
> what I think is needed to have a test where we have an input file that contains velocities and without any rescale_velocities in the input we keep the user velocities

Is that not what `test_no_reset_velocities` does?

lj-traj.xyz includes velocities, so I just reused it for the input file, and I check that with `rescale_velocities=False` (the default, but included to be explicit) the velocities are unchanged.

oerc0122 (approved):

Issue #673, opened 2026-02-27
Custom (Arch/Calculator dependent) properties in written structures (traj etc.)


harveydevereux:
### Summary

This may be possible some other way but it was not obvious to me (`write_kwargs` that goes to `ase.io`?) but for example MACE has the `node_energy` result which is the list of predicted per-atom energies. This would be very useful for thermal-conductivity and more analysis.

This is stored in `atoms.calc.results['node_energy'])` after a single_point.

```Python
from janus_core.calculations.single_point import SinglePoint
from ase.io import read

atoms = read("LiF.extxyz")
sp = SinglePoint(
    struct=atoms, 
    arch="mace",
    write_results=True
)

sp.run()

print(atoms.calc.results['node_energy'])
# array([-3.04302047, -5.38119082, -2.85913875, -5.24258456, -2.88435103, ...
# The predicted "Per-atom" energies, useful for thermal-conductivity.
```

At the moment though in the written results here we just have these properties,

```
Properties=species:S:1:pos:R:3:momenta:R:3:masses:R:1:mace_mp_forces:R:3:mace_forces:R:3:forces:R:3
```

But If you do,

```Python
atoms.set_array('atom_energy', atoms.calc.results['node_energy'])
write("LiF-sp.extxyz", atoms)
```

Then `atom_energy` is in the properties

```
Properties=species:S:1:pos:R:3:momenta:R:3:masses:R:1:mace_mp_forces:R:3:mace_forces:R:3:atom_energy:R:1:forces:R:3
```

### Example

User controlled properties like so (`atom_energy`),

```
Properties=species:S:1:pos:R:3:momenta:R:3:masses:R:1:mace_mp_forces:R:3:mace_forces:R:3:atom_energy:R:1:forces:R:3
```

harveydevereux:
It appears `Calculator.results` is standard https://gitlab.com/ase/ase/-/blob/master/ase/calculators/calculator.py?ref_type=heads#L853

There must be something I'm missing as to how to include these values in the output?

harveydevereux:
There may be other properties on the new models that could benefit from this (charges etc in MACE Polar) #682

PR #687, opened 2026-03-20
Mace polar

alinelena:


janus_core/helpers/mlip_calculators.py
```diff
@@ -229,6 +229,25 @@ def choose_calculator(
...
+            from chgnet.model.dynamics import CHGNetCalculator
+            from chgnet.model.model import CHGNet
+            import torch
+
+            # Set before loading to avoid type mismatches
+            torch.set_default_dtype(torch.float32)
+
         case "chgnet":
             from chgnet import __version__
             from chgnet.model.dynamics import CHGNetCalculator
```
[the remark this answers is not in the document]
alinelena: yap shall be fixed.... eager yyp

pyproject.toml
```diff
@@ -55,7 +55,7 @@ grace = [
     "tensorpotential == 0.5.1",
 ]
 mace = [
-    "mace-torch==0.3.14",
+    "mace-torch==0.3.15",
```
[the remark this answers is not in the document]
alinelena: yes I shall put it draft... mainly cause I was doing in various meetings.

ElliottKasoar:
https://github.com/ACEsuit/mace/releases/tag/v0.3.16 should makes things a little easier and resolve the thread above

trdurrant:
I had a quick go building this branch, bumping the requirement to `mace-torch==0.3.16` and hit two issues (which you most likely already know about):

- The `graph_longrange` dependency is needed
- ~~The default model polar-1-s isn't included in mace v0.3.16 release (which has polar-1-m and polar-1-l)~~ My mistake, it looks like this is included in the repo, but not part of the paper or documentation https://github.com/ACEsuit/mace-foundations/issues/65

With these addressed it looks like it is working (although I don't have anything to test against)

trdurrant:
The `mace_polar` charges do not seem to be directly available in the results, but I didn't have any problem fishing them out of the `atoms`  object as given in the example here https://github.com/stfc/janus-core/issues/673:

`charges = atoms.calc.results['charges']`

docs/source/tutorials/python/single_point.ipynb
```diff
@@ -51,17 +47,7 @@
...
-    "# locale.getpreferredencoding = lambda: \"UTF-8\"\n",
-    "\n",
-    "# ! pip uninstall numpy -y # Uninstall pre-installed numpy\n",
-    "\n",
-    "# ! pip uninstall torch torchaudio torchvision transformers -y # Uninstall pre-installed torch\n",
-    "# ! uv pip install torch==2.5.1 # Install pinned version of torch\n",
-    "\n",
-    "# ! uv pip install janus-core[mace,orb,chgnet,visualise] data-tutorials --system # Install janus-core with MACE, Orb, CHGNet, and WeasWidget, and data-tutorials\n",
-    "\n",
-    "# get_ipython().kernel.do_shutdown(restart=True) # Restart kernel to update libraries. This may warn that your session has crashed."
```
ElliottKasoar, line 64: Is this supposed to be removed? I think it does work without it at the moment, but I think we have this section in all the other tutorials (including the CLI singlepoint that's also changed.

ElliottKasoar (approved):
Other than maybe being consistent with the Colab install commands, this looks great

docs/source/tutorials/python/single_point.ipynb
```diff
@@ -203,6 +189,78 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(res.calc.results)\n",
+    "print(res.calc.results['dipole'])"
```
ElliottKasoar:
```suggestion
    "print(sp.struct.calc.results)\n",
    "print(sp.struct.calc.results['dipole'])"
```

docs/source/tutorials/python/single_point.ipynb
```diff
@@ -203,6 +189,78 @@
...
+    "benzene.info[\"charge\"] = 0\n",
+    "benzene.info[\"spin\"] = 1\n",
+    "sp = SinglePoint(\n",
+    "    struct=benzene,\n",
+    "    arch=\"mace_polar\",\n",
+    "    device=device,\n",
+    "    model=\"polar-1-s\",\n",
+    ")\n",
+    "res = sp.run()\n",
+    "print(res.calc.results)"
```
ElliottKasoar:
```suggestion
    "print(sp.struct.calc.results)"
```
or `print(res)`, depending on the aim here

ElliottKasoar (changes requested):
Couple of small fixes actually.

The return from our `SinglePoint.run` is a custom dictionary of results corresponding to what the user asked for (e.g. energy, forces, stresses, hessian), not a copy of the `calc.results` dict, so we need to access this to get the dipole etc.

docs/source/tutorials/python/single_point.ipynb
alinelena: yes works now without the workaround, maybe i remove the rest in a different pr

alinelena: I did them in this

alinelena:
> Couple of small fixes actually.
> 
> The return from our `SinglePoint.run` is a custom dictionary of results corresponding to what the user asked for (e.g. energy, forces, stresses, hessian), not a copy of the `calc.results` dict, so we need to access this to get the dipole etc.

one note here i think we shall copy all the calc.results info onto the atoms.info and similar in new ase4 this will have to be done explicity anyhow

ElliottKasoar (approved):
Looks good, and the tutorials run locally for me, so hopefully a release should work

PR #689, opened 2026-03-28
pass explicit the old method for neb to reestablish default behaviour…

alinelena:
… for test, also disable fmax check since seems is too sensible to be meaningful

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 readme = "README.md"
 
 dependencies = [
-    "ase<4.0,>=3.27",
+    "ase<4.0,==3.27",
     "click<9,>=8.2.1",
     "codecarbon<4.0.0,>=3.0.7",
     "numpy<3.0.0,>=1.26.4",
--- a/tests/test_neb.py
+++ b/tests/test_neb.py
@@ -61,7 +61,6 @@ def test_neb(tmp_path, LFPO_start_b, LFPO_end_b):
     assert all(key in neb.results for key in ("barrier", "delta_E", "max_force"))
     assert neb.results["barrier"] == pytest.approx(8117.328587956959)
     assert neb.results["delta_E"] == pytest.approx(-3.0149328722473e-07)
-    # assert neb.results["max_force"] == pytest.approx(148695.846153840771)
 
 
 def test_neb_pymatgen(tmp_path, LFPO_start_b, LFPO_end_b):
--- a/tests/test_neb.py
+++ b/tests/test_neb.py
@@ -147,7 +147,6 @@ def test_neb_functions(tmp_path, LFPO_start_b, LFPO_end_b):
     assert all(key in neb.results for key in ("barrier", "delta_E", "max_force"))
     assert neb.results["barrier"] == pytest.approx(8117.328587986063)
     assert neb.results["delta_E"] == pytest.approx(-3.0149328722473e-07)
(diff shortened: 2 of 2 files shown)


tests/test_neb.py
```diff
@@ -60,7 +61,7 @@ def test_neb(tmp_path, LFPO_start_b, LFPO_end_b):
     assert all(key in neb.results for key in ("barrier", "delta_E", "max_force"))
     assert neb.results["barrier"] == pytest.approx(8117.328587956959)
     assert neb.results["delta_E"] == pytest.approx(-3.0149328722473e-07)
-    assert neb.results["max_force"] == pytest.approx(148695.846153840771)
+    # assert neb.results["max_force"] == pytest.approx(148695.846153840771)
```
ElliottKasoar: Is this value no longer correct?

pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 readme = "README.md"
 
 dependencies = [
-    "ase<4.0,>=3.27",
+    "ase<4.0,==3.27",
```
ElliottKasoar, line 29: Does't this pin it, so there's no point also having `<4`?

I wonder if we should put this as a constraint in the dev group, rather than making this a full pin?

alinelena: needs rdf fixing... otherwise we need to fix rdf in this... is just temp.

tests/test_neb.py
alinelena: difference is too high commented it for test we can reinstate it if needed but reality is at these big numbers makes little sense to test.

pyproject.toml
ElliottKasoar: I understand the pin, just not why it wouldn't just be `"ase==3.27"`, since it currently looks like a range still on first appearance.

For the alternative, I mean we could do

```
[dependency-groups]
dev = [
    "ase==3.27",
...
]
```

This would be the same fix for RDFs, but potentially could be less temporary. It is essentially true that our tests depend on a fixed version of ASE due to what are basically regression tests.

tests/test_neb.py
ElliottKasoar: Ok let's just delete this line then

tests/test_neb.py
```diff
@@ -120,7 +122,7 @@ def test_set_calc(tmp_path, LFPO_start_b, LFPO_end_b):
     assert all(key in neb.results for key in ("barrier", "delta_E", "max_force"))
     assert neb.results["barrier"] == pytest.approx(8117.328587986063)
     assert neb.results["delta_E"] == pytest.approx(-3.0149328722473e-07)
-    assert neb.results["max_force"] == pytest.approx(148695.846153840771)
```
harveydevereux, line 123: can the others be removed as well? Another on line 150

alinelena: for a reason it does not fail

harveydevereux: sure, but you just removed the other comment test but not these

tests/test_neb.py
```diff
@@ -120,7 +122,7 @@ def test_set_calc(tmp_path, LFPO_start_b, LFPO_end_b):
...
+    # assert neb.results["max_force"] == pytest.approx(148695.846153840771)
```
ElliottKasoar:
```suggestion
```

harveydevereux (approved):
Looks good, once it passes should allow for #691 too

ElliottKasoar (approved):

PR #690, opened 2026-03-29
Tests cuda

alinelena:
adding posibility some of the tests to run on cuda. create a workflow for them.. shall we force the other tests to run cpu only?

### DIFF
--- a/.github/workflows/bluesky.yml
+++ b/.github/workflows/bluesky.yml
@@ -8,8 +8,6 @@ jobs:
     runs-on: [self-hosted, gpu]
     if: github.repository == 'stfc/janus-core'
     timeout-minutes: 60
-    env:
-      CODECARBON_DISABLED: 'true'
     strategy:
       matrix:
         python-version: ["3.10","3.11","3.12"]
--- a/.github/workflows/bluesky.yml
+++ b/.github/workflows/bluesky.yml
@@ -1,0 +1,58 @@
+name: gpu-ci
+
+on: [push, pull_request]
+
+jobs:
+
+  tests:
+    runs-on: [self-hosted, gpu]
+    if: github.repository == 'stfc/janus-core'
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        python-version: ["3.10","3.11","3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install "all" dependencies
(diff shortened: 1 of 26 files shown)


alinelena:
needs !689 to get in

ElliottKasoar:
I think this is based off of #689, so we should probably merge that first and rebase.

Tests are looking better re-running now. I'm wondering if the error was from https://github.com/mlco2/codecarbon/pull/1122, which has been fixed as of a release yesterday. It says it was a warning, but perhaps the underlying `KeyError` meant the emissions tracker wasn't set up, and so we kill it.

Looks like we need to loosen the dispersion comparison slightly too.

janus_core/helpers/log.py
```diff
@@ -213,5 +213,6 @@ def config_tracker(
                 "reconfiguring, or disable the tracker.",
                 stacklevel=2,
             )
+            tracker = None
```
alinelena, line 216: is not already set on line 191?

ElliottKasoar: Yes, but then `tracker = OfflineEmissionsTracker(...` doesn't fail, it just fails to set up correctly, which is why is doesn't have `_emissions`, so it's overwritten

.github/workflows/bluesky.yml
```diff
@@ -0,0 +1,60 @@
...
+on: [push, pull_request]
+
+jobs:
+
+  tests:
+    runs-on: [self-hosted, gpu]
+    if: github.repository == 'stfc/janus-core'
+    timeout-minutes: 60
+    env:
+      CODECARBON_DISABLED: 'true'
```
ElliottKasoar:
```suggestion
```
I assume this wasn't doing anything?

.github/workflows/bluesky.yml
```diff
@@ -0,0 +1,60 @@
...
+    strategy:
+      matrix:
+        python-version: ["3.10","3.11","3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
```
ElliottKasoar:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/bluesky.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
ElliottKasoar:
```suggestion
        uses: astral-sh/setup-uv@v7
```
(8 is also released, but we use 7 everywhere else)

.github/workflows/bluesky.yml
alinelena: nope was part of some implementation i abandoned

PR #691, opened 2026-04-02
Fix RDFs for ase 3.28 by using get_rdf directly.

### LINKED ISSUE #630: final rdfs should be usable
### Summary

at the moment seems our rdfs are not most useful
see the note 

N.B. mixed RDFs (e.g. C-H) include all self-RDFs (e.g. C-C),
        to get the pure (C-H) RDF subtract the self-RDFs.

I suspect at time of the merge looked ok but is not the case anymore.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

harveydevereux:
Post-process now uses `ase.geometry.get_rdf` directly rather than the `Analysis` object.

Fixes #630 by supporting the 3.28 ASE changes.

There is also currently a refactor in ASE master that I think may require further changes later on.

Note that the failing "by-elements" test is the C-H part which makes sense given the behaviour in #630. 

```Python
        ("C", "H"): ( # old
            1.075,
            1.375,
            2.175,
            2.425,
            2.475,
            2.775,
            3.425,
            3.875,
            4.275,
            4.975,
        ),
        ("C", "H"): (1.075, 2.175, 3.425, 3.875), # new
```

### DIFF
--- a/janus_core/processing/post_process.py
+++ b/janus_core/processing/post_process.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Sequence
 from itertools import combinations_with_replacement
+from warnings import warn
 
 from ase import Atoms
 from ase.geometry.analysis import Analysis
@@ -41,7 +42,7 @@ def compute_rdf(
     data
         Dataset to compute RDF of.
     ana
-        ASE Analysis object for data reuse.
+        Deprecated. Please do not use. ASE Analysis object for data reuse.
     filenames
         Filenames to output data to. Must match number of RDFs computed.
     by_elements
@@ -70,8 +71,17 @@ def compute_rdf(
         If `by_elements` is true returns a `dict` of RDF by element pairs.
         Otherwise returns RDF of total system filtered by elements.
     """
-    if ana is not None and by_elements:
-        raise ValueError("Using Analysis.get_rdf split by elements has known bugs.")
+    if ana is not None:
+        warn(
+            "ana has been deprecated.",
+            FutureWarning,
+            stacklevel=2,
+        )
+        if by_elements:
+            raise ValueError(
+                "Analysis.get_rdf has known bugs with by_elements."
+                "Call without ana to use ase.geometry.rdf.get_rdf directly."
+            )
 
     index = slicelike_to_startstopstep(index)
 
@@ -135,17 +145,25 @@ def compute_rdf(
                         print(dist, rdf_i, file=out_file)
 
     else:
-        if ana is None:
-            ana = Analysis(data)
+        if ana is not None:
+            rdf = ana.get_rdf(
+                rmax=rmax,
+                nbins=nbins,
+                imageIdx=slice(*index),
+                return_dists=True,
+                volume=volume,
+            )
+        else:
+            rdf = [
+                get_rdf(
+                    atoms,
+                    rmax,
+                    nbins,
+                    volume=volume,
+                )
+                for atoms in data
+            ]
 
-        rdf = ana.get_rdf(
-            rmax=rmax,
-            nbins=nbins,
-            elements=elements,
-            imageIdx=slice(*index),
-            return_dists=True,
-            volume=volume,
-        )
         assert isinstance(rdf, list)
 
         # Compute RDF average
--- a/janus_core/processing/post_process.py
+++ b/janus_core/processing/post_process.py
@@ -6,7 +6,7 @@ from collections.abc import Sequence
 from itertools import combinations_with_replacement
 
 from ase import Atoms
-from ase.geometry.analysis import Analysis
+from ase.geometry.rdf import get_rdf
 import numpy as np
 from numpy import float64
 from numpy.typing import NDArray
@@ -21,7 +21,6 @@ from janus_core.helpers.utils import build_file_dir, slicelike_to_startstopstep
 
 def compute_rdf(
     data: MaybeSequence[Atoms],
-    ana: Analysis | None = None,
     /,
     *,
     filenames: MaybeSequence[PathLike] | None = None,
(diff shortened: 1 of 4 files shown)


janus_core/processing/post_process.py
```diff
@@ -21,7 +21,6 @@
 
 def compute_rdf(
     data: MaybeSequence[Atoms],
-    ana: Analysis | None = None,
```
ElliottKasoar, line 24: So previously this meant we could swap out the RDF analysis function. Does that not work/do you not think it's useful?

harveydevereux: Yes right, the error is `RuntimeError: There are bugs for `elements` is not `None`. Use `ase.geometry.rdf.get_rdf` instead.` so we can use Analysis if `elements is None`

ElliottKasoar: Sorry, to clarify, I know the change with respect to `Analysis` specifically (more or less...).

I thought that when this was introduced the idea was that we could also use a non-ASE function here instead, e.g. something more efficient, but perhaps I'm misremembering and/or it's not really feasible since we don't know exactly how things will be returned.

harveydevereux: Yes that would definitely be useful but as you say comes down to how similar the options are.

I thought it best to solve this RDF issue soon, then we should plan it properly perhaps. Currently it's blocking 3.28 right?

Its worth a look to incorporate something that can be swapped out, a two examples are MDAnalysis https://docs.mdanalysis.org/1.1.1/documentation_pages/analysis/rdf.html, and there was a CUDA thing Alin mentioned at some point

ElliottKasoar: Yep true I'm fine for this to be a multi-step thing, we're not currently exploiting it in the previous form, and I'm not sure anything else could actually be subbed in as is anyway.

ElliottKasoar:
There are NEB errors that we can ignore since they're fixed by #689, but there are some post-processing errors from this change to be fixed.

janus_core/processing/post_process.py
```diff
@@ -129,6 +135,9 @@ def compute_rdf(
                         print(dist, rdf_i, file=out_file)
 
     else:
+        if ana is None:
+            ana = Analysis(data)
+
```
ElliottKasoar: Does this not mean if we're not doing `by_elements`, we default to using `Analysis`, which is deprecated? 

Is there actually a reason to keep `Analysis` at all?

Given that as we discussed elsewhere, we don't properly supported generic analysis drop-ins, I don't think we need to keep supporting `Analysis` unless it does things we can't do otherwise?

harveydevereux: It should throw the error at the start

```python
if ana is not None and by_elements:
        raise ValueError("Using Analysis.get_rdf split by elements has known bugs.")
```

But it seems only deprecated when by_elements is true, just going by the error text `RuntimeError: There are bugs for elements is not None` and there is no error when you use it.

I don't know what Analysis can do and I don't use it, so personally I have no reason to keep it. We can drop it as I did previously but since you mentioned 

> So previously this meant we could swap out the RDF analysis function. Does that not work/do you not think it's useful?

I thought you wanted it left since it could be useful

ElliottKasoar: I meant in the case where `by_elements` is `False`.

From the code, the deprecation doesn't seem to be qualified:

```
    @deprecated("Use `ase.geometry.rdf.get_rdf` instead.")
    def get_rdf(self, rmax, nbins, imageIdx=None, elements=None, return_dists=False,
                volume: float | None = None):
```

If it's easy to have it working with both the recommended `get_rdf` and `Analysis`, I'm happy for it to still work - as you say I am in favour of more generality, although in general I'm more imagining swapping for non-ASE functions.

I don't think it should be the default for any of the logic branches though (unless perhaps it needs to be if we support older ASEs and get_rdf didn't work?) if the entire function is being deprecated at some point.

harveydevereux: I see, I did not actually see the deprecation. I thought they just threw an error while they fixed it...

How about we just not have analysis as an argument? I'm not pushing for it in anyway, that default use was just because previously it was defaulted.

I have no idea about support in older version though...

ElliottKasoar: That's fine with me.

We currently have `>=3.27`, so I'd probably just run the relevant tests with that locally, and if everything still passes (and for 3.28 in the CI), we're probably ok.

harveydevereux: > That's fine with me.
> 
> We currently have `>=3.27`, so I'd probably just run the relevant tests with that locally, and if everything still passes (and for 3.28 in the CI), we're probably ok.

With the exception of the changed peaks for C-H, the tests do pass with 3.27

tests/test_post_process.py
```diff
@@ -132,6 +133,51 @@ def test_rdf():
...
+            3.875,
+            4.275,
+            4.975,
+        )
+    )
+    peaks = np.where(rdf[1] > 0.0)
+    assert (np.isclose(expected_peaks, rdf[0][peaks])).all()
+
+
+def test_rdf_by_elementw_analysis_error():
```
ElliottKasoar:
```suggestion
def test_rdf_by_elements_analysis_error():
```

ElliottKasoar (approved):
Looks good if the tests pass, thanks @harveydevereux!

I'll double check before doing a release but tutorials all seem ok still too.

harveydevereux:
> Looks good if the tests pass, thanks @harveydevereux!
> 
> I'll double check before doing a release but tutorials all seem ok still too.



> Looks good if the tests pass, thanks @harveydevereux!
> 
> I'll double check before doing a release but tutorials all seem ok still too.

great, locally it was ok so lets hope

PR #698, opened 2026-04-15
Add NVT post-processing for TDEP input generation

junwen94:
## Summary

This PR adds support for generating standard TDEP input files from NVT molecular dynamics runs in `janus-core`.

The new workflow allows TDEP inputs to be generated automatically through `post_process_kwargs` during an `NVT` simulation, rather than requiring a separate manual conversion step.

## Changes

- add a new TDEP processing module in `janus_core/processing/tdep.py`
- add `build_tdep_inputs_from_nvt` for generating standard TDEP input files
- integrate TDEP input generation into MD post-processing
- extend `PostProcessKwargs` with TDEP-related options
- add integration test coverage in `tests/test_tdep.py`
- add a new Python tutorial notebook for anharmonic phonons in `docs/source/tutorials/python/anharmonic_phonons.ipynb`
- register the new tutorial in `docs/source/tutorials/python/index.rst`

## Notes

- this is an initial integration for the NVT workflow
- the generated files are `infile.ucposcar`, `infile.ssposcar`, `infile.positions`, `infile.forces`, `infile.stat`, and `infile.meta`

## Testing

```bash
uv run pytest tests/test_tdep.py -q

### CODE REVIEW

File: janus_core/processing/tdep.py [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+
+    trajectory = read(traj_file, index=":")
+    n_atoms = len(trajectory[0])
+    n_steps = len(trajectory)
+
+    with open(meta_path, "w", encoding="utf-8") as file:
+        file.write(f"{n_atoms}    # N atoms\n")
+        file.write(f"{n_steps}    # N timesteps\n")
+        file.write(f"{timestep}    # timestep in fs\n")
+        file.write(f"{temperature}    # temperature in K\n")
```
oerc0122:
```suggestion
    meta_path.write_text(f"""\
{n_atoms}    # N atoms
{n_steps}    # N timesteps
{timestep}    # timestep in fs
{temperature}    # temperature in K
    """)
```

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+
+    unit_cell = read(unit_cell_file)
+    if len(supercell_atoms) != n_atoms:
+        raise ValueError(
+            "The number of atoms in supercell must match the number of atoms "
+            "in traj_file."
+        )
```
oerc0122, line 79: Why is this check down here?

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+
+    write(ucposcar_path, unit_cell, format="vasp")
+    write(ssposcar_path, supercell_atoms, format="vasp")
+
+    with open(positions_path, "w", encoding="utf-8") as file:
+        for atoms in trajectory:
+            for position in atoms.get_scaled_positions():
+                file.write(
+                    f"{position[0]:.12f} {position[1]:.12f} {position[2]:.12f}\n"
+                )
```
oerc0122: The following allows a faster buffered write.
```suggestion
        file.writelines(
             f"{position[0]:.12f} {position[1]:.12f} {position[2]:.12f}\n"
             for atoms in trajectory:
             for position in atoms.get_scaled_positions():
         )
```

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+
+    with open(forces_path, "w", encoding="utf-8") as file:
+        for atoms in trajectory:
```
oerc0122: As above

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+            for force in atoms.get_forces():
+                file.write(f"{force[0]:25.12f}{force[1]:25.12f}{force[2]:25.12f}\n")
```
oerc0122: Are you sure you want no spaces between these?

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+
+    stats_data = np.atleast_2d(np.loadtxt(stats_file, comments="#"))
+
+    if len(stats_data) != n_steps:
+        raise ValueError(
+            "The number of rows in stats_file must match the number of frames "
+            "in traj_file."
+        )
```
oerc0122: I would prefer these checks done before half the files have been dumped.

janus_core/processing/tdep.py
```diff
@@ -0,0 +1,147 @@
...
+    pressure = stats_data[:, 9]
+    stress_xx = stats_data[:, 10]
+    stress_yy = stats_data[:, 11]
+    stress_zz = stats_data[:, 12]
+    stress_yz = stats_data[:, 13]
+    stress_xz = stats_data[:, 14]
+    stress_xy = stats_data[:, 15]
+
+    with open(stat_path, "w", encoding="utf-8") as file:
+        for i in range(len(time_fs)):
```
oerc0122: As above

janus_core/processing/tdep.py
junwen94: Done, thanks. I moved the validation checks before any TDEP files are written.

janus_core/processing/tdep.py
junwen94: Done, thanks. I simplified this using `Path.write_text(...)`.

janus_core/processing/tdep.py
junwen94: Done, thanks. Switched this to buffered `writelines(...)`.

janus_core/processing/tdep.py
junwen94: Done, thanks. I switched this to `writelines(...)` and made the columns explicitly space-separated.

janus_core/processing/tdep.py
junwen94: Done, thanks. Switched this to buffered `writelines(...)` as well.

junwen94:
Addressed Jacob's review comments and pushed an update. Verified with `uv run pytest tests/test_tdep.py -q`.

janus_core/processing/tdep.py
junwen94: Modified, thanks.

janus_core/processing/tdep.py
```diff
@@ -69,41 +69,44 @@ def build_tdep_inputs_from_nvt(
...
+        raise ValueError(
+            "The number of rows in stats_file must match the number of frames "
+            "in traj_file."
+        )
+
+    meta_path.write_text(
+        f"{n_atoms}    # N atoms\n"
+        f"{n_steps}    # N timesteps\n"
+        f"{timestep}    # timestep in fs\n"
+        f"{temperature}    # temperature in K\n",
```
oerc0122: Can still avoid more repetition here:
```python
        f"""\
{n_atoms}    # N atoms
{n_steps}    # N timesteps
{timestep}    # timestep in fs
{temperature}    # temperature in K
        """,
```

janus_core/processing/tdep.py
junwen94: Moved, thanks!

janus_core/processing/tdep.py
```diff
@@ -69,41 +69,44 @@ def build_tdep_inputs_from_nvt(
...
     write(ssposcar_path, supercell_atoms, format="vasp")
 
     with open(positions_path, "w", encoding="utf-8") as file:
-        for atoms in trajectory:
-            for position in atoms.get_scaled_positions():
-                file.write(
-                    f"{position[0]:.12f} {position[1]:.12f} {position[2]:.12f}\n"
-                )
+        file.writelines(
+            f"{position[0]:.12f} {position[1]:.12f} {position[2]:.12f}\n"
```
oerc0122: If you make this format your standard you could either do:
```python
out_fmt = "{:12f}".format
```
or
```python
out_fmt = partial(format, "%.12f")
```
then this would become
```suggestion
            " ".join(map(out_fmt, position) + "\n"
```

janus_core/processing/tdep.py
```diff
@@ -120,22 +123,22 @@ def build_tdep_inputs_from_nvt(
...
+            f"{e_pot[i]:.12f} "
+            f"{e_kin[i]:.12f} "
+            f"{temperature_data[i]:.12f} "
+            f"{pressure[i]:.12f} "
+            f"{stress_xx[i]:.12f} "
+            f"{stress_yy[i]:.12f} "
+            f"{stress_zz[i]:.12f} "
+            f"{stress_xz[i]:.12f} "
+            f"{stress_yz[i]:.12f} "
+            f"{stress_xy[i]:.12f}\n"
```
oerc0122: Likewise with the above modification
```python
            f"{i:d} " +
            " ".join(
                out_fmt(dat[i]) for dat in
                (
                    time_fs,
                    e_tot,
                    e_pot,
                    e_kin,
                    temperature_data,
                    pressure,
                    stress_xx,
                    stress_yy,
                    stress_zz,
                    stress_xz,
                    stress_yz,
                    stress_xy,
                )
            ) + "\n"
```

oerc0122 (dismissed):
Just some minor recommendations, otherwise looks ok.

alinelena:
be sure we add tdep in CI and we actually test it

docs/source/tutorials/python/anharmonic_phonons.ipynb
ElliottKasoar, line 1: Normally these are run automatically via `nbsphinx`, so we don't usually save the output files (e.g. the png) or any of the notebook cell outputs, and it means we can test this all works when we build for the docs these on release.

Is there any reason not to do the same for this?

.gitignore
```diff
@@ -39,3 +39,5 @@ results/
 logs/
 *.model
 tests/models/extra
+!docs/source/tutorials/python/anharmonic_phonons.ipynb
+!docs/source/tutorials/data/outfile.dispersion_relations.png
```
ElliottKasoar: Generally it's easier just to `git add -f` these sorts of things else this will become an incredibly long list

ElliottKasoar (commented):
Thanks for this, @junwen94!

Would you be able to add a (very brief) CLI tutorial for this too?

Related to that, it would be great to include these kwargs in the example configs found in `docs/source/examples/full/`, similarly to how we already show `rdf_compute` etc. This provides another example for how you can specify these.

alinelena:
@junwen94 can you go through the comments please? and also we will need to install tdep in the ci to have a proper test, in the sense tdep runs and we get to test the real data.

.github/workflows/ci.yml
```diff
@@ -83,11 +83,15 @@ jobs:
 
   docs:
```
oerc0122, line 84: I'm concerned this is being done on the docs workflow.

.github/workflows/ci.yml
```diff
@@ -101,7 +105,43 @@ jobs:
...
+        shell: bash
+        run: sudo apt-get update && sudo apt-get install -y poppler-utils gnuplot-nox
+
+      - name: Set up Miniconda for TDEP
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          miniconda-version: latest
+          auto-update-conda: true
+          activate-environment: tdep
+          python-version: "3.10"
```
oerc0122, line 118: Very concerned that we not only have two package managers, but two different versions of python simultaneously. That's a recipe for disaster.

.github/workflows/ci.yml
```diff
@@ -101,7 +105,43 @@ jobs:
       - name: Install pandoc
         uses: pandoc/actions/setup@v1
 
-      - name: Install dependencies
+      - name: Install poppler-utils and gnuplot
+        shell: bash
+        run: sudo apt-get update && sudo apt-get install -y poppler-utils gnuplot-nox
```
oerc0122, line 110: Why do we need to add gnuplot? Poppler utils? Are these needed for users? none of this is documented anywhere.

oerc0122 (changes requested):

tests/test_tdep.py
ElliottKasoar, line 1: I think ideally we want to (optionally) test more than just the input generation.

With PLUMED, for example, we have `tests/test_md_plumed.py`, which is skipped if it is not installed correctly, but if it is (which we can set up in the CI), then we check it is actually integrated correctly.

.github/workflows/ci.yml
ElliottKasoar: In case it's not clear, @junwen94, the docs that are built during `ci.yml` (which runs on every push, and for every PR) do not run the notebook tutorials (`make all`), they just check the docs build from the .rst files (`make html`).

The tutorials will be run as part of the `publish-on-pypi.yml` workflow, which is triggered on new tagged releases, so you should still test locally that both `make` commands work.

You will need installations in `ci.yml`, but only for the `tests` section of the workflow, as we have for PLUMED (assuming, as I suggest elsewhere, we actually run tdep, and not just test the input generation).

PR #717, opened 2026-09-17
add --fix-symmetry option

alinelena:


tests/test_geomopt_cli.py
```diff
@@ -778,6 +778,38 @@ def test_symmetrize(tmp_path):
...
+            log_path,
+            "--summary",
+            summary_path,
+        ],
+    )
+    assert result.exit_code == 0
+    assert results_path.exists()
+    with open(summary_path, encoding="utf-8") as file:
+        summary = yaml.safe_load(file)
+    assert summary["config"]["fix_symmetry"] is True
```
ElliottKasoar, line 810: It might be worth also adding a check that the log contains the info message, as we do elsewhere, as proxy for checking that part of the code is reached.

ElliottKasoar (commented):
Looks great, thanks! A question that came up in #313 is whether these constraints are written out to the extxyz file.

At the very least, it's worth confirming the behaviour, but there was also a suggestion to strip it from the output.

tests/test_geomopt_cli.py
alinelena: there is no extra info in extxyz for this type of constraint, the ones written in extxyz were about fixed atoms if I recall correctly

ElliottKasoar: Ok that addresses my other question, but this one was poorly worded. You added:

```
            if self.logger:
                self.logger.info(
                    "FixSymmetry constraint applied with symprec=%s, adjust_cell=%s",
                    self.symmetry_tolerance,
                    self.filter_class is not None,
                )
```

Normally then in the test we then have something along the lines of:

```
    assert_log_contains(
        log_path,
        includes=["Starting geometry optimization", "Using optimizer: FIRE"],
    )
```

to check this has been triggered in the code

alinelena: updates is that what you meant?

ElliottKasoar (approved):
Looks good to me if the tests pass

PR #719, opened 2026-09-17
add support for fennix/fennol models

alinelena:


pyproject.toml
```diff
@@ -56,6 +56,9 @@ d3 = [
 grace = [
     "tensorpotential == 0.6.0",
 ]
+fennix = [
+    "fennol >= 2026.6.29",
```
ElliottKasoar:
```suggestion
    "fennol == 2026.6.29",
```
We normally pin these else we end up with the CI failing whenever a version changes

tests/models/ani2x.fnx
ElliottKasoar, line 1: Is this supposed to be added when we also have the `try_retrieve`?

ElliottKasoar (commented):
Generally looks good, just a couple of minor questions

PR #722, opened 2026-09-17
Drop Python 3.10

alinelena:


ElliottKasoar:
We need to update `pyproject.toml` too (`requires-python`/`classifiers`), and we may nee to be careful about dependency issues on 3.14, since it's only quite recent that we managed to get enough support for 3.13

alinelena:
> We need to update `pyproject.toml` too (`requires-python`/`classifiers`), and we may nee to be careful about dependency issues on 3.14, since it's only quite recent that we managed to get enough support for 3.13

yes 3.14 was more a case does it work... I will clean it up tomorrow

ElliottKasoar (approved):
Looks good, thanks!

We'll need to do some dependency updates/fixes before the next release after this, as currently Orb is pinned to <3.13, which will mean some tutorials will fail

PR #724, opened 2026-09-18
Unpin phonopy

ElliottKasoar:
I think changes added in #717 may resolve the error that was raised for isolated molecules


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "codecarbon<4.0.0,>=3.0.7",
     "numpy<3,>=1.26.4; python_version < '3.13'",
     "numpy<3,>=2.1.0; python_version >= '3.13'",
-    "phonopy<2.48,>=2.39.0",
+    "phonopy<3,>=2.48.0",
```
alinelena: current version is 4.5 so you may want something like that

alinelena (approved):

PR #725, opened 2026-09-19
Update click dependency version constraints

alinelena:


ElliottKasoar:
I think #726 is simpler

PR #723, opened 2026-09-18
Update dependencies

ElliottKasoar:
Updates several MLIPs, including Orb, which has changed its API.

UPET and Orb now conflict, so these are tested separately

PR #726, opened 2026-09-21
Drop click decency

ElliottKasoar:
#713 removed all of our explicit references to `click`, but it was still a dependency of `typer` until 0.26.0, which is why I (think I) kept the dependency, as earlier versions could still break older versions of `typer`.

This update also makes the dependency a lot stricter than previously, since minor version bumps of `typer` are still "breaking". This hasn't been a problem yet, as these changes have almost always been relatively cosmetic, but it's probably better to be safe.


alinelena:
ok let us try... can you release after this is in?

alinelena (approved):

ElliottKasoar:
> ok let us try... can you release after this is in?

#723 is needed (it could be similar if we run into any sticking points), otherwise the tutorials will fail to run fully and it won't publish

But yes, as soon as we can, I will

